PR #45, opened 2022-04-28
CI test PR from fork

regisss:
Fake PR from fork to test CI.


regisss:
CI does not currently work when a PR comes from a fork. This is due to the fact that Github Actions secrets are currently needed to launch jobs on AWS and these secrets are not accessible by forks.

PR #47, opened 2022-04-29
Add Albert XXL 8x and concurrency to CI/CD

regisss:
# What does this PR do?

This PR modifies CI/CD:
- ALBERT XXL 8x is now added to non-regression tests. 1x version is still filtered because it takes 6 hours, which is the maximum time a job can run in a Github workflow (so there is a risk that the job is interrupted right before the end).
- There is a better separation between 1x, 8x and example diff tests. Each kind of test now has its own command in the Makefile.
- Add *concurrency* to all workflows to manage computation resources better:
  - if a fast test is run in a PR while there is already one running (from the same PR), the former one will be cancelled
  - if a slow test is run while there is already one running, the most recent one will be queued

## Before submitting

### DIFF
--- a/tests/ci/slow_tests_8x.sh
+++ b/tests/ci/slow_tests_8x.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+pip install --upgrade pip
+export RUN_SLOW=true
+make slow_tests_8x
--- a/.github/workflows/fast_tests.yml
+++ b/.github/workflows/fast_tests.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
     branches: [ main ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   start-runner:
     name: Start self-hosted EC2 runner
--- a/.github/workflows/slow_tests.yml
+++ b/.github/workflows/slow_tests.yml
@@ -6,7 +6,7 @@ on:
     - cron: 0 23 * * *  # every day at 1am CET
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}
 
 jobs:
   start-runner:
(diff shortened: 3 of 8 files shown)


tests/ci/slow_tests_8x.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+pip install --upgrade pip
+export RUN_SLOW=true
```
libinta, line 4: where this RUN_SLOW is used?

libinta (commented):
Regis, 
Is there any concept like partial test? For the longer one, do you think any partial test will work?thx

tests/ci/slow_tests_8x.sh
regisss: In `tests/test_examples.py`, at line 120, there is a decorator called `slow` that wraps all non-regression tests. With this decorator, you need to declare the environment variable `RUN_SLOW` as true so that these tests are not skipped.
It basically avoids running all those long tests if one launches `pytest tests/`.

regisss:
> Regis, Is there any concept like partial test? For the longer one, do you think any partial test will work?thx

I'm not sure I got you here. You're asking if we could make those tests shorter (like one epoch instead of 2 for instance)?

libinta:
Right, like run full regression every other day or every couple of days, but get partial test run for each PR? Do we have partial test defined?
Thanks,
Libin

From: regisss ***@***.***>
Reply-To: huggingface/optimum-habana ***@***.***>
Date: Monday, May 2, 2022 at 12:13 AM
To: huggingface/optimum-habana ***@***.***>
Cc: Libin Tang ***@***.***>, Review requested ***@***.***>
Subject: Re: [huggingface/optimum-habana] Add Albert XXL 8x and concurrency to CI/CD (PR #47)


Regis, Is there any concept like partial test? For the longer one, do you think any partial test will work?thx

I'm not sure I got you here. You're asking if we could make those tests shorter (like one epoch instead of 2 for instance)?

libinta (approved):

PR #48, opened 2022-05-02
Add a new argument for taking warmup steps into account in throughput calculation

regisss:
# What does this PR do?

This PR adds a new argument `throughput_warmup_steps` that enables to remove warmup steps from the throughput calculation at the end of the training (in lazy mode for instance).

This requires to override the `speed_metrics` method in `trainer_utils.py`.

Examples and tests have been updated accordingly.

## Before submitting

### DIFF
--- a/examples/question-answering/README.md
+++ b/examples/question-answering/README.md
@@ -53,7 +53,8 @@ python run_qa.py \
   --doc_stride 128 \
   --output_dir ./output/debug_squad/ \
   --use_habana \
-  --use_lazy_mode
+  --use_lazy_mode \
+  --throughput_warmup_steps 2
 ```
 
 Training with the previously defined hyper-parameters yields the following results:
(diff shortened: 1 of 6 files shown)


examples/question-answering/README.md
```diff
@@ -53,7 +53,8 @@ python run_qa.py \
   --doc_stride 128 \
   --output_dir ./output/debug_squad/ \
   --use_habana \
-  --use_lazy_mode
+  --use_lazy_mode \
+  --throughput_warmup_steps 2
```
libinta, line 57: is this througput_warmup_steps configurable since for 8x it may take more than 2

regisss: Yeah you can give it any value. I put 2 in the examples because that corresponds to what I observed in practice, but this could be `--throughput_warmup_steps N` with N a positive integer.

libinta (approved):

PR #54, opened 2022-06-01
Update baselines and CI script

regisss:
# What does this PR do?

The script for regression tests is slightly refactored so that it is easier to add new tasks with different metrics. Baselines have been updated accordingly.

## Before submitting


regisss:
> @regisss the change looks good, is it only for cleanup?

Actually it's mostly for being able to add new evaluation metrics more easily.

PR #69, opened 2022-07-20
Fix emojis in table

lewtun:
# What does this PR do?

Fixes the rendering of some emojis in the supported models table.

## Before submitting


regisss (approved):
Thanks for fixing this @lewtun!!

PR #72, opened 2022-07-25
Run CI regression tests on on-prem Gaudi

mfuntowicz:
Move the CI regression tests to on-prem Gaudi.

Fast (unit and integration) tests run for each commit and PR are still run on AWS instances to avoid collision with other fast tests or runs triggered by users. Since all those cannot be planned beforehand, it is better to run them on different AWS instances.

This PR also modifies the following:
- All the slow tests are now performed, whether one of these tests failed or not. Tests following a failed test used to be skipped.
- Update some CI baselines.
- Add a diff file for the `run_image_classification.py` example. It should have been added in #80.

### DIFF
--- a/.github/workflows/fast_tests.yml
+++ b/.github/workflows/fast_tests.yml
@@ -32,8 +32,8 @@ jobs:
           ls -la ./
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Check for dockerenv file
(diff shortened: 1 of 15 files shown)


regisss (approved):
Thanks a lot @mfuntowicz :fire: :rocket: 
I'm going to launch the regression tests on this branch tonight and if everything works as expected I will merge this PR tomorrow.

PR #78, opened 2022-08-04
Checkout the PR commit in Docker container when building the doc for PR

regisss:
# What does this PR do?

The documentation of the repo is built in a Docker container. When building the Docker image, the repo is cloned and so it checkouts by default the `main` branch. However, we want to build the doc for the current PR so we need to pass the commit sha of the current PR in order to check it out when building the Docker image.

## Before submitting


regisss:
cc @lewtun

PR #83, opened 2022-08-15
Add run_mlm.py in language-modeling examples

regisss:
# What does this PR do?

This PR adds the masked-language-modeling example from Transformers: `run_mlm.py`.

## Before submitting


libinta (commented):
@regisss  do we have which transformer version should be installed? and where is the instruction?

regisss:
> @regisss do we have which transformer version should be installed? and where is the instruction?

@libinta What we have done so far is to install the latest Transformers version. It is regsitered as a dependency in `setup.py` so it will be installed when `optimum-habana` is installed. If any issue appears with a new version of Transformers, we will see it in the CI and we fix it as fast as possible.

What we could do to manage this better is the following:
- When a new version of `optimum-habana` is released, freeze the version of Transformers in `setup.py`.
- Run the CI with the `main` branch of Transformers so that we catch changes and potential issues as soon as possible and before the next release of the package.

What do you think? We could start doing that from the next release of `optimum-habana`.

examples/language-modeling/run_mlm.py
```diff
@@ -0,0 +1,614 @@
...
+from optimum.habana import GaudiConfig, GaudiTrainer, GaudiTrainingArguments
+from transformers import (
+    CONFIG_MAPPING,
+    MODEL_FOR_MASKED_LM_MAPPING,
+    AutoConfig,
+    AutoModelForMaskedLM,
+    AutoTokenizer,
+    DataCollatorForLanguageModeling,
+    HfArgumentParser,
+    Trainer,
```
philschmid, line 44: can be removed

regisss: Good catch, thanks!

philschmid (approved):

PR #89, opened 2022-08-19
Memory stats

regisss:
# What does this PR do?

This PR adds live memory statistics on Gaudi in the logs:
- `memory_allocated (GB)` which is the current memory consumption,
- `max_memory_allocated (GB)` which is the maximum memory consumption reached during the run,
- `total_memory_available (GB)` which is the total memory provided by the device.

Here is an example of how the training logs look like:
```
{'loss': 0.6188, 'learning_rate': 2.9130434782608696e-05, 'epoch': 0.09, 'memory_allocated (GB)': 3.82, 'max_memory_allocated (GB)': 6.32, 'total_memory_available (GB)': 29.85}
```

## Before submitting


philschmid:
Have you benchmarked how much this slows down the training? 

Wouldn't it be enough to log before the training starts? or does the memory usage change?

philschmid:
Also what do you think about a utility method where you could pass in your mode, dataset & args to get this value back?
,e.g. `Trainer.calculate_memory_usage` or something
 
=> This would make it easy first to find the correct config and then start the training process.

regisss:
> Have you benchmarked how much this slows down the training?
> 
> Wouldn't it be enough to log before the training starts? or does the memory usage change?

Yes I did a benchmark and didn't notice any change in training speeds (the regression tests I launched all passed, I can run all of them during the weekend if needed).

Indeed, the memory usage does change over time and especially after the first 2 steps where compilation takes place in lazy mode. @libinta Do you know when the peak memory consumption is supposed to happen?

regisss:
> Also what do you think about a utility method where you could pass in your mode, dataset & args to get this value back? ,e.g. `Trainer.calculate_memory_usage` or something
> 
> => This would make it easy first to find the correct config and then start the training process.

It would be interesting, but we would need to preprocess the dataset and run a few training steps before returning this information. It may take some time.

regisss:
> > It would be interesting, but we would need to preprocess the dataset and run a few training steps before returning this information. It may take some time.
> 
> Yeah we could simply overwrite `max_steps` and run for like 100 steps and return back the results. Feels like a nicer UX than running `.train` to check utilization.

Why not! Actually we could probably have `max_steps` not larger than 10 I think.
Then, running 10 steps in lazy mode with ALBERT XXL for instance will take at least 3-4 minutes.

philschmid:
> Then, running 10 steps in lazy mode with ALBERT XXL for instance will take at least 3-4 minutes.

Sounds like a good UX. That way a DS/MLE could "evaluate" the HW setup before starting `.train()`. This could even be automated in the future to find the max BS & gradient accumulation step.

libinta (commented):
@regisss  did you see any performance degradation by adding memory status tracking?

regisss:
> @regisss did you see any performance degradation by adding memory status tracking?

No, I didn't notice any performance degradation. But I'm going to change the feature following the discussion with Phillip above.
Libin, do you know after how many steps the peak memory consumption is reached during training? It does not seem to happen in the first two steps where graph compilation takes place.

libinta (commented):
@regisss  should we have configuration to enable/disable this track?Also peak memory is model dependent. And there is no specific time.

libinta (approved):

regisss:
@philschmid The method to estimate memory consumption by launching a few short runs with different batch sizes is a bit more complex to implement properly than I thought. There are reproducibility issues because we need to reset the RNG and the dataloaders. I propose to add memory stats to the training logs in this PR and I'm going to open a feature request for estimating memory consumptions for various batch sizes so that I do it in another PR.

philschmid (approved):

regisss:
@libinta Memory stats are now displayed in the logs every `logging_steps` training steps. I gave an example of how it looks like in the description of this PR. There is no argument to turn it on/off, it is automatically done when logging (the same way as DeepSpeed).

PR #141, opened 2022-12-11
Upgrade to SynapseAI v1.7.1

regisss:
# What does this PR do?

With the new release of SynapseAI v1.7.1, this PR updates the CI Docker images and the DeepSpeed version to install.

## Before submitting

### DIFF
--- a/.github/workflows/fast_tests.yml
+++ b/.github/workflows/fast_tests.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       AWS_REGION: us-east-1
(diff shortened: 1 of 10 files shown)


regisss:
I executed fast tests on an AWS instance and they all passed.

regisss:
For the regression tests, one of them failed: ViT multi card (see [here](https://github.com/huggingface/optimum-habana/actions/runs/3669623731)). My baseline was 80s and I now get around 86s. This run is a bit different from others because evaluation is performed at the end of each epoch. So not sure what to think of this, especially since all other regression tests passed (including ViT 1x).

libinta (approved):

PR #151, opened 2023-01-18
Add support for inference through HPU graphs in GaudiTrainer

regisss:
# What does this PR do?

This PR adds support for inference through [HPU graphs](https://docs.habana.ai/en/latest/PyTorch/Inference_on_Gaudi/Inference_using_HPU_Graphs/Inference_using_HPU_Graphs.html) in `GaudiTrainer`. This is used for prediction/evaluation outside of training.

- [x] Integrate HPU graphs to `GaudiTrainer`
- [x] Integrate HPU graphs to `GaudiTrainerSeq2Seq`
- [x] Documentation (mention warmup steps)
- [x] Add the `--use_hpu_graphs` flag to all examples in the GH repo and in the documentation
- [x] Pass regression tests
- [x] Add unit tests

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

PR #179, opened 2023-03-03
Enable HPU graphs for distributed runs and generation

regisss:
# What does this PR do?

This PR follows #151 and enables HPU graphs for distributed runs and text generation.

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -1100,6 +1100,7 @@ class GaudiTrainer(Trainer):
         # Do not use HPU graphs if the training is ongoing because it detaches gradients
         if args.use_hpu_graphs and not self.is_in_train:
             logger.info("Using HPU graphs for inference.")
+            # Do not wrap the model in HPU graphs if it has already been done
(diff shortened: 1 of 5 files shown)


optimum/habana/transformers/trainer.py
```diff
@@ -1099,20 +1099,20 @@ def evaluation_loop(
...
-                )
+            logger.info("Using HPU graphs for inference.")
+            # Do not wrap the model in HPU graphs if it has already been done
+            if not self.already_wrapped_for_hpu_graphs:
+                # TODO: delete the three following lines when SynapseAI 1.9 is released
+                from transformers.models.t5.modeling_t5 import T5Attention
+
+                from .models.t5 import _gaudi_relative_position_bucket
+
+                T5Attention._relative_position_bucket = _gaudi_relative_position_bucket
```
libinta, line 1110: these are specific for T5 model, but the function is generic, do you need to check if you are running with specific model?

regisss: No because the function is updated only in the class `T5Attention`. So other models won't be affected.

libinta (approved):

PR #183, opened 2023-03-21
Update README.md

hsubramony:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ pip install git+https://github.com/huggingface/optimum-habana.git
 > ```bash
 > git clone https://github.com/huggingface/optimum-habana.git
 > cd optimum-habana
-> python setup.py install
+> pip install -e .
 > ```
 
 Last but not least, don't forget to install the requirements for every example:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,7 +53,7 @@ pip install git+https://github.com/huggingface/optimum-habana.git
 > ```bash
 > git clone https://github.com/huggingface/optimum-habana.git
 > cd optimum-habana
-> python setup.py install
+> pip install -e .
 > ```
```
regisss, line 57: I would actually remove the whole comment from line 52 to line 57. `python setup.py install` doesn't work while `pip install` does.

regisss (commented):
Hi @hsubramony! Thanks for this PR. As stated in the comment, I think we should just remove this part of the README.

README.md
hsubramony: got it ..will do

regisss (approved):
LGTM, thanks!

PR #188, opened 2023-03-28
Adding dataloader_num_workers into example command for better performance

ZhaiFeiyue:
# What does this PR do?

Add dataloader_num_workers  into example command line for ViT and Swin for better performance
here is the perfomrance data of ViT I measured in my local machine .

- Gaudi2

| dataloader_num_workers | 8X  |  1X  |
|--|--|--|
|0 (default)| 2040 | 508 |
|1| 3954| 797|
|2| 3894| 870|
|4| 3896| 985|
|8| 3831| 948|

-Gaudi
| dataloader_num_workers | 8X  |  1X  |
|--|--|--|
|0 (default) | 1605| 315|
|1| 2008| 316|
|2| 2030| 318|
|4| 1979| 332|
|8| 1911| 316|

```bash
python run_image_classification.py --model_name_or_path google/vit-base-patch16-224-in21k --dataset_name cifar10 --output_dir /tmp/outputs/  --remove_unused_columns False --do_train  --learning_rate 3e-5 --num_train_epochs 1 --per_device_train_batch_size 64 --evaluation_strategy no --save_strategy no --load_best_model_at_end no --save_total_limit 3 --seed 1337 --use_habana --use_lazy_mode --use_hpu_graphs --throughput_warmup_steps 2 --overwrite_output_dir --logging_steps 20 --gaudi_config_name Habana/vit --dataloader_num_workers xxx
```

If adding the `--dataloader_num_workers` in the  example command, user could have better performance by default. 
From the performance data I measured, we could see adding `dataloader_num_workers` not hurt performance of Gaudi but improved ~2x on Gaudi2.

profile of time cost of dataloader(Gaudi2)
berfore: 90ms
  
image

after set the `dataloader_num_workers`: 5.2ms

image

## Before submitting

### DIFF
--- a/examples/image-classification/README.md
+++ b/examples/image-classification/README.md
@@ -47,6 +47,7 @@ python run_image_classification.py \
     --use_hpu_graphs \
     --gaudi_config_name Habana/vit \
     --throughput_warmup_steps 2
+    --dataloader_num_workers 1
 ```
 
 For Swin, you need to change/add the following arguments:
@@ -94,6 +95,7 @@ python run_image_classification.py \
     --use_hpu_graphs \
     --gaudi_config_name Habana/vit \
     --throughput_warmup_steps 2
+    --dataloader_num_workers 1
 ```
 
 Internally, the script will use the [`ImageFolder`](https://huggingface.co/docs/datasets/v2.0.0/en/image_process#imagefolder) feature which will automatically turn the folders into 🤗 Dataset objects.
@@ -194,6 +196,7 @@ python ../gaudi_spawn.py \
     --use_hpu_graphs \
     --gaudi_config_name Habana/vit \
     --throughput_warmup_steps 2
+    --dataloader_num_workers 1
 ```
 
 For Swin, you need to change/add the following arguments:
@@ -232,6 +235,7 @@ python ../gaudi_spawn.py \
     --gaudi_config_name Habana/vit \
     --throughput_warmup_steps 2 \
     --deepspeed path_to_my_deepspeed_config
+    --dataloader_num_workers 1
 ```
 
 You can look at the [documentation](https://huggingface.co/docs/optimum/habana/usage_guides/deepspeed) for more information about how to use DeepSpeed in Optimum Habana.
--- a/tests/baselines/swin_base_patch4_window7_224_in22k.json
+++ b/tests/baselines/swin_base_patch4_window7_224_in22k.json
@@ -12,7 +12,8 @@
                 "extra_arguments": [
                     "--remove_unused_columns False",
                     "--seed 1337",
-                    "--ignore_mismatched_sizes"
+                    "--ignore_mismatched_sizes",
+                    "--dataloader_num_workers 1"
                 ]
             },
             "multi_card": {
@@ -24,7 +25,8 @@
                 "extra_arguments": [
                     "--remove_unused_columns False",
                     "--seed 1337",
-                    "--ignore_mismatched_sizes"
+                    "--ignore_mismatched_sizes",
+                    "--dataloader_num_workers 1"
                 ]
             }
         }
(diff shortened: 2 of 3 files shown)


ZhaiFeiyue:
@regisss Could you have a quick look on this PR? This PR is aims to improve the default performance of CV tasks(such as ViT and Swin), since the workers of dataloader is very sensitive for these tasks.

regisss (commented):
Hi @ZhaiFeiyue! Thanks for this nice PR :slightly_smiling_face: 

It looks good to me. Could you also add the `--dataloader_num_workers 1` argument [here](https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/vit_base_patch16_224_in21k.json) and [there](https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/swin_base_patch4_window7_224_in22k.json) in the `extra_arguments` field for both single- and multi-card runs please? That way regression tests for ViT and Swin will also be run with this speedup. Don't worry about the throughput numbers, I'll update them.

Also, this only benefits CV models right? There is no such speedup with text or audio models?

ZhaiFeiyue:
@regisss  will follow your comments.
But I have one question, I notice that the benchmark with [config](https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/vit_base_patch16_224_in21k.json)
is the E2E performance which means include training and evaluation and checkpoints saving and loading. 
But the performance data I measured is the training(focus on the fwd, bwd and weights update) and removed evaluation and checkpoint saving and loading with the args `--evaluation_strategy no --save_strategy no --load_best_model_at_end no `, is this a correct way or should I follow the [config](https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/vit_base_patch16_224_in21k.json) to do  the benchmark.

regisss:
> @regisss will follow your comments. But I have one question, I notice that the benchmark with [config](https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/vit_base_patch16_224_in21k.json) is the E2E performance which means include training and evaluation and checkpoints saving and loading. But the performance data I measured is the training only (focus on the fwd, bwd and weights update) and removed evaluation and checkpoint saving and loading with the args `--evaluation_strategy no --save_strategy no --load_best_model_at_end no `, is this a correct way or should I follow the [config](https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/vit_base_patch16_224_in21k.json) to do the benchmark.

@ZhaiFeiyue I tried a quick run and I saw a speedup even using the benchmark config (so with checkpoint saving and evaluation at the end of each epoch). Which makes sense, since saving the model should not be affected by this change and evaluation should also get some speedup since `dataloader_num_workers` is also used for the evaluation dataloader. So no worries regarding this.

regisss:
> > Hi @ZhaiFeiyue! Thanks for this nice PR slightly_smiling_face
> > It looks good to me. Could you also add the `--dataloader_num_workers 1` argument [here](https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/vit_base_patch16_224_in21k.json) and [there](https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/swin_base_patch4_window7_224_in22k.json) in the `extra_arguments` field for both single- and multi-card runs please? That way regression tests for ViT and Swin will also be run with this speedup. Don't worry about the throughput numbers, I'll update them.
> > Also, this only benefits CV models right? There is no such speedup with text or audio models?
> 
> it depend on the workload on the dataloader, if the workload is heavy the dataloader_num_workers will help, but we could always set it to 1 at least, which also means the workload on the dataloader will be another thread won't block the main thread. For text model like NLP(transformer based gpt model), the workload on the dataloader is not heavy, so I think the performance won't be improved too much.

Okay I see, thanks for the explanation :slightly_smiling_face: 
I'll try it with other modalities when I have some time.

regisss (approved):
Nice contribution to speedup CV models @ZhaiFeiyue :fire: :rocket:

PR #189, opened 2023-03-28
add mark_step between fwd and bwd for better performance

ZhaiFeiyue:
# What does this PR do?

improve Habana HPU performance by adding a new command line args `pipelining_fwd_bwd ` .

when running model on Habana HPU, it is very important to pipelining host graph building and device computing, this PR is to add an additional mark_step between `forward` and `backward` in `training_step` to pipeline the host graph building of backward and device computing of forward. 

profile is bellow:

disable `pipelining_fwd_bwd `
device fwd is after host bwd
image

enable `pipelining_fwd_bwd `
host bwd and device fwd is pipelined
image

the final time cost saving is the `cost of host bwd building graph`

here is the performance measured on my local Gaudi2 and Gaudi machine

-gaudi2

| pipelining_fwd_bwd        | 8x        | 1x        |
|                                        -|           -|           -|
|   True                               |     4849                   |             834     |
|   False                              |      3954                 |                  784|

-gaudi

| pipelining_fwd_bwd        | 8x        | 1x        |
|                                        -|           -|           -|
|   False|     2008|        316|
|   True|      2238|                  342|

On Gaudi2 8X performance improved from 3900 to 4800 but on Gaudi there is only 200 improved, the reason is the computing on Gaudi2 is faster than Gaudi(~3X-4X), and the device time cost of Gaudi2 is less than Gaudi, so when pipelining device fwd and host bwd, there is more benefits on Gaudi2.

test command:
```bash
# enable mark_step between fwd and bwd

# 8x
python ../gaudi_spawn.py     --world_size 8 --use_mpi run_image_classification.py     --model_name_ or_path google/vit-base-patch16-224-in21k --dataset_name cifar10     --output_dir /tmp/outputs/  --remove_unused_columns False     --do_train       --num_train_epochs 5     --per_device_train_batch_size 64     - -evaluation_strategy no     - -save_strategy no --load_best_model_at_end False     --save_total_limit 3 --seed 1337     --use_habana     --use_lazy_mode     --use_hpu_graphs     --gaudi_config_name Habana/vit --throughput_warmup_steps 2 --logging_steps 20 --dataloader_num_workers 1 --learning_rate 3.3e-4  --overwrite_output_dir --pipelining_fwd_bwd True

# 1x
python run_image_classification.py --model_name_or_path google/vit-base-patch16-224-in21k --dataset_name cifar10 --output_dir /tmp/outputs/  --remove_unused_columns False --do_train  --learning_rate 3e-5 --num_train_epochs 1 --per_device_train_batch_size 64 --evaluation_strategy no --save_strategy no --load_best_model_at_end no --save_total_limit 3 --seed 1337 --use_habana --use_lazy_mode --use_hpu_graphs --throughput_warmup_steps 2 --overwrite_output_dir --logging_steps 20 --gaudi_config_name Habana/vit --dataloader_num_workers 1 --pipelining_fwd_bwd True

# disable mark_step between fwd and bwd(default)

# 8x
python ../gaudi_spawn.py     --world_size 8 --use_mpi run_image_classification.py     --model_name_ or_path google/vit-base-patch16-224-in21k --dataset_name cifar10     --output_dir /tmp/outputs/  --remove_unused_columns False     --do_train       --num_train_epochs 5     --per_device_train_batch_size 64     - -evaluation_strategy no     - -save_strategy no --load_best_model_at_end False     --save_total_limit 3 --seed 1337     --use_habana     --use_lazy_mode     --use_hpu_graphs     --gaudi_config_name Habana/vit --throughput_warmup_steps 2 --logging_steps 20 --dataloader_num_workers 1 --learning_rate 3.3e-4  --overwrite_output_dir --pipelining_fwd_bwd False

# 1x
python run_image_classification.py --model_name_or_path google/vit-base-patch16-224-in21k --dataset_name cifar10 --output_dir /tmp/outputs/  --remove_unused_columns False --do_train  --learning_rate 3e-5 --num_train_epochs 1 --per_device_train_batch_size 64 --evaluation_strategy no --save_strategy no --load_best_model_at_end no --save_total_limit 3 --seed 1337 --use_habana --use_lazy_mode --use_hpu_graphs --throughput_warmup_steps 2 --overwrite_output_dir --logging_steps 20 --gaudi_config_name Habana/vit --dataloader_num_workers 1 --pipelining_fwd_bwd False
```

## Before submitting

### DIFF
--- a/docs/source/usage_guides/accelerate_training.mdx
+++ b/docs/source/usage_guides/accelerate_training.mdx
@@ -92,6 +92,28 @@ If these operators are not specified, their default values are set to be the one
 ]
 ```
 
+## Pipelining forward and backward passes
+
+There are two stages when running models on Habana HPU: python code interpretation on CPU and HPU recipe computation. The HPU computation stage can be triggered manually or when a copy to the CPU is requested, and generally HPU computation is triggered after `loss.backward()` to make the CPU code interpretation and HPU recipe computation overlap as shown in the following illustration:
+
+```
+CPU:...forward + backward   ...optimizer  ...forward + backward   ...optimizer  ...
+HPU:........................forward + backward...optimizer......forward + backward...optimizer
+```
+
+However, when CPU code interpretation takes longer than HPU computation, it becomes the bottleneck and HPU computation can not be triggered until CPU code interpretation is done. So one potential optimization for such cases is to trigger the HPU *forward* computation right after the CPU *forward* interpretation and before the CPU *backward* interpretation. You can see an example below where the CPU *backward* interpretation overlaps with the HPU *forward* computation:
+
+```
+CPU:...forward   ...backward   ...optimizer  ...forward   ...backward   ...optimizer   ...
+HPU:.............forward.......backward......optimizer......forward.....backward.......optimizer
+```
+
+To enable this optimization,you can set the following training argument `--pipelining_fwd_bwd True`.
+
+### Recommendations
+
+- When training models on Habana Gaudi2, we suggest to enable this feature because Gaudi2 is faster than Gaudi1.
+- When training models that require large device memory, we suggest to disable this optimization because it will increase the HPU memory usage.
 
 ## Custom Operators
 
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -499,6 +499,50 @@ class GaudiTrainer(Trainer):
             ignore_keys_for_eval=ignore_keys_for_eval,
         )
 
+    def training_step(self, model: torch.nn.Module, inputs: Dict[str, Union[torch.Tensor, Any]]) -> torch.Tensor:
+        """
+        Perform a training step on a batch of inputs.
+
+        Subclass and override to inject custom behavior.
+
+        Args:
+            model (`torch.nn.Module`):
+                The model to train.
+            inputs (`Dict[str, Union[torch.Tensor, Any]]`):
+                The inputs and targets of the model.
+
+                The dictionary will be unpacked before being fed to the model. Most models expect the targets under the
+                argument `labels`. Check your model's documentation for all accepted arguments.
+
+        Return:
+            `torch.Tensor`: The tensor with training loss on this batch.
+        """
+        model.train()
+        inputs = self._prepare_inputs(inputs)
+
+        with self.compute_loss_context_manager():
+            loss = self.compute_loss(model, inputs)
+
+        if self.args.n_gpu > 1:
+            loss = loss.mean()  # mean() to average on multi-gpu parallel training
+
+        if self.args.gradient_accumulation_steps > 1 and not self.deepspeed:
+            # deepspeed handles loss scaling by gradient_accumulation_steps in its `backward`
+            loss = loss / self.args.gradient_accumulation_steps
+
+        if self.args.pipelining_fwd_bwd:
+            self.htcore.mark_step()
+
+        if self.do_grad_scaling:
+            self.scaler.scale(loss).backward()
+        elif self.deepspeed:
+            # loss gets scaled under gradient_accumulation_steps in deepspeed
+            loss = self.deepspeed.backward(loss)
+        else:
+            loss.backward()
+
+        return loss.detach()
+
     def _inner_training_loop(
         self,
         batch_size=None,
@@ -1322,7 +1366,7 @@ class GaudiTrainer(Trainer):
         Perform an evaluation step on `model` using `inputs`.
         Subclass and override to inject custom behavior.
         Args:
-            model (`nn.Module`):
+            model (`torch.nn.Module`):
                 The model to evaluate.
             inputs (`Dict[str, Union[torch.Tensor, Any]]`):
                 The inputs and targets of the model.
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -142,6 +142,16 @@ class GaudiTrainingArguments(TrainingArguments):
         },
     )
 
+    pipelining_fwd_bwd: Optional[bool] = field(
+        default=False,
+        metadata={
+            "help": (
+                "Whether to add an additional mark_step between forward and backward for pipelining "
+                "host BWD building and HPU FWD computing."
+            )
+        },
+    )
+
     def __post_init__(self):
         if (self.use_lazy_mode or self.use_hpu_graphs or self.gaudi_config_name) and not self.use_habana:
             raise ValueError(
--- a/tests/baselines/swin_base_patch4_window7_224_in22k.json
+++ b/tests/baselines/swin_base_patch4_window7_224_in22k.json
@@ -13,7 +13,8 @@
                     "--remove_unused_columns False",
                     "--seed 1337",
                     "--ignore_mismatched_sizes",
-                    "--dataloader_num_workers 1"
+                    "--dataloader_num_workers 1",
+                    "--pipelining_fwd_bwd True"
                 ]
             },
             "multi_card": {
@@ -26,7 +27,8 @@
                     "--remove_unused_columns False",
                     "--seed 1337",
                     "--ignore_mismatched_sizes",
-                    "--dataloader_num_workers 1"
+                    "--dataloader_num_workers 1",
+                    "--pipelining_fwd_bwd True"
                 ]
             }
         }
--- a/tests/baselines/vit_base_patch16_224_in21k.json
+++ b/tests/baselines/vit_base_patch16_224_in21k.json
@@ -12,7 +12,8 @@
                 "extra_arguments": [
                     "--remove_unused_columns False",
                     "--seed 1337",
-                    "--dataloader_num_workers 1"
+                    "--dataloader_num_workers 1",
+                    "--pipelining_fwd_bwd True"
                 ]
             },
             "multi_card": {
@@ -24,7 +25,8 @@
                 "extra_arguments": [
                     "--remove_unused_columns False",
                     "--seed 1337",
-                    "--dataloader_num_workers 1"
+                    "--dataloader_num_workers 1",
+                    "--pipelining_fwd_bwd True"
                 ]
             }
         }


ZhaiFeiyue:
@regisss this is the patch for ViT and Swin

optimum/habana/transformers/trainer.py
```diff
@@ -53,6 +54,7 @@
     nested_numpify,
     nested_truncate,
     reissue_pt_warnings,
+    is_sagemaker_mp_enabled,
```
regisss, line 57: You can remove this, we don't need it in Optimum Habana

optimum/habana/transformers/trainer.py
```diff
@@ -88,6 +90,15 @@
...
 
+if is_sagemaker_mp_enabled():
+    import smdistributed.modelparallel.torch as smp
+    from smdistributed.modelparallel import __version__ as SMP_VERSION
+
+    IS_SAGEMAKER_MP_POST_1_10 = version.parse(SMP_VERSION) >= version.parse("1.10")
+
+    from .trainer_pt_utils import smp_forward_backward, smp_forward_only, smp_gather, smp_nested_concat
+else:
+    IS_SAGEMAKER_MP_POST_1_10 = False
```
regisss, line 101: Same here

optimum/habana/transformers/trainer.py
```diff
@@ -483,6 +494,58 @@ def train(
...
+
+        Return:
+            `torch.Tensor`: The tensor with training loss on this batch.
+        """
+        model.train()
+        inputs = self._prepare_inputs(inputs)
+
+        if is_sagemaker_mp_enabled():
+            loss_mb = smp_forward_backward(model, inputs, self.args.gradient_accumulation_steps)
+            return loss_mb.reduce_mean().detach().to(self.args.device)
```
regisss, line 520: Same here, you can remove this block

optimum/habana/transformers/trainer.py
```diff
@@ -483,6 +494,58 @@ def train(
...
+            loss = loss / self.args.gradient_accumulation_steps
+
+        if self.args.pipelining_fwd_bwd:
+            self.htcore.mark_step()
+
+        if self.do_grad_scaling:
+            self.scaler.scale(loss).backward()
+        elif self.use_apex:
+            with amp.scale_loss(loss, self.optimizer) as scaled_loss:
+                scaled_loss.backward()
```
regisss, line 539: APEX-related code can be removed also

optimum/habana/transformers/trainer.py
```diff
@@ -26,6 +26,7 @@
 
 import numpy as np
 import torch
+from torch import nn
```
regisss, line 29: Not a big fan of importing `nn`, let's just use `torch.nn` below

optimum/habana/transformers/trainer.py
```diff
@@ -483,6 +494,58 @@ def train(
             ignore_keys_for_eval=ignore_keys_for_eval,
         )
 
+    def training_step(self, model: nn.Module, inputs: Dict[str, Union[torch.Tensor, Any]]) -> torch.Tensor:
```
regisss, line 497:
```suggestion
    def training_step(self, model: torch.nn.Module, inputs: Dict[str, Union[torch.Tensor, Any]]) -> torch.Tensor:
```

optimum/habana/transformers/trainer.py
```diff
@@ -483,6 +494,58 @@ def train(
...
+        """
+        Perform a training step on a batch of inputs.
+
+        Subclass and override to inject custom behavior.
+
+        Args:
+            model (`nn.Module`):
```
regisss, line 504:
```suggestion
            model (`torch.nn.Module`):
```

regisss (commented):
Hi @ZhaiFeiyue! Very interesting PR :slightly_smiling_face: 

I left a couple of preliminary comments, I'll run a quick benchmark next week when I'm back from vacation.

Could you also add the pipelining argument to the [image-classification README](https://github.com/huggingface/optimum-habana/blob/main/examples/image-classification/README.md) as you did in the other PR with `--dataloader_num_workers` please?

Besides, same question as in the other PR, does this only speed up ViT and Swin? Or could it work with text and/or audio models too?

ZhaiFeiyue:
> Hi @ZhaiFeiyue! Very interesting PR 🙂
> 
> I left a couple of preliminary comments, I'll run a quick benchmark next week when I'm back from vacation.
> 
> Could you also add the pipelining argument to the [image-classification README](https://github.com/huggingface/optimum-habana/blob/main/examples/image-classification/README.md) as you did in the other PR with `--dataloader_num_workers` please?
> 
> Besides, same question as in the other PR, does this only speed up ViT and Swin? Or could it work with text and/or audio models too?

@regisss have a nice vocation. I will follow your comments to update this patch.

For your question, I think this patch works for the most of models when running on Gaudi2, because the Gaudi2 is faster than Gaudi, so the host graph building will be the bottleneck, but not works for the device bounding model(Gaudi is slower than Gaudi2, so when running models on Gaudi, it is a HPU bounding case, that's why not works).

another important point is that: the final performance improved is also depend on the BS, means when increase the BS, the workload on the device is also increased, so it could be a device bounding case(device is always busy) , so we break the FWD and BWD to pipelining host and device won't work.

And break the FWD and BWD not like `dataloader_num_workers` always bring positive values to the model at least not hurt the performance, one  short things of `pipelining argument ` is more device memory usage, because the activation of FWD if used by the BWD will be the graph IO of the FWD graph and BWD graph, the graph IO is the persistent memory not like the activation can't be resued.

For text and audio models it depend on the workload between host graph building and device computing and the model itself, so what I think is to expose this flag so user could have a try without change the source code.

optimum/habana/transformers/trainer.py
ZhaiFeiyue: removed

optimum/habana/transformers/trainer.py
ZhaiFeiyue: removed

optimum/habana/transformers/trainer.py
ZhaiFeiyue: removed

optimum/habana/transformers/trainer.py
ZhaiFeiyue: removed

optimum/habana/transformers/trainer.py
ZhaiFeiyue: updated

optimum/habana/transformers/trainer.py
ZhaiFeiyue: done

optimum/habana/transformers/trainer.py
ZhaiFeiyue: done

ZhaiFeiyue:
@regisss I have removed the code not used for optimum-habana, and also change nn.xxx to torch.nn.xxx
but regarding to `adding the pipelining argument into Readme.md`, I have a concern here, because `pipelining argument` has side effect(consumes more persistent memory), and which is a advanced feature for the user who is familiar with Habana HPU not for the beginner. That's why I only expose this argument to user but not adding it to example by default. 
how do you think, if you think it's fine that we add `pipelining argument`  into Readme.md, I could update this patch.

regisss:
> > Hi @ZhaiFeiyue! Very interesting PR slightly_smiling_face
> > I left a couple of preliminary comments, I'll run a quick benchmark next week when I'm back from vacation.
> > Could you also add the pipelining argument to the [image-classification README](https://github.com/huggingface/optimum-habana/blob/main/examples/image-classification/README.md) as you did in the other PR with `--dataloader_num_workers` please?
> > Besides, same question as in the other PR, does this only speed up ViT and Swin? Or could it work with text and/or audio models too?
> 
> @regisss have a nice vocation. I will follow your comments to update this patch.
> 
> For your question, I think this patch works for the most of models when running on Gaudi2, because the Gaudi2 is faster than Gaudi, so the host graph building will be the bottleneck, but not works for the device bounding model(Gaudi is slower than Gaudi2, so when running models on Gaudi, it is a HPU bounding case, that's why not works).
> 
> another important point is that: the final performance improved is also depend on the BS, means when increase the BS, the workload on the device is also increased, so it could be a device bounding case(device is always busy) , so we break the FWD and BWD to pipelining host and device won't work.
> 
> And break the FWD and BWD not like `dataloader_num_workers` always bring positive values to the model at least not hurt the performance, one short things of `pipelining argument ` is more device memory usage, because the activation of FWD if used by the BWD will be the graph IO of the FWD graph and BWD graph, the graph IO is the persistent memory not like the activation can't be resued.
> 
> For text and audio models it depend on the workload between host graph building and device computing and the model itself, so what I think is to expose this flag so user could have a try without change the source code.

I see , thanks for the detailed explanation :slightly_smiling_face:

regisss:
> @regisss I have removed the code not used for optimum-habana, and also change nn.xxx to torch.nn.xxx but regarding to `adding the pipelining argument into Readme.md`, I have a concern here, because `pipelining argument` has side effect(consumes more persistent memory), and which is a advanced feature for the user who is familiar with Habana HPU not for the beginner. That's why I only expose this argument to user but not adding it to example by default. how do you think, if you think it's fine that we add `pipelining argument` into Readme.md, I could update this patch.

@ZhaiFeiyue It makes sense, let's keep it that way then. So I suggest documenting this feature here: https://huggingface.co/docs/optimum/habana/usage_guides/accelerate_training
You can update this markdown file to update the documentation: https://github.com/huggingface/optimum-habana/blob/main/docs/source/usage_guides/accelerate_training.mdx
I think you can just create a new section called "Pipelining forward and backward passes" with the explanation you wrote above.

regisss:
@ZhaiFeiyue Also the code quality check failed, could you run the command `make style` at the root of the repo please?

ZhaiFeiyue:
@regisss I have updated the user guide please feel free to comments, also the CI always fail,
image
after check the log, seems that is the `Credentials issue`, any ideas?

regisss:
> @regisss I have updated the user guide please feel free to comments, also the CI always fail, image after check the log, seems that is the `Credentials issue`, any ideas?

@ZhaiFeiyue Thanks for these changes, I'm going to take a look at them today!
For the CI, I noticed it was failing when you opened the PR. It happens because the branch you pushed is located in a fork, we are working on it so that PRs from forks don't encounter this issue anymore. In the meantime, I'll just clone your fork and will run the CI locally, and I'll let you know if there is any issue :+1:

docs/source/usage_guides/accelerate_training.mdx
```diff
@@ -92,6 +92,28 @@ If these operators are not specified, their default values are set to be the one
...
+## Pipelining forward and backward passes
+
+There are two stages when running models on Habana HPU: python code interpretation and HPU recipe computation. The HPU computation stage could be triggered by invoking `mark_step()` or when the copy to the CPU is requested, and generally there is a `mark_step()` after `loss.backward()` to pipelining the CPU code interpretation and HPU recipe computation see the following illustration:
+
+```
+CPU:...forward + backward   ...optimizer  ...forward + backward   ...optimizer  ...
+HPU:........................forward + backward...optimizer......forward + backward...optimizer
+```
+
+But when the time cost of HPU computation is less than the CPU code interpretation, which means CPU code interpretation is the bottleneck, and the HPU computation can not be triggered untill long time cost of CPU code interpretation. So one potential optimization for such kind of case is adding an addtianal `mark_step()` after CPU forward before CPU backward, see bellow, to pipelining CPU backward code interpretation and device forward computation:
```
regisss, line 104:
```suggestion
However, when CPU code interpretation takes longer than HPU computation, it becomes the bottleneck and HPU computation can not be triggered until CPU code interpretation is done. So one potential optimization for such cases is to trigger the HPU *forward* computation right after the CPU *forward* interpretation and before the CPU *backward* interpretation. You can see an example below where the CPU *backward* interpretation overlaps with the HPU *forward* computation:
```

regisss (commented):
Thnaks for improving the documentation @ZhaiFeiyue :fire:  We are almost there!

I have a few comments to make the doc a bit simpler. We don't expect users to necessarily know about `mark_step` so I tried to rephrase it.

Also, is it possible to add the argument `--pipelining_fwd_bwd True` to https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/vit_base_patch16_224_in21k.json and https://github.com/huggingface/optimum-habana/blob/main/tests/baselines/swin_base_patch4_window7_224_in22k.json please?

docs/source/usage_guides/accelerate_training.mdx
```diff
@@ -92,6 +92,28 @@ If these operators are not specified, their default values are set to be the one
 ]
 ```
 
+## Pipelining forward and backward passes
+
+There are two stages when running models on Habana HPU: python code interpretation and HPU recipe computation. The HPU computation stage could be triggered by invoking `mark_step()` or when the copy to the CPU is requested, and generally there is a `mark_step()` after `loss.backward()` to pipelining the CPU code interpretation and HPU recipe computation see the following illustration:
```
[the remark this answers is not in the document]
ZhaiFeiyue: done

docs/source/usage_guides/accelerate_training.mdx
```diff
@@ -92,6 +92,28 @@ If these operators are not specified, their default values are set to be the one
...
+```
+
+But when the time cost of HPU computation is less than the CPU code interpretation, which means CPU code interpretation is the bottleneck, and the HPU computation can not be triggered untill long time cost of CPU code interpretation. So one potential optimization for such kind of case is adding an addtianal `mark_step()` after CPU forward before CPU backward, see bellow, to pipelining CPU backward code interpretation and device forward computation:
+
+```
+CPU:...forward   ...backward   ...optimizer  ...forward   ...backward   ...optimizer   ...
+HPU:.............forward.......backward......optimizer......forward.....backward.......optimizer
+```
+
+To enable this optimization just setting training argument `--pipelining_fwd_bwd True`.
```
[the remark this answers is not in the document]
ZhaiFeiyue: done

docs/source/usage_guides/accelerate_training.mdx
```diff
@@ -92,6 +92,28 @@ If these operators are not specified, their default values are set to be the one
...
+
+### Limitation
```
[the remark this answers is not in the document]
ZhaiFeiyue: done

docs/source/usage_guides/accelerate_training.mdx
```diff
@@ -92,6 +92,28 @@ If these operators are not specified, their default values are set to be the one
...
+
+- when running models on Habana Gaudi2, we suggest to enable this feature because Gaudi2 is faster than Gaudi1.
+- when running models that requires large device memory, we suggest to disable this optimization, because this optimization will increase the HPU memory usage.
```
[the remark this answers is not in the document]
ZhaiFeiyue: done

docs/source/usage_guides/accelerate_training.mdx
```diff
@@ -92,6 +92,28 @@ If these operators are not specified, their default values are set to be the one
...
+```
+CPU:...forward   ...backward   ...optimizer  ...forward   ...backward   ...optimizer   ...
+HPU:.............forward.......backward......optimizer......forward.....backward.......optimizer
+```
+
+To enable this optimization just setting training argument `--pipelining_fwd_bwd True`.
+
+### Limitation
+
+- when running models on Habana Gaudi2, we suggest to enable this feature because Gaudi2 is faster than Gaudi1.
```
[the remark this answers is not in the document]
ZhaiFeiyue: done

docs/source/usage_guides/accelerate_training.mdx
ZhaiFeiyue: done

ZhaiFeiyue:
@regisss Yes, hide the `mark_steps()` for users is a good idea. And the doc looks much better than before after apply your comments 👍, I have rebased the commit and added the `--pipelining_fwd_bwd True` into the vit and swin test config. please have a look.

regisss (approved):
@ZhaiFeiyue I ran the tests locally and they all passed! Great contribution to get nice speedups with CV models :slightly_smiling_face:

PR #208, opened 2023-04-23
Enable asynchronous data copy to get better performance

jychen-habana:
# What does this PR do?
Improve Habana HPU performance, especially the host side performance by enabling 
asynchronous data copy from host to device with arg `non_blocking_data_copy`.

Before setting `non_blocking_data_copy` to True
image

After setting `non_blocking_data_copy` to True
image

Here is the overall performance before and after enabling `non_blocking_data_copy` (take Swin as an example):

Gaudi2
| non_blocking_data_copy | 1x | 8x |
|----------|:-------------:|------:|
| True | 374 | 1894 |
| False | 354 | 1857 |

Gaudi
| non_blocking_data_copy | 1x | 8x |
|----------|:-------------:|------:|
| True | 183 | 1060 |
| False | 183 | 1061 |

Both 1x and 8x performance get improvement on Gaudi2 but on Gaudi, none of them get any improvement. The reason is that the device computing on Gaudi2 is faster than Gaudi, when asynchronous data copy is enabled, the host can continue to execute other tasks like graph building, then we can get a better pipelining between the host and the device.  But on Gaudi, the device executing time is much longer which already hides the host time, so this optimization is more appropriate for Gaudi2.

Test command
```
# Single-HPU
python ../gaudi_spawn.py \
    --world_size 8 --use_mpi run_image_classification.py \
    --model_name_or_path microsoft/swin-base-patch4-window7-224-in22k \
    --dataset_name cifar10 \
    --output_dir /tmp/outputs/ \
    --remove_unused_columns False \
    --do_train \
    --learning_rate 2e-4 \
    --num_train_epochs 5 \
    --per_device_train_batch_size 64 \
    --evaluation_strategy no \
    --save_strategy no \
    --load_best_model_at_end False \
    --save_total_limit 3 \
    --seed 1337 \
    --use_habana \
    --use_lazy_mode \
    --use_hpu_graphs \
    --gaudi_config_name Habana/swin \
    --throughput_warmup_steps 2 \
    --ignore_mismatched_sizes True \
    --overwrite_output_dir \
    --dataloader_num_workers 1 \
    --non_blocking_data_copy True
```

```
# Multi-HPU
python run_image_classification.py \
    --model_name_or_path microsoft/swin-base-patch4-window7-224-in22k \
    --dataset_name cifar10 \
    --output_dir /tmp/outputs/ \
    --remove_unused_columns False \
    --do_train \
    --learning_rate 3e-5 \
    --num_train_epochs 1 \
    --per_device_train_batch_size 64 \
    --evaluation_strategy no \
    --save_strategy no \
    --load_best_model_at_end False \
    --save_total_limit 3 \
    --seed 1337 \
    --use_habana \
    --use_lazy_mode \
    --use_hpu_graphs \
    --gaudi_config_name Habana/swin \
    --throughput_warmup_steps 2 \
    --ignore_mismatched_sizes True \
    --overwrite_output_dir \
    --dataloader_num_workers 1 \
    --non_blocking_data_copy True
```

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -22,6 +22,7 @@ import sys
 import tempfile
 import time
 import warnings
+from collections.abc import Mapping
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
@@ -498,6 +499,28 @@ class GaudiTrainer(Trainer):
             trial=trial,
             ignore_keys_for_eval=ignore_keys_for_eval,
         )
+
+    def _prepare_input(self, data: Union[torch.Tensor, Any]) -> Union[torch.Tensor, Any]:
+        """
+        Prepares one `data` before feeding it to the model, be it a tensor or a nested list/dictionary of tensors.
+        Compared to Transformers, it is also possible to enable non-blocking data copy.
+        """
+        if isinstance(data, Mapping):
+            return type(data)({k: self._prepare_input(v) for k, v in data.items()})
+        elif isinstance(data, (tuple, list)):
+            return type(data)(self._prepare_input(v) for v in data)
+        elif isinstance(data, torch.Tensor):
+            kwargs = {"device": self.args.device}
+            if self.deepspeed and (torch.is_floating_point(data) or torch.is_complex(data)):
+                # NLP models inputs are int/uint and those get adjusted to the right dtype of the
+                # embedding. Other models such as wav2vec2's inputs are already float and thus
+                # may need special handling to match the dtypes of the model
+                kwargs.update({"dtype": self.args.hf_deepspeed_config.dtype()})
+            if self.args.non_blocking_data_copy:
+                return data.to(**kwargs, non_blocking=True)
+            else:
+                return data.to(**kwargs)
+        return data
 
     def training_step(self, model: torch.nn.Module, inputs: Dict[str, Union[torch.Tensor, Any]]) -> torch.Tensor:
         """
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -152,6 +152,11 @@ class GaudiTrainingArguments(TrainingArguments):
         },
     )
 
+    non_blocking_data_copy: Optional[bool] = field(
+        default=False,
+        metadata={"help": ("Whether to enable async data copy when preparing inputs.")},
+    )
+
     def __post_init__(self):
         if (self.use_lazy_mode or self.use_hpu_graphs or self.gaudi_config_name) and not self.use_habana:
             raise ValueError(
--- a/tests/baselines/swin_base_patch4_window7_224_in22k.json
+++ b/tests/baselines/swin_base_patch4_window7_224_in22k.json
@@ -14,7 +14,8 @@
                     "--seed 1337",
                     "--ignore_mismatched_sizes",
                     "--dataloader_num_workers 1",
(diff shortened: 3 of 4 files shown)


optimum/habana/transformers/training_args.py
```diff
@@ -152,6 +152,15 @@ class GaudiTrainingArguments(TrainingArguments):
         },
     )
 
+    datacopy_non_blocking: Optional[bool] = field(
```
regisss, line 155:
```suggestion
    non_blocking_data_copy: Optional[bool] = field(
```

optimum/habana/transformers/training_args.py
```diff
@@ -152,6 +152,15 @@ class GaudiTrainingArguments(TrainingArguments):
...
+        default=False,
+        metadata={
+            "help": (
+                "Whether to enable async data copy when prepare inputs."
```
regisss, line 159:
```suggestion
                "Whether to enable async data copy when preparing inputs."
```

optimum/habana/transformers/trainer.py
```diff
@@ -499,6 +500,27 @@ def train(
...
+        elif isinstance(data, (tuple, list)):
+            return type(data)(self._prepare_input(v) for v in data)
+        elif isinstance(data, torch.Tensor):
+            kwargs = {"device": self.args.device}
+            if self.deepspeed and (torch.is_floating_point(data) or torch.is_complex(data)):
+                # NLP models inputs are int/uint and those get adjusted to the right dtype of the
+                # embedding. Other models such as wav2vec2's inputs are already float and thus
+                # may need special handling to match the dtypes of the model
+                kwargs.update({"dtype": self.args.hf_deepspeed_config.dtype()})
+            if self.args.datacopy_non_blocking:
```
regisss, line 518:
```suggestion
            if self.args.non_blocking_data_copy:
```

optimum/habana/transformers/trainer.py
```diff
@@ -499,6 +500,27 @@ def train(
             ignore_keys_for_eval=ignore_keys_for_eval,
         )
 
+    def _prepare_input(self, data: Union[torch.Tensor, Any]) -> Union[torch.Tensor, Any]:
+        """
+        Prepares one `data` before feeding it to the model, be it a tensor or a nested list/dictionary of tensors.
```
regisss, line 505:
```suggestion
        Prepares one `data` before feeding it to the model, be it a tensor or a nested list/dictionary of tensors.
        Compared to Transformers, it is also possible to enable non-blocking data copy.
```

regisss (commented):
Hi @jychen-habana, nice PR to get an interesting speedup!

I've left a few comments which basically focuses on renaming the argument.

I see that the style check failed, could you run `make style` from the root of the repo please?

Also out of curiosity, I'm wondering if we should not enable this argument by default. Are there use cases where it would run slower?

jychen-habana:
> Hi @jychen-habana, nice PR to get an interesting speedup!
> 
> I've left a few comments which basically focuses on renaming the argument.
> 
> I see that the style check failed, could you run `make style` from the root of the repo please?
> 
> Also out of curiosity, I'm wondering if we should not enable this argument by default. Are there use cases where it would run slower?

Hi @regisss, thank you very much for your comments!

I updated the code. Also, I added the flag to ci tests for both ViT and Swin.

This optimization is for a model which has a high cost on copy data from the host to the device and theoretically, its performance will always be better than or at least the same as synchronous copy.

regisss:
Closing this one as it is continued in #211.

PR #214, opened 2023-04-25
Rewrite gelu_xxx function to gelu, which is a fused kernel on Gaudi

ZhaiFeiyue:
# What does this PR do?
There are many gelu implementations in HF transformers, see [here](https://github.com/huggingface/transformers/blob/main/src/transformers/activations.py#L213) and could be configured by model config(refer `activation_function` in [gpt2 model config](https://huggingface.co/gpt2/blob/main/config.json))

This PR rewrite the `activation_function` in model config  to "gelu" if is `"gelu_fast", "gelu_new", "gelu_python"`, because on Gaudi, `gelu`(`torch.nn.functional.gelu`), could be mapped to one single fused gelu kernel not several separate ops.

before:
image

after:
image

command line to reproduce:
```bash
python run_clm.py --model_name_or_path gpt2 --dataset_name wikitext --dataset_config_name wikitext-2-raw-v1 --per_device_train_batch_size 1 --per_device_eval_batch_size 4 --do_eval --output_dir /tmp/test-clm --gaudi_config_name Habana/gpt2 --use_habana --use_lazy_mode --use_hpu_graphs --throughput_warmup_steps 3 --overwrite_output_dir --max_steps 100 --use_fused_gelu True
```

## Before submitting


examples/language-modeling/run_clm.py
```diff
@@ -404,6 +404,10 @@ def main():
             config.update_from_string(model_args.config_overrides)
             logger.info(f"New config: {config}")
 
+    if hasattr(config, "activation_function"):
+        if config.activation_function in ["gelu_fast", "gelu_new", "gelu_python"] and training_args.use_fused_gelu:
+            config.activation_function = "gelu"
```
regisss, line 409: Do we want to do this only in the `run_clm.py` example?
If yes, the argument `use_fused_gelu` should be defined in the example arguments [here](https://github.com/huggingface/optimum-habana/blob/ba79326a01d05b7f5b5771bc5d1823046838fae4/examples/language-modeling/run_clm.py#L67).
If no, we should rather do it in the GaudiTrainer.

optimum/habana/transformers/training_args.py
```diff
@@ -152,6 +152,11 @@ class GaudiTrainingArguments(TrainingArguments):
         },
     )
 
+    use_fused_gelu: Optional[bool] = field(
+        default=False,
```
regisss, line 156: Shouldn't we use it by default? Or is there any use case where it should not be enabled?

regisss (commented):
@ZhaiFeiyue Cool that we can use a fused GELU kernel on Gaudi!

My main comment is about where we should apply the change: only in `run_clm.py` or in the GaudiTrainer class if can benefit other use cases?

optimum/habana/transformers/training_args.py
```diff
@@ -152,6 +152,11 @@ class GaudiTrainingArguments(TrainingArguments):
...
+        metadata={"help": ("Whether to use original pytorch gelu, which could be mapped to Gaudi fused gelu kernel")},
```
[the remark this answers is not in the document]
ZhaiFeiyue: done

ZhaiFeiyue:
@regisss yes, put the changes in the GaudiTrainer could benefit other cases, but need  more code changes(find GELU_XXX activation modules recursively and replace it to GELU, it should not be a problem🙂) because the model has been initialized in [here](https://github.com/huggingface/optimum-habana/blob/ba79326a01d05b7f5b5771bc5d1823046838fae4/examples/language-modeling/run_mlm.py#L404), and my understanding about the` GaudiTrainer`  is that `GaudiTrainer ` is used to control how to train the model not used to rewrite the model structure, model structure should be rewrite in each of `modeling_xxx.py` in the `models` or change the original [config](https://huggingface.co/gpt2/blob/main/config.json)(but we can not change the original config right?), so my current changes is rewriting the [model config](https://github.com/huggingface/optimum-habana/blob/ba79326a01d05b7f5b5771bc5d1823046838fae4/examples/language-modeling/run_mlm.py#L378) after it loaded before model created, so I put it in `run_clm.py`🙂
correct me is I'm wrong.

optimum/habana/transformers/training_args.py
ZhaiFeiyue: yes, I will set the default value to True

examples/language-modeling/run_clm.py
ZhaiFeiyue: see my comments

regisss:
> @regisss yes, put the changes in the GaudiTrainer could benefit other cases, but need more code changes(find GELU_XXX activation modules recursively and replace it to GELU, it should not be a problemslightly_smiling_face) because the model has been initialized in [here](https://github.com/huggingface/optimum-habana/blob/ba79326a01d05b7f5b5771bc5d1823046838fae4/examples/language-modeling/run_mlm.py#L404), and my understanding about the` GaudiTrainer` is that `GaudiTrainer ` is used to control how to train the model not used to rewrite the model structure, model structure should be rewrite in each of `modeling_xxx.py` in the `models` or change the original [config](https://huggingface.co/gpt2/blob/main/config.json)(but we can not change the original config right?), so my current changes is rewriting the [model config](https://github.com/huggingface/optimum-habana/blob/ba79326a01d05b7f5b5771bc5d1823046838fae4/examples/language-modeling/run_mlm.py#L378) after it loaded before model created, so I put it in the `run_clm.py`slightly_smiling_face correct me is I'm wrong.

Yes you're fully right. Let me think about it a bit more to see what the best solution is :slightly_smiling_face:

regisss:
@ZhaiFeiyue I'm going to run a quick POC with [Torch FX](https://pytorch.org/docs/1.13/fx.html), which enables to dinamically modify the modeling. Basically, right after the model is instantiated, I'll parse the graph, look for a `gelu_xxx` node and if I find one I replace it by `gelu`.

regisss:
> @ZhaiFeiyue I'm going to run a quick POC with [Torch FX](https://pytorch.org/docs/1.13/fx.html), which enables to dynamically modify the modeling. Basically, right after the model is instantiated, I'll parse the graph, look for a `gelu_xxx` node and if I find one I replace it by `gelu`.

@ZhaiFeiyue I'm advancing on this and it's looking good :slightly_smiling_face: I'll let you know when I open a PR so that you can use it in this one.

Quick question: should we also take care of [`"gelu_accurate"`](https://github.com/huggingface/transformers/blob/849367ccf741d8c58aa88ccfe1d52d8636eaf2b7/src/transformers/activations.py#L124) and [`"quick_gelu"`](https://github.com/huggingface/transformers/blob/849367ccf741d8c58aa88ccfe1d52d8636eaf2b7/src/transformers/activations.py#L90) since they don't rely on `torch.nn.functional.gelu` either?

ZhaiFeiyue:
@regisss  I will try your PR #223  tomorrow after our holiday.
The GELU implementation of SynapseAI is the one with `tanh`,  seems that we could also convert `quick_gelu`, `gelu_accurate` and `gelu_python_tan`

regisss:
> @regisss I will try your PR #223 tomorrow after our holiday. The GELU implementation of SynapseAI is the one with `tanh`, seems that we could also convert `quick_gelu`, `gelu_accurate` and `gelu_python_tan` to `gelu`

Okay I see :+1: 
The FX transformation works well with GPT2 but I have an issue with T5. Anyway, I'll post here a quick example today to show you how to use it.

regisss:
I updated the PR with a few modifications to the `GaudiTrainer` class to trace models with FX and apply the transformation before training/inference. It should show you how to use the transformation.
A few things to note:

- For whatever reason, tracing GPT2 without even using the transformation gives a better throughput on `run_clm.py`. That's very weird, I don't have a good explanation for it :slightly_frowning_face: It only happens with this model, I'll have to profile the run.
- This returns a segmentation fault with T5. Not sure if this is due to a bug in Synapse or if the transformation messed up the model.
- It seems more memory is consumed with the transformation, leading to worse performance with DeepSpeed.

So there are still a few issues to solve. I'll take more time in the next few days to see if I can get a cleaner implementation (not of the transformation itself, but rather the FX integration to the trainer). Let me know if you have any idea regarding the points mentioned above :slightly_smiling_face:

ZhaiFeiyue:
@regisss don't worry, I will try you PR #223 first and then look into the above issues with you🙂

ZhaiFeiyue:
closed and works in PR #223

PR #218, opened 2023-04-26
Extend the text-generation example to other models and to support non-DeepSpeed inference 

### LINKED ISSUE #213: Run text-generation with non-deepspeed mode
### Feature request

Current [text-generation](https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation) only support bloom & bloomz with deepspeed, but not support other generation model like gpt2, gpt-j, neox.
There also a [PR](https://github.com/huggingface/optimum-habana/pull/209) shows how to run inference in each of examples, but the inference is training evaluation not a real generation case.
so is there any possible to support the following features?
- `text-generation` support other generation model(e.g. gpt2, gpt-j)
- `text-generation` support non-deepspeed mode

### Motivation

text-generation only support bloom and bloomz, can not run text-generation with gpt2, gpt-j, neox,...

### Your contribution

submitting a PR

regisss:
# What does this PR do?

This PR extends the text-generation example:

- Generation can be performed without DeepSpeed-inference on one device (multi-HPU will come in another PR)
- Other models than BLOOM can be used

Fixes #213.

## Before submitting


examples/text-generation/checkpoint_utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def write_checkpoints_json(model_name_or_path, local_rank, checkpoints_json):
+    """
+    Dumps metadata into a JSON file for DeepSpeed-inference.
+    """
+    checkpoint_files = get_checkpoint_files(model_name_or_path, local_rank)
+    if local_rank == 0:
+        data = {"type": "BLOOM", "checkpoints": checkpoint_files, "version": 1.0}
+        json.dump(data, open(checkpoints_json, "w"))
```
ZhaiFeiyue, line 58: should we use
``` python
with open(checkpoints_json, "w") as fp:
    json.dump(data, fp)
```
since `with statement` will auto close fp

regisss: Yes! Good catch

regisss: Done

examples/text-generation/run_generation.py
```diff
@@ -149,17 +99,45 @@ def main():
...
 
-    # Initialize process(es) for DeepSpeed
-    deepspeed.init_distributed(dist_backend="hccl")
+        # Initialize process(es) for DeepSpeed
+        deepspeed.init_distributed(dist_backend="hccl")
+        logger.info("DeepSpeed is enabled.")
+    else:
+        if args.gaudi_config_name_or_path is None:
+            logger.warning(
+                "`--gaudi_config_name_or_path` was not specified so not using Habana Mixed Precision for this run."
```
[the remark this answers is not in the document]
ZhaiFeiyue: if user set `gaudi_config_name_or_path` but there is no bf16 ops in it like [GPT2 gaudi_config](https://huggingface.co/Habana/gpt2/blob/main/gaudi_config.json), the precision is still FP32 not BF16 right?

PR #223, opened 2023-05-01
Add FX transformation to replace GELU modules by fused GELU modules

regisss:
# What does this PR do?

This PR enables to replace all GELU modules by `GELUActivation()` if necessary. This relies on Torch FX and on Optimum's FX interface. It will be useful for #214.

## Before submitting


optimum/habana/fx/optimization/transformations.py
```diff
@@ -0,0 +1,64 @@
...
+    `torch.nn.functional.gelu` is mapped to a single fused GELU kernel on Gaudi so the transformed model will be faster.
+    """
+
+    def transform(self, graph_module: "GraphModule") -> "GraphModule":
+        self.original_gelu_module = None
+        for node in graph_module.graph.nodes:
+            if node.op == "call_module":
+                module = graph_module.get_submodule(node.target)
+                if isinstance(module, GELU_CLASSES_TO_TRACK):
+                    if hasattr(module, "act") and getattr(module, "act") == torch.nn.functional.gelu:
```
michaelbenayoun, line 41:
```suggestion
                    if getattr(module, "act") == torch.nn.functional.gelu:
```
Is enough no?

optimum/habana/fx/optimization/transformations.py
```diff
@@ -0,0 +1,64 @@
...
+class GeluToFusedGelu(ReversibleTransformation):
+    """
+    Transformation that replaces GELU modules by `torch.nn.functional.gelu` (i.e. `GELUActivation()`) if different.
+    `torch.nn.functional.gelu` is mapped to a single fused GELU kernel on Gaudi so the transformed model will be faster.
+    """
+
+    def transform(self, graph_module: "GraphModule") -> "GraphModule":
+        self.original_gelu_module = None
+        for node in graph_module.graph.nodes:
+            if node.op == "call_module":
```
michaelbenayoun, line 38: I think `torch.nn.functional.gelu` is more of a `call_function` node?

So you could check by doing:

```python
if node.op == "call_function" and node.target is torch.nn.functional.gelu:
```

optimum/habana/fx/optimization/transformations.py
```diff
@@ -0,0 +1,64 @@
...
+                if isinstance(module, GELU_CLASSES_TO_TRACK):
+                    if hasattr(module, "act") and getattr(module, "act") == torch.nn.functional.gelu:
+                        # GELUActivation does not necessarily rely on torch.nn.functional.gelu
+                        # If it does, no need to change anything so break the loop
+                        break
+                    parent_name, _, name = node.target.rpartition(".")
+                    parent_module = graph_module.get_submodule(parent_name)
+                    # Keep track of the original GELU module to be able to reverse the transformation
+                    if self.original_gelu_module is None:
+                        self.original_gelu_module = getattr(parent_module, name)
```
michaelbenayoun, line 49: What happens if you have multiple gelu classes? While I agree it might not happen here, your code wont work.
Instead of attaching this information to the Transformation, attach it to the Node itself:

```python
node.original_gelu_module = getattr(parent_module, name)
```

optimum/habana/fx/optimization/transformations.py
```diff
@@ -0,0 +1,64 @@
...
+        return graph_module
+
+    def reverse(self, graph_module: "GraphModule") -> "GraphModule":
+        # If self.original_gelu_module is None, that means nothing was changed so the reverse transformation is the identity
+        if self.original_gelu_module is not None:
+            for node in self.get_transformed_nodes(graph_module):
+                parent_name, _, name = node.target.rpartition(".")
+                parent_module = graph_module.get_submodule(parent_name)
+                # Set the GELU module back to what it originally was
+                setattr(parent_module, name, self.original_gelu_module)
```
michaelbenayoun, line 62:
```suggestion
                if hasattr(node, "original_gelu_module"):
                    setattr(parent_module, name, node.original_gelu_module)
```

optimum/habana/fx/tracer.py
```diff
@@ -0,0 +1,72 @@
...
+    Extends the [`HFTracer`](https://github.com/huggingface/transformers/blob/dfeb5aa6a9d0cb95c008854c4e67ceecfeff6ccc/src/transformers/utils/fx.py#L713) to be able to track down GELU modules.
+
+    By default, only modules in `torch.nn` are leaf modules so tracing will go inside other modules to track down elementary operations.
+    By making GELU modules leaf ones, we make sure tracking them down is easy as they won't be decomposed into elementary operations.
+    """
+
+    def is_leaf_module(self, m: torch.nn.Module, module_qualified_name: str) -> bool:
+        if isinstance(m, GELU_CLASSES_TO_TRACK):
+            return True
+        return super().is_leaf_module(m, module_qualified_name)
```
michaelbenayoun, line 41:
```suggestion
        return  isinstance(m, GELU_CLASSES_TO_TRACK) or super().is_leaf_module(m, module_qualified_name)
```

optimum/habana/fx/tracer.py
```diff
@@ -0,0 +1,72 @@
...
+        input_names = model.dummy_inputs.keys()
+
+    input_names = list(input_names)
+    concrete_args = get_concrete_args(model, input_names)
+
+    if not disable_check:
+        check_if_model_is_supported(model)
+
+    # Tracing.
+    tracer = OptimumHabanaTracer()
```
michaelbenayoun, line 62: You ned to create your own `symbolic tracer` to change the Tracer class?

optimum/habana/fx/optimization/transformations.py
regisss: True, great remark! I'm going to change that

optimum/habana/fx/optimization/transformations.py
regisss: The thing is that not all GELU modules have a `act` attribute, so `getattr(module, "act")` will raise an error for those that don't have one.

optimum/habana/fx/tracer.py
regisss: Yep! As discussed, I will submit a PR in Transformers to enable specifying a custom tracer when calling `symbolic_trace`.

optimum/habana/fx/optimization/transformations.py
michaelbenayoun: You can specify a default in this case:

```python
if getattr(module, "act", None) == torch.nn.functional.gelu:
```

regisss: Good call!

optimum/habana/transformers/trainer.py
```diff
@@ -226,6 +226,40 @@ def __init__(
...
+            training_traced = symbolic_trace(
+                model,
+                input_names=input_names_for_tracing,
+            )
+            transformed_training_traced = transformation(training_traced)
+            self.training_fwd = transformed_training_traced.forward
+
+            input_names_for_tracing = []
+            sig = inspect.signature(model.forward)
+            for name in list(eval_dataset.features.keys()):
```
ZhaiFeiyue, line 250: what happens if `eval_dataset` is `None`, which means user not set `--do_eval`

regisss: Yes I will add a `if self.eval_dataset is not None` statement as I did above with the training dataset. I was only testing so I have not done it yet but I will.

ZhaiFeiyue:
@regisss There is no performance drop(1X) when running GPT2 training with branch `fx_gelu`, compared with branch `main`, in my side gaudi2 server, here is the perf data
|branch| gaudi2|gaudi|
|-|-|-|
|fx_gelu| 49| 11|
|main| 48| 11|

my command is: 
```bash
python run_clm.py --model_name_or_path gpt2 --dataset_name wikitext --dataset_config_name wikitext-2-raw-v1 --per_device_train_batch_size 4 --per_device_eval_batch_size 4 --do_train --do_eval --output_dir /tmp/test-clm --gaudi_config_name Habana/gpt2 --use_habana --use_lazy_mode --use_hpu_graphs --use_cache False --throughput_warmup_steps 3 --overwrite_output_dir
```

ZhaiFeiyue:
running model T5, I could reproduce the crash
image
seems that tensor device issue

optimum/habana/transformers/trainer.py
michaelbenayoun: One solution as well is to create one batch of data and use it to infer the input names. But this might be more costly.

michaelbenayoun:
What might be happening here is a "constant" tensor being traced as a model attribute. Hence being on the device used for tracing instead of the HPU. Just a guess though.

regisss:
Yes this implementation is quite experimental as it was mainly for testing a few things. I'll push a cleaner one soon.

PR #226, opened 2023-05-05
Enable usage of PyTorch autocast on Gaudi during training

jwieczorekhabana:
This change introduces usage of PyTorch autocast in a similar way to huggingface transformers trainer. To use autocast context  in GaudiTrainer "--half_precision_backend=hpu_amp" needs to be set.

### DIFF
--- a/examples/audio-classification/README.md
+++ b/examples/audio-classification/README.md
@@ -47,7 +47,8 @@ python run_audio_classification.py \
     --use_lazy_mode \
     --use_hpu_graphs \
     --gaudi_config_name Habana/wav2vec2 \
-    --throughput_warmup_steps 3
+    --throughput_warmup_steps 3 \
+    --half_precision_backend=hpu_amp
 ```
 
 On a single HPU, this script should run in ~13 minutes and yield an accuracy of **97.96%**.
@@ -83,7 +84,8 @@ python ../gaudi_spawn.py \
     --use_lazy_mode \
     --use_hpu_graphs \
     --gaudi_config_name Habana/wav2vec2 \
-    --throughput_warmup_steps 3
+    --throughput_warmup_steps 3 \
+    --half_precision_backend=hpu_amp
 ```
 
 On 8 HPUs, this script should run in ~12 minutes and yield an accuracy of **80.49%**.
--- a/optimum/habana/transformers/gaudi_configuration.py
+++ b/optimum/habana/transformers/gaudi_configuration.py
@@ -57,6 +57,7 @@ class GaudiConfig(BaseConfig):
         self.hmp_bf16_ops = kwargs.pop("hmp_bf16_ops", DEFAULT_BF16_OPS)
         self.hmp_fp32_ops = kwargs.pop("hmp_fp32_ops", DEFAULT_FP32_OPS)
         self.hmp_is_verbose = kwargs.pop("hmp_is_verbose", False)
+        self.disable_autocast = kwargs.pop("disable_autocast", False)
 
         # Use Habana's custom AdamW implementation
         self.use_fused_adam = kwargs.pop("use_fused_adam", False)
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -142,6 +142,13 @@ class GaudiTrainer(Trainer):
             logger.info(f"No `GaudiTrainingArguments` passed, using `output_dir={output_dir}`.")
             args = GaudiTrainingArguments(output_dir=output_dir)
 
+        from dataclasses import asdict
+        if args.half_precision_backend is not None and args.half_precision_backend == 'hpu_amp':
+            delattr(args, 'half_precision_backend')
+            self.use_hpu_amp = True
+        else:
+            self.use_hpu_amp = False
+
         super().__init__(
             model,
             args,
@@ -162,6 +169,9 @@ class GaudiTrainer(Trainer):
             self.gaudi_config = copy.deepcopy(gaudi_config)
 
         if self.args.use_habana:
+            if self.gaudi_config.disable_autocast ==True:
+                self.use_hpu_amp = False
+
             if self.args.use_lazy_mode:
                 try:
                     import habana_frameworks.torch.core as htcore
@@ -181,7 +191,7 @@ class GaudiTrainer(Trainer):
                 # To avoid warnings about parallelism in tokenizers
                 os.environ["TOKENIZERS_PARALLELISM"] = "false"
 
-            if self.gaudi_config.use_habana_mixed_precision:
+            if self.gaudi_config.use_habana_mixed_precision and not self.use_hpu_amp:
                 try:
                     from habana_frameworks.torch.hpex import hmp
                 except ImportError as error:
@@ -864,7 +874,7 @@ class GaudiTrainer(Trainer):
                             self.FusedNorm.clip_norm(model.parameters())
                         else:
                             # Revert to normal clipping otherwise
-                            if args.use_habana and self.gaudi_config.use_habana_mixed_precision:
+                            if args.use_habana and (not self.use_hpu_amp) and self.gaudi_config.use_habana_mixed_precision:
                                 with self.hmp.disable_casts():
                                     torch.nn.utils.clip_grad_norm_(model.parameters(), args.max_grad_norm)
                             else:
@@ -880,7 +890,8 @@ class GaudiTrainer(Trainer):
                     elif (
                         args.use_habana
                         and self.gaudi_config.use_habana_mixed_precision
-                        and not (self.gaudi_config.use_fused_adam)
+                        and (not self.gaudi_config.use_fused_adam)
+                        and (not self.use_hpu_amp)
                     ):
                         with self.hmp.disable_casts():
                             self.optimizer.step()
@@ -1790,3 +1801,24 @@ class GaudiTrainer(Trainer):
         # Moving a model to HPU disconnects the tied weights, so we have to retie them.
         if self.args.use_habana and hasattr(model, "tie_weights"):
             model.tie_weights()
+
+    def autocast_smart_context_manager(self, cache_enabled: Optional[bool] = True):
+        """
+        A helper wrapper that creates an appropriate context manager for `autocast` while feeding it the desired
+        arguments, depending on the situation. Modified by Habana to include using `autocast` on Gaudi devices.
+        """
+        if self.use_cuda_amp or self.use_cpu_amp or self.use_hpu_amp:
+            if version.parse(version.parse(torch.__version__).base_version) >= version.parse("1.10"):
+                if self.use_cpu_amp:
+                    ctx_manager = torch.cpu.amp.autocast(cache_enabled=cache_enabled, dtype=self.amp_dtype)
+                elif self.use_hpu_amp:
+                    ctx_manager = torch.autocast(device_type='hpu', dtype=torch.bfloat16, enabled=True)
+                else:
+                    ctx_manager = torch.cuda.amp.autocast(cache_enabled=cache_enabled, dtype=self.amp_dtype)
+            else:
+                ctx_manager = torch.cuda.amp.autocast()
+        else:
+            import contextlib
+            ctx_manager = contextlib.nullcontext() if sys.version_info >= (3, 7) else contextlib.suppress()
+
+        return ctx_manager
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -57,7 +57,6 @@ UNSUPPORTED_ARGUMENTS = [
     "fp16_full_eval",
     "fp16_opt_level",
     "fsdp",
-    "half_precision_backend",  # not supported, Habana Mixed Precision should be used and specified in Gaudi configuration
     "mp_parameters",
     "sharded_ddp",
     "tf32",
@@ -175,6 +174,14 @@ class GaudiTrainingArguments(TrainingArguments):
                 "When using distributed training, the value of the flag `find_unused_parameters` passed to "
                 "`DistributedDataParallel`."
             )
+        },
+    )
+
+    half_precision_backend: str = field(
+        default="auto",
+        metadata={
+            "help": "The backend to be used for half precision.",
+            "choices": ["auto", "cuda_amp", "apex", "cpu_amp", "hpu_amp"],
         },
     )
 
--- a/examples/contrastive-image-text/README.md
+++ b/examples/contrastive-image-text/README.md
@@ -108,7 +108,8 @@ python run_clip.py \
     --use_lazy_mode \
     --use_hpu_graphs \
     --gaudi_config_name Habana/clip \
-    --throughput_warmup_steps 3
+    --throughput_warmup_steps 3 \
+    --half_precision_backend=hpu_amp
 ```
 
 
@@ -136,7 +137,8 @@ python ../gaudi_spawn.py --world_size 8 --use_mpi run_clip.py \
     --use_lazy_mode \
     --use_hpu_graphs \
     --gaudi_config_name Habana/clip \
-    --throughput_warmup_steps 3
+    --throughput_warmup_steps 3\
+    --half_precision_backend=hpu_amp
 ```
 
 
--- a/examples/image-classification/README.md
+++ b/examples/image-classification/README.md
@@ -47,7 +47,8 @@ python run_image_classification.py \
     --use_hpu_graphs \
     --gaudi_config_name Habana/vit \
     --throughput_warmup_steps 3 \
-    --dataloader_num_workers 1
+    --dataloader_num_workers 1 \
+    --half_precision_backend=hpu_amp
 ```
 
 For Swin, you need to change/add the following arguments:
@@ -95,7 +96,8 @@ python run_image_classification.py \
     --use_hpu_graphs \
     --gaudi_config_name Habana/vit \
     --throughput_warmup_steps 3 \
-    --dataloader_num_workers 1
+    --dataloader_num_workers 1 \
+    --half_precision_backend=hpu_amp
 ```
 
 Internally, the script will use the [`ImageFolder`](https://huggingface.co/docs/datasets/v2.0.0/en/image_process#imagefolder) feature which will automatically turn the folders into 🤗 Dataset objects.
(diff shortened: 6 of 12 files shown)


regisss:
Great addition to the library @jwieczorekhabana! Is it still a WIP or can I already review it?

optimum/habana/transformers/training_args.py
```diff
@@ -178,6 +177,14 @@ class GaudiTrainingArguments(TrainingArguments):
         },
     )
 
+    half_precision_backend: str = field(
+        default="auto",
+        metadata={
+            "help": "The backend to be used for half precision.",
+            "choices": ["auto", "cuda_amp", "apex", "cpu_amp", "hpu_amp"],
+        },
+    )
```
regisss, line 186: We can remove this if everything is managed with `use_torch_autocast`.

optimum/habana/transformers/trainer.py
```diff
@@ -1790,3 +1801,24 @@ def _move_model_to_device(self, model, device):
         # Moving a model to HPU disconnects the tied weights, so we have to retie them.
         if self.args.use_habana and hasattr(model, "tie_weights"):
             model.tie_weights()
+
+    def autocast_smart_context_manager(self, cache_enabled: Optional[bool] = True):
+        """
+        A helper wrapper that creates an appropriate context manager for `autocast` while feeding it the desired
+        arguments, depending on the situation. Modified by Habana to include using `autocast` on Gaudi devices.
+        """
+        if self.use_cuda_amp or self.use_cpu_amp or self.use_hpu_amp:
```
regisss, line 1810: We can remove all the occurrences of `use_cuda_amp` in this method as they will never be used.

regisss (commented):
Cool PR, Torch autocast will be a great addition :fire: :rocket: 

Here is a first review with two main comments:
- I suggest to change `disable_autocast` in the Gaudi config by `use_torch_autocast`
- We should manage everything linked to autocast in the GaudiTrainer with `self.gaudi_config.use_torch_autocast`. Although it differs from Transformers, it will be easier to maintain and offers the exact same way of managing it as HMP.

optimum/habana/transformers/trainer.py
```diff
@@ -142,6 +142,13 @@ def __init__(
             logger.info(f"No `GaudiTrainingArguments` passed, using `output_dir={output_dir}`.")
             args = GaudiTrainingArguments(output_dir=output_dir)
 
+        from dataclasses import asdict
```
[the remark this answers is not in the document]
jwieczorekhabana: No, thanks for noticing :)

optimum/habana/transformers/gaudi_configuration.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, **kwargs):
         self.hmp_bf16_ops = kwargs.pop("hmp_bf16_ops", DEFAULT_BF16_OPS)
         self.hmp_fp32_ops = kwargs.pop("hmp_fp32_ops", DEFAULT_FP32_OPS)
         self.hmp_is_verbose = kwargs.pop("hmp_is_verbose", False)
+        self.disable_autocast = kwargs.pop("disable_autocast", False)
```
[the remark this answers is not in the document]
jwieczorekhabana: You are right it might be better for now to negate this parameter.

optimum/habana/transformers/trainer.py
```diff
@@ -142,6 +142,13 @@ def __init__(
...
+        if args.half_precision_backend is not None and args.half_precision_backend == 'hpu_amp':
+            delattr(args, 'half_precision_backend')
+            self.use_hpu_amp = True
+        else:
+            self.use_hpu_amp = False
```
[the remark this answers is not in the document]
jwieczorekhabana: Autocast enablement is supposed to be an HMP substitution not an alternative. I' would like to decouple those two approaches as it is easier to remove any HMP related code in case of feature deprecation. This way we also stick closer to
 🤗 Transfromers. I believe we should use GaudiConfig as little as possible if there is a way to reflect native Transformers usage. I stand by my change because of those reasons, but of course if you think your approach is more practical/sensible I will change it :)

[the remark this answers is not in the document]
regisss: Interesting, I had the feeling HMP would be there longer :sweat_smile: 
I have a sync with Sree next Tuesday, I'll let you know what we decide to do moving forward :+1: 
In the meantime, I'll try to run some regression tests on this PR.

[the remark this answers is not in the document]
regisss: So we decided to keep the Transformers way that you proposed :slightly_smiling_face: 
I'm going to resolve comments that are linked to the Gaudi config and review this PR again.

optimum/habana/transformers/training_args.py
```diff
@@ -178,6 +177,14 @@ class GaudiTrainingArguments(TrainingArguments):
         },
     )
 
+    half_precision_backend: str = field(
+        default="auto",
```
regisss, line 181:
```suggestion
        default="hpu_amp",
```
I would just set the default to `hpu_amp`. `auto` enables to automatically switch to `cuda_amp` or `cpu_amp` depending on the device that is being used. But in our case, `cpu_map` will be barely used.

optimum/habana/transformers/training_args.py
```diff
@@ -57,7 +57,6 @@
     "fp16_full_eval",
     "fp16_opt_level",
     "fsdp",
-    "half_precision_backend",  # not supported, Habana Mixed Precision should be used and specified in Gaudi configuration
```
regisss, line 60: I think we should also remove `bf16` (cf. line 53) from this list and use it as the argument for enabling bf16 autocast.

optimum/habana/transformers/trainer.py
```diff
@@ -142,6 +142,12 @@ def __init__(
             logger.info(f"No `GaudiTrainingArguments` passed, using `output_dir={output_dir}`.")
             args = GaudiTrainingArguments(output_dir=output_dir)
 
+        if args.half_precision_backend is not None and args.half_precision_backend == 'hpu_amp':
+            delattr(args, 'half_precision_backend')
+            self.use_hpu_amp = True
+        else:
+            self.use_hpu_amp = False
```
regisss, line 149:
```suggestion
        self.use_hpu_amp = False
        self.use_cpu_amp = False
        if args.bf16:
            if args.half_precision_backend == 'hpu_amp':
                self.use_hpu_amp = True
                delattr(args, 'half_precision_backend')
            else:
                self.use_cpu_amp = True
```
if we use `args.bf16`

optimum/habana/transformers/trainer.py
```diff
@@ -162,6 +168,9 @@ def __init__(
             self.gaudi_config = copy.deepcopy(gaudi_config)
 
         if self.args.use_habana:
+            if self.gaudi_config.use_torch_autocast == False:
+                self.use_hpu_amp = False
```
regisss, line 172: To remove

optimum/habana/transformers/trainer.py
```diff
@@ -1790,3 +1800,19 @@ def _move_model_to_device(self, model, device):
...
+        arguments, depending on the situation. Modified by Habana to include using `autocast` on Gaudi devices.
+        """
+        if self.use_cpu_amp or self.use_hpu_amp:
+            if self.use_cpu_amp:
+                ctx_manager = torch.cpu.amp.autocast(cache_enabled=cache_enabled, dtype=self.amp_dtype)
+            else:
+                ctx_manager = torch.autocast(device_type='hpu', dtype=torch.bfloat16, enabled=True)
+        else:
+            import contextlib
+            ctx_manager = contextlib.nullcontext() if sys.version_info >= (3, 7) else contextlib.suppress()
```
regisss, line 1816:
```suggestion
        if self.use_cpu_amp:
            ctx_manager = torch.cpu.amp.autocast(cache_enabled=cache_enabled, dtype=self.amp_dtype)
        elif self.use_hpu_amp:
            ctx_manager = torch.autocast(device_type='hpu', dtype=torch.bfloat16, enabled=True)
        else:
            import contextlib
            ctx_manager = contextlib.nullcontext() if sys.version_info >= (3, 7) else contextlib.suppress()
```

optimum/habana/transformers/trainer.py
```diff
@@ -181,7 +190,7 @@ def __init__(
                 # To avoid warnings about parallelism in tokenizers
                 os.environ["TOKENIZERS_PARALLELISM"] = "false"
 
-            if self.gaudi_config.use_habana_mixed_precision:
+            if self.gaudi_config.use_habana_mixed_precision and not self.use_hpu_amp:
```
regisss, line 193: We can remove `and not self.use_hpu_amp` here and raise an error just before if both `self.gaudi_config.use_habana_mixed_precision` and `self.use_hpu_amp` are True (the same check as in gaudi_configuration.py).

examples/audio-classification/README.md
```diff
@@ -47,7 +47,8 @@ python run_audio_classification.py \
     --use_lazy_mode \
     --use_hpu_graphs \
     --gaudi_config_name Habana/wav2vec2 \
-    --throughput_warmup_steps 3
+    --throughput_warmup_steps 3 \
+    --half_precision_backend=hpu_amp
```
regisss, line 51:
```suggestion
    --bf16
```

regisss (commented):
@jwieczorekhabana There are a few changes that I suggested last time that we need to remove, sorry for this.

My main comment is about using the training argument `bf16` to trigger the autocast as it is done in Transformers.

Let's also modify the integration test here: https://github.com/huggingface/optimum-habana/blob/063f4bc259e6a595955508837852244e0752abbf/tests/test_trainer.py#L568
We can just uncomment the commented snippet and remove both the TODO comment and the `self.assertRaises(ValueError)` block.

optimum/habana/transformers/gaudi_configuration.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, **kwargs):
         self.hmp_bf16_ops = kwargs.pop("hmp_bf16_ops", DEFAULT_BF16_OPS)
         self.hmp_fp32_ops = kwargs.pop("hmp_fp32_ops", DEFAULT_FP32_OPS)
         self.hmp_is_verbose = kwargs.pop("hmp_is_verbose", False)
+        self.use_torch_autocast = kwargs.pop("use_torch_autocast", False)
+
+        if self.use_habana_mixed_precision and self.use_torch_autocast:
+            raise ValueError("`use_habana_mixed_precision` and `use_torch_autocast` cannot be both `True` in your Gaudi configuration, you must choose one or the other to perform mixed-precision training.")
```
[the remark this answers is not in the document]
jwieczorekhabana: Do I understand correctly that anything related to autocast should be removed from GaudiConfig? For now not every model is fit to run under autocast context. I think the same stands for hmp. Not every model uses it. Should we somehow note that specific model shouldn't be used with autocast or maybe it would be good to leave it up to GaudiConfig?

optimum/habana/transformers/trainer.py
jwieczorekhabana: We can't raise any exceptions as it would be impossible to use autocast. This if statement is responsible for converting ops using hmp. Given that user hasn't specified to use autocast, decision to cast using hmp is left to HabanaConfig.

optimum/habana/transformers/gaudi_configuration.py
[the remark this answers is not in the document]
regisss: Yes you're right, I went a bit fast. Leaving it up to the Gaudi config sounds good to me, and that way users shouldn't have to care about this unless they want to go at a lower level.
So let's keep `use_torch_autocast` in the Gaudi configuration. And then, at trainer initialization, if `use_torch_autocast` is True we can set `args.bf16 = True` and also raise the error there if `use_habana_mixed_precision` is True too. WDYT?

optimum/habana/transformers/trainer.py
regisss: Not sure I understand. Basically what I'm proposing here is the following:
```py
if self.gaudi_config.use_habana_mixed_precision and (args.bf16 or self.gaudi_config.use_torch_autocast):
    raise ValueError(...)

[...]

if self.gaudi_config.use_habana_mixed_precision:
    hmp.convert(...)
``` 
The user can enable autocast with `args.bf16` or `gaudi_config.use_torch_autocast` (the first one being easier to use), but these are not compatible with `gaudi_config.use_habana_mixed_precision`.

optimum/habana/transformers/gaudi_configuration.py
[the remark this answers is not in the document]
jwieczorekhabana: I don't think we can enable bf16 argument without deleting it like [here](https://github.dev/jwieczorekhabana/optimum-habana/blob/9fdd92ec6f4eba0b4d98fd752ed207a067426340/optimum/habana/transformers/trainer.py#L149) , because it would be propagated to parent HF Trainer. Correct me if I'm wrong but I think that it would cause error [here](https://github.com/huggingface/transformers/blob/a7920065f2cfd2549b838f9a30afd7c265fcdd88/src/transformers/trainer.py#LL611C16-L611C16). And were we to not delete args.half_precision_backend I think it would throw an error [here](https://github.com/huggingface/transformers/blob/a7920065f2cfd2549b838f9a30afd7c265fcdd88/src/transformers/trainer.py#L652).

I think it's better to familiarize users with autocast usage on Gaudi by setting half_precision_backend=hpu_amp (as we can't utilize auto option) rather than adding another argument that needs to be used and works only as a dummy.

optimum/habana/transformers/trainer.py
jwieczorekhabana: Yes but in this case user first needs to download and alter/create their own gaudi_config.json instead of using preexisting ones. In  what I propose hpu_precision_backend argument takes precedence over gaudi_config.use_habana_mixed_precision. Args.bf16 is not easier to use as it also requires the change of gaudi_config.json for now. Of course it is not the case if you will change all configs use_habana_mixed_precision to false but I have no information of how are those configuration files going to be altered.

jwieczorekhabana: I admit that usage of gaudi_config.use_torch_autocast seems kind of strange in this scenario. I thought of it rather as a temporary workaround to control whether autocast is enabled for specific models as changes in GaudiTrainer are model agnostic.

optimum/habana/transformers/gaudi_configuration.py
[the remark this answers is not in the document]
regisss: Well the `--bf16` argument is the one that actually enables mixed-precision training in Transformers. `half_precision_backend` is just there to specify the backend and doesn't enable anything.

I'm okay with not using it. But in that case, users are not really going to use `half_precision_backend` either since it will default to `hpu_amp` (unless they want to use the CPU, which should almost never happen). They will just control it with the `use_torch_autocast` field of the Gaudi config.

optimum/habana/transformers/trainer.py
jwieczorekhabana: I agree that bf16 would resemble Transformers more but then we would use auto backend which looks at the device used. I don't believe any of those two approaches is perfect (either bf16 or half_precision_backend). My line of reasoning looks like this. Imagine there is no GaudiTrainer and you can use models on Gaudi with plain Transformers library. What would you need to do to run mixed precision training? Set bf16 which is the same as it was. You also need to specify half_precision_backend and yes it is inferred from device in case of auto but specifying it seems more verbose. Idk it was just the gut feeling or sth like it. I'm not really attached to my solution, only thing is that I think we should either use bf16 or half_precision_backend not both, your call :)

optimum/habana/transformers/gaudi_configuration.py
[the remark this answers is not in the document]
regisss: Regarding the errors that would be triggered with `bf16` in the parent init, I think we should just set `args.bf16 = False` right before calling `super().__init__` and set it back to its value right after. I know it's very dirty, but 1) we don't need the code in the `if args.bf16` blocks to be run, 2) since we constantly adapt to Transformers & co, things are not always pretty and I prefer having something easier to maintain even though it's dirty.

regisss:
@jwieczorekhabana Not sure if everything I wrote about using `--bf16` for enabling autocast is clear, let me know if it's not :slightly_smiling_face:

jwieczorekhabana:
@regisss I've changed the script so that bf16 argument is now responsible for enabling autocast.

regisss (approved):
LGTM @jwieczorekhabana !

Have you tried it with SynapseAI 1.10 to see if it works with more examples/models?

optimum/habana/transformers/trainer.py
```diff
@@ -864,7 +878,7 @@ def hpu_deepspeed_checkpointing(function, *checkpoint_args):
                             self.FusedNorm.clip_norm(model.parameters())
                         else:
                             # Revert to normal clipping otherwise
-                            if args.use_habana and self.gaudi_config.use_habana_mixed_precision:
+                            if args.use_habana and (not (self.use_hpu_amp self.use_cpu_amp)) and self.gaudi_config.use_habana_mixed_precision:
```
regisss, line 881:
```suggestion
                            if args.use_habana and (not (self.use_hpu_amp or self.use_cpu_amp)) and self.gaudi_config.use_habana_mixed_precision:
```

optimum/habana/transformers/trainer.py
```diff
@@ -880,7 +894,8 @@ def hpu_deepspeed_checkpointing(function, *checkpoint_args):
                     elif (
                         args.use_habana
                         and self.gaudi_config.use_habana_mixed_precision
-                        and not (self.gaudi_config.use_fused_adam)
+                        and (not self.gaudi_config.use_fused_adam)
+                        and (not (self.use_hpu_amp of self.use_cpu_amp))
```
regisss, line 898:
```suggestion
                        and (not (self.use_hpu_amp or self.use_cpu_amp))
```

regisss (commented):
I spotted a few typos in the code that make the tests fail.

Also, could you run the following to make sure the style check pass please?
``` bash
pip install black ruff
make style
```

jwieczorekhabana:
@regisss thanks for finding typos. I've fixed styling and verified whether this patch works by running example summarization script

PR #229, opened 2023-05-10
text-generation support sampling algo

ZhaiFeiyue:
# What does this PR do?

As title, user could use `sampling` algo for text-generation with args `--do_sample True`

## Before submitting

### DIFF
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -74,7 +74,7 @@ with
 - `number_of_iterations` the number of iterations to perform in the benchmark
 - `use_hpu_graphs` enables HPU graphs which are recommended for faster latencies
 - `use_kv_cache` enables a key-value cache to speed up the generation process.
-- `enable_sampling` enables sampling algorithm for the generation process.
+- `do_sample` enables sampling algorithm for the generation process.
 
 For example, you can reproduce the results presented in [this blog post](https://huggingface.co/blog/habana-gaudi-2-bloom) with the following command:
 ```bash
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -63,7 +63,7 @@ python (../gaudi_spawn.py --use_deepspeed --world_size number_of_devices) run_ge
 --n_iterations number_of_iterations \
 --use_hpu_graphs \
 --use_kv_cache \
---do_sample enable_sampling
+--do_sample
 ```
 with
 - `number_of_devices` the number of HPUs you want to use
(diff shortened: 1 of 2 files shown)


examples/text-generation/README.md
```diff
@@ -62,7 +62,8 @@ python (../gaudi_spawn.py --use_deepspeed --world_size number_of_devices) run_ge
 --batch_size batch_size \
 --n_iterations number_of_iterations \
 --use_hpu_graphs \
---use_kv_cache
+--use_kv_cache \
+--do_sample enable_sampling
```
regisss, line 66:
```suggestion
--do_sample
```
This should be enough.

examples/text-generation/README.md
```diff
@@ -73,6 +74,7 @@ with
 - `number_of_iterations` the number of iterations to perform in the benchmark
 - `use_hpu_graphs` enables HPU graphs which are recommended for faster latencies
 - `use_kv_cache` enables a key-value cache to speed up the generation process.
+- `enable_sampling` enables sampling algorithm for the generation process.
```
regisss, line 77:
```suggestion
- `do_sample` enables sampling algorithm for the generation process.
```
The name of the argument is `do_sample`.

regisss (commented):
Thanks for adding this @ZhaiFeiyue! I left a few comments regarding the naming and the  management of this new argument.

Quick question, have you tried generating with sampling? When I tried it a few months ago, it was extremely slow because of `torch.multinomial` :slightly_frowning_face:

examples/text-generation/run_generation.py
```diff
@@ -78,6 +78,13 @@ def main():
         help="Optional argument if you want to assess your model on a given dataset of the HF Hub, this will be the name of the column to use as prompts for generation.",
     )
 
+    parser.add_argument(
+        "--do_sample",
+        default=False,
+        type=bool,
+        help="Whether to use sampling for generation.",
+    )
```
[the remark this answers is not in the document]
ZhaiFeiyue: done

examples/text-generation/README.md
ZhaiFeiyue: done

ZhaiFeiyue:
@regisss  we have enabled GPT2 with sampling and in our side the vocab size shi 250K, we have done some optimizations special for the `multinomial`, I will submit a PR later if the optimization is acceptable, and seems not too bad with BF16. Since we are running GPT2 under BF16 , but the GPT2 `gaudi_config` is disable the HMP for training, so that's why I open an issue #230 to ask you if we can enable the BF16 for GPT2 training(perplexity drop a little when enable HMP)

we also have some optimizations on HPUGraph, we do capture , replay and caching recipe by ourselves, but it is a little complex , we could have a discussion it with later.

regisss (approved):
LGTM!

PR #231, opened 2023-05-10
optimize conv1d forward

ZhaiFeiyue:
# What does this PR do?

current the implementation of Conv1D is using `torch.addmm`, with which there is a `view` op after bias add, the `view` op is mapped to `reshape` op on Gaudi, a `reshape` will break the tpc auto fusion, so this PR is try to optimize the Conv1D by put the bias add after the `view`.

here is the perf of text-generation with GPT2 on Gaudi2

branch optimize_conv1d: 509
branch main: 422

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/modeling_all_models.py
+++ b/optimum/habana/transformers/models/modeling_all_models.py
@@ -93,3 +93,15 @@ def gaudi_get_extended_attention_mask(
     extended_attention_mask = extended_attention_mask * torch.finfo(extended_attention_mask.dtype).min
 
     return extended_attention_mask
+
+
+def gaudi_conv1d_forward(self, x):
+    # move reshape before view for tpc auto fusion
+    size_out = x.size()[:-1] + (self.nf,)
+    x = torch.mm(x.view(-1, x.size(-1)), self.weight)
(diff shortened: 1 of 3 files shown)


optimum/habana/transformers/models/modeling_all_models.py
```diff
@@ -93,3 +93,15 @@ def gaudi_get_extended_attention_mask(
     extended_attention_mask = extended_attention_mask * torch.finfo(extended_attention_mask.dtype).min
 
     return extended_attention_mask
+
+
+def gaudi_conv1d_forward(self, x):
+    # move reshape before view for tpc auto fusion
```
regisss, line 99:
```suggestion
    """
    Same as https://github.com/huggingface/transformers/blob/3335724376319a0c453049d0cd883504f530ff52/src/transformers/pytorch_utils.py#L100
    but moves reshape before view for tpc auto fusion.
    """
```

regisss (commented):
Nice :rocket: 

I just left one comment to document the new forward slightly better (basically adding a pointer to the Transformers Conv1D forward).

regisss (approved):
LGTM!

PR #237, opened 2023-05-17
Added an option to remove save checkpoint time from throughput calculation

libinta:
In the case that train time is very short, the distributed training throughput is greatly impacted by saving checkpoint time, and it gives misleading scaling effciency. This patch provides an option to remove the save checkpoint time from throughput calculation.

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -77,7 +77,6 @@ from transformers.utils import (
     WEIGHTS_NAME,
     is_datasets_available,
     is_safetensors_available,
-    is_torch_tpu_available,
 )
 
 from optimum.utils import logging
@@ -417,9 +416,6 @@ class GaudiTrainer(Trainer):
 
     def _maybe_log_save_evaluate(self, tr_loss, model, trial, epoch, ignore_keys_for_eval):
         if self.control.should_log:
-            if is_torch_tpu_available():
-                xm.mark_step()
-
             logs: Dict[str, float] = {}
 
             # all_gather + mean() to get average loss over all processes
@@ -456,7 +452,7 @@ class GaudiTrainer(Trainer):
                 save_start = time.time()
             self._save_checkpoint(model, trial, metrics=metrics)
             if self.args.throughput_rm_save_ckpt_time:
-                self.save_ckpt_time += (time.time() - save_start)
+                self.save_ckpt_time += time.time() - save_start
             self.control = self.callback_handler.on_save(self.args, self.state, self.control)
 
     def train(
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -417,6 +417,8 @@ class GaudiTrainer(Trainer):
     def _maybe_log_save_evaluate(self, tr_loss, model, trial, epoch, ignore_keys_for_eval):
         if self.control.should_log:
             logs: Dict[str, float] = {}
+            if self.args.adjust_throughput:
+                save_start = time.time()
 
             # all_gather + mean() to get average loss over all processes
             tr_loss_scalar = self._nested_gather(tr_loss).mean().item()
@@ -448,11 +450,9 @@ class GaudiTrainer(Trainer):
             self._report_to_hp_search(trial, self.state.global_step, metrics)
 
         if self.control.should_save:
-            if self.args.throughput_rm_save_ckpt_time:
-                save_start = time.time()
             self._save_checkpoint(model, trial, metrics=metrics)
-            if self.args.throughput_rm_save_ckpt_time:
-                self.save_ckpt_time += time.time() - save_start
+            if self.args.adjust_throughput:
+                self.log_evaluate_save_time += time.time() - save_start
             self.control = self.callback_handler.on_save(self.args, self.state, self.control)
 
     def train(
@@ -814,10 +814,10 @@ class GaudiTrainer(Trainer):
                     # Otherwise we need to call the whooooole sampler cause there is some random operation added
                     # AT THE VERY END!
                     _ = list(train_dataloader.sampler)
-        if self.args.throughput_rm_save_ckpt_time:
-            self.save_ckpt_time = 0
+        if self.args.adjust_throughput:
+            self.log_evaluate_save_time = 0
         else:
-            self.save_ckpt_time = None
+            self.log_evaluate_save_time = None
         for epoch in range(epochs_trained, num_train_epochs):
             if isinstance(train_dataloader, DataLoader) and isinstance(train_dataloader.sampler, DistributedSampler):
                 train_dataloader.sampler.set_epoch(epoch)
@@ -991,7 +991,7 @@ class GaudiTrainer(Trainer):
             num_samples=num_samples_for_speed_metrics,
             num_steps=num_steps_for_speed_metrics,
             start_time_after_warmup=start_time_after_warmup,
-            save_ckpt_time=self.save_ckpt_time,
+            log_evaluate_save_time=self.log_evaluate_save_time,
         )
         self.store_flos()
         metrics["total_flos"] = self.state.total_flos
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -416,10 +416,11 @@ class GaudiTrainer(Trainer):
         return model
 
     def _maybe_log_save_evaluate(self, tr_loss, model, trial, epoch, ignore_keys_for_eval):
+        if self.args.adjust_throughput:
+            save_start = time.perf_counter()
+
         if self.control.should_log:
             logs: Dict[str, float] = {}
-            if self.args.adjust_throughput:
-                save_start = time.time()
 
             # all_gather + mean() to get average loss over all processes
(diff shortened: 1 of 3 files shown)


optimum/habana/transformers/trainer.py
```diff
@@ -70,13 +70,14 @@
     get_last_checkpoint,
     has_length,
 )
-from transformers.training_args import TrainingArguments
+from transformers.training_args import TrainingArguments, ParallelMode
```
regisss, line 73:
```suggestion
from transformers.training_args import TrainingArguments
```
Not needed

optimum/habana/transformers/trainer.py
```diff
@@ -414,6 +415,50 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
+                        metric_key_prefix=f"eval_{eval_dataset_name}",
+                    )
+                    metrics.update(dataset_metrics)
+            else:
+                metrics = self.evaluate(ignore_keys=ignore_keys_for_eval)
+            self._report_to_hp_search(trial, self.state.global_step, metrics)
+
+        if self.control.should_save:
+            if self.args.throughput_rm_save_ckpt_time:
+                save_start = time.time()
```
regisss, line 456: Actually I think we should also measure the time taken for logging and evaluating right above in this method.
Maybe we can just measure the time at the begining and at the end of this method, WDYT?

optimum/habana/transformers/training_args.py
```diff
@@ -123,6 +125,11 @@ class GaudiTrainingArguments(TrainingArguments):
         },
     )
 
+    throughput_rm_save_ckpt_time:  bool = field(
+        default=False,
+        metadata={"help": "Whether to remove save checkpoint time from throughput calculation."},
+    )
```
regisss, line 131:
```suggestion
    adjust_throughput:  bool = field(
        default=False,
        metadata={"help": "Whether to remove the time taken for logging, evaluating and saving from throughput calculation."},
    )
```
A proposal to take into account everything.

optimum/habana/utils.py
```diff
@@ -70,6 +70,7 @@ def speed_metrics(
     num_samples: int = None,
     num_steps: int = None,
     start_time_after_warmup: float = None,
+    save_ckpt_time: float = None,
```
regisss, line 73:
```suggestion
    log_evaluate_save_time: float = None,
```

optimum/habana/utils.py
```diff
@@ -82,6 +83,7 @@ def speed_metrics(
         num_samples (int, optional): number of samples processed. Defaults to None.
         num_steps (int, optional): number of steps performed. Defaults to None.
         start_time_after_warmup (float, optional): time after warmup steps have been performed. Defaults to None.
+        save_ckpt_time (float, optional): time save checkpoint have been performed. Defaults to None.
```
regisss, line 86:
```suggestion
        log_evaluate_save_time (float, optional): time spent to log, evaluate and save. Defaults to None.
```

optimum/habana/utils.py
```diff
@@ -90,6 +92,10 @@ def speed_metrics(
     runtime = time.time() - start_time
     result = {f"{split}_runtime": round(runtime, 4)}
 
+    # Adjust runtime if save ckpt time is not included
+    if save_ckpt_time is not None:
+        runtime = runtime - save_ckpt_time
```
regisss, line 97:
```suggestion
    # Adjust runtime if log_evaluate_save_time should not be included
    if log_evaluate_save_time is not None:
        runtime = runtime - log_evaluate_save_time
```

regisss (commented):
Thanks for this PR @libinta!!

My main comment is about counting also the time spent to log and evaluate in the `_maybe_log_save_evaluate` method. It doesn't really change the PR, just the places where timers are set and a few variable names.

Also, could you run `make style` from the root of the repo please? I see the style check failed.

optimum/habana/transformers/trainer.py
libinta: will that be too much difference with transformers? also there has loss over all ranks for 8x.

regisss: I think it's fine since we are not modifying the base behavior but only adding a new argument to enable another behavior.

regisss: You mean that the loss is logged by all processes in distributed training?

libinta: right, loss is averaged across all ranks/processes in distributed training.

regisss: Is this an issue?

libinta: I think maybe we should be count that portion as it's part of 8x training.

optimum/habana/transformers/training_args.py
```diff
@@ -85,6 +85,8 @@ class GaudiTrainingArguments(TrainingArguments):
             Number of steps to ignore for throughput calculation. For example, with `throughput_warmup_steps=N`,
             the first N steps will not be considered in the calculation of the throughput. This is especially
             useful in lazy mode where the first two or three iterations typically take longer.
+        throughput_rm_save_ckpt_time ('bool', *optional*, defaults to `False`):
+            Whether to remove save checkpoint time from throughput calculation.
```
regisss, line 89:
```suggestion
        adjust_throughput ('bool', *optional*, defaults to `False`):
            Whether to remove the time taken for logging, evaluating and saving from throughput calculation.
```

optimum/habana/transformers/trainer.py
```diff
@@ -414,6 +414,47 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
```
regisss, line 452:
```suggestion
            if self.args.adjust_throughput:
                save_start = time.time()
```
To move at the beginning of this method.

optimum/habana/transformers/trainer.py
```diff
@@ -414,6 +414,47 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
+            self._save_checkpoint(model, trial, metrics=metrics)
+            if self.args.throughput_rm_save_ckpt_time:
```
regisss, line 454:
```suggestion
            if self.args.adjust_throughput:
```

optimum/habana/transformers/trainer.py
```diff
@@ -947,6 +991,7 @@ def hpu_deepspeed_checkpointing(function, *checkpoint_args):
             num_samples=num_samples_for_speed_metrics,
             num_steps=num_steps_for_speed_metrics,
             start_time_after_warmup=start_time_after_warmup,
+            save_ckpt_time=self.save_ckpt_time,
```
regisss, line 994:
```suggestion
            log_evaluate_save_time=self.log_evaluate_save_time,
```

optimum/habana/transformers/trainer.py
```diff
@@ -773,7 +814,10 @@ def hpu_deepspeed_checkpointing(function, *checkpoint_args):
                     # Otherwise we need to call the whooooole sampler cause there is some random operation added
                     # AT THE VERY END!
                     _ = list(train_dataloader.sampler)
-
+        if self.args.throughput_rm_save_ckpt_time:
+            self.save_ckpt_time = 0
+        else:
+            self.save_ckpt_time = None
```
regisss, line 820:
```suggestion
        if self.args.adjust_throughput:
            self.log_evaluate_save_time = 0
        else:
            self.log_evaluate_save_time = None
```

optimum/habana/transformers/trainer.py
regisss: Anyway even in 8x training, the throughput that is logged is the throughput of the main process. So not sure that changes anything.

optimum/habana/transformers/trainer.py
```diff
@@ -414,6 +414,47 @@ def _wrap_model(self, model, training=True, dataloader=None):
 
         return model
 
+    def _maybe_log_save_evaluate(self, tr_loss, model, trial, epoch, ignore_keys_for_eval):
+        if self.control.should_log:
+            logs: Dict[str, float] = {}
+            if self.args.adjust_throughput:
+                save_start = time.time()
```
regisss, line 421: This should move up before line 418 because we may not enter the log block.

optimum/habana/transformers/trainer.py
```diff
@@ -414,6 +414,47 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
+                    )
+                    metrics.update(dataset_metrics)
+            else:
+                metrics = self.evaluate(ignore_keys=ignore_keys_for_eval)
+            self._report_to_hp_search(trial, self.state.global_step, metrics)
+
+        if self.control.should_save:
+            self._save_checkpoint(model, trial, metrics=metrics)
+            if self.args.adjust_throughput:
+                self.log_evaluate_save_time += time.time() - save_start
```
regisss, line 455: This should move down at the end of the method because we not enter the save block.

regisss:
@libinta I pushed the changes I suggested because I will use this new argument for the blog post I'm working on. Let me know if it's okay to merge.

PR #242, opened 2023-05-23
Refine doc and README

regisss:
# What does this PR do?

As per title.

## Before submitting


regisss:
Unit and integration tests failed because we need to update a token to get access to a AWS instance. This is not related to this PR so merging.

PR #249, opened 2023-05-26
Enable GPTNeoX

ZhaiFeiyue:
# What does this PR do?
as title

## Before submitting


ZhaiFeiyue:
@regisss  this is the last model 😄, please have a look if free.
I'm using a customer's 2.7B model for testing since 20B is too large to downloading in my side.

regisss (approved):
@ZhaiFeiyue You're on fire :fire: 

I just tried the 20B-parameter checkpoint on Gaudi2 and it seems to work well :slightly_smiling_face: 
bf16 also works and same issue with DeepSpeed-inference as with other models.

Issue #245, opened 2023-05-25
Adding profiling


ZhaiFeiyue:
### Feature request

HabanaAI support PyTorch profiler, and currently if user want to figure out the bottleneck of training or inference, they should modify the source code(`adding torch.profiler` into `GaudiTrainer `or `generate`), with which  the user experience is not good.
If we support `torch.profiler` in optimum-habana, and user could generate the profiling data just with `--do_profiling 10`
- default value of do_profiling  is 0, means not generate the profiling data
- 10 means how many steps or iterations will be captured

refer [profiling-with-pytorch](https://docs.habana.ai/en/latest/Profiling/Profiling_with_PyTorch.html?highlight=profiling)

### Motivation

- easy to address the bottleneck of model both for training and inference

### Your contribution

submit a PR

ZhaiFeiyue:
@regisss is there any possible that we support such kind of feature?

regisss:
@ZhaiFeiyue Sounds good to me! Do you want to open a PR to add this?

ZhaiFeiyue:
@regisss I will do that

ZhaiFeiyue:
fixed by PR #250, and close

PR #250, opened 2023-05-29
Enable profiling

ZhaiFeiyue:
# What does this PR do?

Adding profiling into `GaudiTrainer` and `text-generation`

Fixes # (issue)
#245 

## Before submitting

### DIFF
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -174,7 +174,7 @@ class GaudiTrainingArguments(TrainingArguments):
 
     profiling_steps: Optional[int] = field(
         default=0,
-        metadata={"help": ("Number of steps to be captured when enable profiling.")},
+        metadata={"help": ("Number of steps to be captured when enabling profiling.")},
     )
 
     # Overriding ddp_find_unused_parameters to make False the default value
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -167,6 +167,16 @@ class GaudiTrainingArguments(TrainingArguments):
         },
     )
 
+    profiling_warmup_steps: Optional[int] = field(
+        default=0,
+        metadata={"help": ("Number of steps to ignore for profling.")},
+    )
+
+    profiling_steps: Optional[int] = field(
+        default=0,
+        metadata={"help": ("Number of steps to be captured when enable profiling.")},
+    )
+
     # Overriding ddp_find_unused_parameters to make False the default value
     ddp_find_unused_parameters: Optional[bool] = field(
         default=False,
--- a/optimum/habana/utils.py
+++ b/optimum/habana/utils.py
@@ -264,7 +264,7 @@ class HabanaProfile(object):
     @staticmethod
     def disable():
         """
-        run only once and must happened before doing profiling
+        Runs only once and must happen before doing profiling.
         """
         if hasattr(HabanaProfile.disable, "invalid"):
             if not HabanaProfile.disable.invalid:
--- a/optimum/habana/utils.py
+++ b/optimum/habana/utils.py
@@ -216,3 +216,69 @@ def get_driver_version():
     if output.returncode == 0:
         return version.parse(output.stdout.split("\n")[2].replace(" ", "").split(":")[1][:-1].split("-")[0])
     return None
+
+
+class HabanaProfile(object):
+    """
+    HPU profiler only could be run once, so HABANA_PROFILE_ENABLED, a class static variable shared by all the instances of HabanaProfile, is used to control which part will be captured
+    """
+
+    HABANA_PROFILE_ENABLED = False
+
+    def __init__(self, warmup: int = 0, active: int = 0, output_dir: str = "./hpu_profile", wait: int = 0):
+        if active <= 0 or warmup <= 0 or HabanaProfile.HABANA_PROFILE_ENABLED:
+
+            def noop():
+                pass
+
+            self.start = noop
+            self.stop = noop
+            self.step = noop
+        else:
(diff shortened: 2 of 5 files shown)


ZhaiFeiyue:
@regisss have a look if free

optimum/habana/utils.py
```diff
@@ -216,3 +216,69 @@ def get_driver_version():
     if output.returncode == 0:
         return version.parse(output.stdout.split("\n")[2].replace(" ", "").split(":")[1][:-1].split("-")[0])
     return None
+
+
+class HabanaProfile(object):
+    """
+    HPU profiler only could be run once, so HABANA_PROFILE_ENABLED, a class static variable shared by all the instances of HabanaProfile, is used to control which part will be captured
```
regisss, line 223:
```suggestion
    HPU profiler only could be run once, so HABANA_PROFILE_ENABLED, a class static variable shared by all the instances of HabanaProfile, is used to control which part will be captured.
```

optimum/habana/utils.py
```diff
@@ -216,3 +216,69 @@ def get_driver_version():
...
+    def start(self):
+        self.start()
+
+    def step(self):
+        self.step()
+
+    @staticmethod
+    def disable():
+        """
+        run only once and must happened before doing profiling
```
regisss, line 267:
```suggestion
        Runs only once and must happen before doing profiling.
```

optimum/habana/utils.py
```diff
@@ -216,3 +216,69 @@ def get_driver_version():
...
+        if hasattr(HabanaProfile.disable, "invalid"):
+            if not HabanaProfile.disable.invalid:
+                HabanaProfile.HABANA_PROFILE_ENABLED = True
+        else:
+            HabanaProfile.HABANA_PROFILE_ENABLED = True
+
+    @staticmethod
+    def enable():
+        """
+        run only once and must happened before doing profiling
```
regisss, line 278:
```suggestion
        Runs only once and must happen before doing profiling.
```

optimum/habana/utils.py
```diff
@@ -216,3 +216,69 @@ def get_driver_version():
...
+                activities=activities,
+                on_trace_ready=torch.profiler.tensorboard_trace_handler(output_dir),
+                record_shapes=True,
+                with_stack=True,
+            )
+            self.start = profiler.start
+            self.stop = profiler.stop
+            self.step = profiler.step
+            HabanaProfile.enable.invalid = True
+            HabanaProfile.disable.invalid = True
```
regisss, line 253: What do these attributes do?

optimum/habana/transformers/training_args.py
```diff
@@ -167,6 +167,16 @@ class GaudiTrainingArguments(TrainingArguments):
...
     )
 
+    profiling_warmup_steps: Optional[int] = field(
+        default=0,
+        metadata={"help": ("Number of steps to ignore for profling.")},
+    )
+
+    profiling_steps: Optional[int] = field(
+        default=0,
+        metadata={"help": ("Number of steps to be captured when enable profiling.")},
```
regisss, line 177:
```suggestion
        metadata={"help": ("Number of steps to be captured when enabling profiling.")},
```

regisss (commented):
That's awesome to be able to easily use the profiler @ZhaiFeiyue :smiley: :rocket: 
I just left a few very minor comments.

optimum/habana/utils.py
ZhaiFeiyue: done

optimum/habana/utils.py
ZhaiFeiyue: done

optimum/habana/utils.py
ZhaiFeiyue: done

optimum/habana/transformers/training_args.py
ZhaiFeiyue: done

optimum/habana/utils.py
ZhaiFeiyue: what I want is:
- `enable` and `disable` could run many times before `return a profiler object to user`
- `enable` and `disable` should not work after   `return a profiler object to user`

that's because in one process we only can do profiling once, after we generate the profiling, we can not enable it again.

optimum/habana/utils.py
```diff
@@ -216,3 +216,69 @@ def get_driver_version():
...
+    @staticmethod
+    def disable():
+        """
+        Runs only once and must happen before doing profiling.
+        """
+        if hasattr(HabanaProfile.disable, "invalid"):
+            if not HabanaProfile.disable.invalid:
+                HabanaProfile.HABANA_PROFILE_ENABLED = True
+        else:
+            HabanaProfile.HABANA_PROFILE_ENABLED = True
```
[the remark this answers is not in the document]
ZhaiFeiyue: yes, you are right, I have set the default value of `HabanaProfile.HABANA_PROFILE_ENABLED` to True

ZhaiFeiyue:
@regisss seems that CI is blocked 
image

regisss:
> @regisss seems that CI is blocked image

Yeah weird things have been happening with the CI. I just relaunched it, let's see what happens.

regisss (approved):
Thanks for adding this @ZhaiFeiyue!!

regisss:
@ZhaiFeiyue There is one run I would like to profile but enabling the profiler triggers a memory allocation error (profiling seems to be requiring more memory). Have you already got such issues with profiling?

ZhaiFeiyue:
@regisss yes regisss, profiling requires more memory(save HW event in device HBM), in our local side, fix this issue by:
1. decrease the number of transformer layer  if profiling trasformers based model
2. decrease the BS

Issue #252, opened 2023-05-31
Bad Performance of text-generation with sampling algo


ZhaiFeiyue:
### System Info

```shell
Ubuntu 20.04
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

with `do_sample`, the throughput is 450 tps
```bash
python run_generation.py --model_name_or_path gpt2  --batch_size 1 --max_new_tokens 100 --use_hpu_graphs --prompt "I am a student from" --do_sample
```

without `do_sample`, the throughput is 900tps

```bash
python run_generation.py --model_name_or_path gpt2  --batch_size 1 --max_new_tokens 100 --use_hpu_graphs --prompt "I am a student from"
```

### Expected behavior

decent performance with `do_sample` algo

ZhaiFeiyue:
@regisss  you have asked me before about the bad performance when you runing text-generation with sampling algo, after profiling, I think the issue is bad pipelining of host and device, see the `to_cpu` part with red rectangle.
![image](https://github.com/huggingface/optimum-habana/assets/80079571/42dbed11-3b77-4111-b8d0-885299241d68)

it block host until device finished. so what I think is try to remove it by introducing `check early stop every N steps`.

regisss:
Interesting! Not sure what you mean by "check early stop every N steps" though, could you explain it a bit more?

ZhaiFeiyue:
from the profile we could see there is a `to_cpu` and `to_hpu` and after investigation, I found it is because tensor.prod is a fallback op in HPU side refer code
![image](https://github.com/huggingface/optimum-habana/assets/80079571/db697e68-8c20-4525-9e35-30d35ae63112)

when there is a CPU fallback op, which means we must stop here and wait the tensor from device ready.
the above part is used for early stop checking, so a straight  forward solution is `checking early stop every 5 or 10 step`, not calling above part code every step.

But I have not seen this issue before, and after check the code I found that we are using a old transformers, with which the implementation is
`unfinished_sequences = unfinished_sequences.mul((sum(next_tokens != i for i in eos_token_id)).long())`

ZhaiFeiyue:
@regisss  seems much better now
before
![image](https://github.com/huggingface/optimum-habana/assets/80079571/11400b2d-77fb-4163-a8b5-58672eb60f51)

[...]

PR #254, opened 2023-05-31
Optimize early stop check

ZhaiFeiyue:
# What does this PR do?

Fixes # (issue)

Fix issue #252 

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -589,6 +589,7 @@ class GaudiGenerationMixin(GenerationMixin):
                 synced_gpus=synced_gpus,
                 streamer=streamer,
                 lazy_mode=lazy_mode,
+                ignore_eos=ignore_eos,
                 profiling_warmup_steps=profiling_warmup_steps,
                 profiling_steps=profiling_steps,
                 **model_kwargs,
@@ -1222,6 +1223,7 @@ class GaudiGenerationMixin(GenerationMixin):
         synced_gpus: Optional[bool] = False,
         streamer: Optional["BaseStreamer"] = None,
         lazy_mode: Optional[bool] = False,
+        ignore_eos: Optional[bool] = None,
         profiling_warmup_steps: Optional[int] = 0,
         profiling_steps: Optional[int] = 0,
         **model_kwargs,
@@ -1275,6 +1277,8 @@ class GaudiGenerationMixin(GenerationMixin):
                 through `streamer.put(token_ids)` and the streamer is responsible for any further processing.
             lazy_mode (`bool`, *optional*, defaults to `False`):
                 Whether the run is executed in lazy mode or not (i.e. eager mode).
+            ignore_eos (`bool`, *optional*):
+                Whether to ignore finished sequences (faster in lazy mode and with HPU graphs) or not (eager mode).
             model_kwargs:
                 Additional model specific kwargs will be forwarded to the `forward` function of the model. If model is
                 an encoder-decoder model the kwargs should include `encoder_outputs`.
@@ -1384,6 +1388,7 @@ class GaudiGenerationMixin(GenerationMixin):
             )
 
         # keep track of which sequences are already finished
+        # TODO: no ignore_eos check here since there is a compilation error, will add ignore_eos here if fixed
         unfinished_sequences = torch.ones(input_ids.shape[0], dtype=torch.long, device=input_ids.device)
         hb_profer = HabanaProfile(warmup=profiling_warmup_steps, active=profiling_steps)
         hb_profer.start()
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/generation/utils.py
```diff
@@ -1469,7 +1469,7 @@ def sample(
             )
 
             # if eos_token was found in one sentence, set sentence to finished
-            if eos_token_id_tensor is not None:
+            if not lazy_mode and eos_token_id_tensor is not None:
```
regisss, line 1472: I think we should use `not ignore_eos` instead of `not lazy_mode`, so that it is consistent with [greedy search](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1172) and that users can still do early stopping in lazy mode.

optimum/habana/transformers/generation/utils.py
```diff
@@ -1478,7 +1478,7 @@ def sample(
             #     self.htcore_generation.mark_step()
             hb_profer.step()
             # stop if we exceed the maximum length, or when each sentence is finished (eager mode only)
-            if stopping_criteria(input_ids, scores) or (unfinished_sequences.max() == 0 and not lazy_mode):
+            if (not lazy_mode and unfinished_sequences.max() == 0) or stopping_criteria(input_ids, scores):
```
regisss, line 1481: Same here

regisss (commented):
@ZhaiFeiyue I think we should use `ignore_eos` here instead of `lazy_mode`, see my comment.
Also, we should declare and use `unfinished_sequences` only if `not ignore_eos` [here](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1387) and [there](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1453).

optimum/habana/transformers/generation/utils.py
ZhaiFeiyue: yes, you are right

optimum/habana/transformers/generation/utils.py
ZhaiFeiyue: will change to `not ignore_eos`

ZhaiFeiyue:
> @ZhaiFeiyue I think we should use `ignore_eos` here instead of `lazy_mode`, see my comment. Also, we should declare and use `unfinished_sequences` only if `not ignore_eos` [here](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1387) and [there](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1453).

@regisss PR updated, I only add a comment [here](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1387) and [there](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1453) since there is a compilation crash in my side, I will look into the crash later

regisss:
> > @ZhaiFeiyue I think we should use `ignore_eos` here instead of `lazy_mode`, see my comment. Also, we should declare and use `unfinished_sequences` only if `not ignore_eos` [here](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1387) and [there](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1453).
> 
> @regisss PR updated, I only add a comment [here](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1387) and [there](https://github.com/huggingface/optimum-habana/blob/2cb7646bdaa41e88e05ef2bea18fce43f2faf0a8/optimum/habana/transformers/generation/utils.py#L1453) since there is a compilation crash in my side, I will look into the crash later

So do you prefer that we merge this PR now and if you find a fix to your crash you will submit it in another PR? Or shall we keep this PR open for now?

ZhaiFeiyue:
yes @regisss , we could merge this PR first and then I will submit another PR if find a fix

regisss:
> yes @regisss , we could merge this PR first and then I will submit another PR if find a fix

Sounds good. Could you add `TODO:` at the begining of the two comments for the lines that trigger the crash please? That way we can find them easily if there is no quick fix. And then we merge!

ZhaiFeiyue:
> > yes @regisss , we could merge this PR first and then I will submit another PR if find a fix
> 
> Sounds good. Could you add `TODO:` at the begining of the two comments for the lines that trigger the crash please? That way we can find them easily if there is no quick fix. And then we merge!

@regisss  done and PR updated

regisss (approved):
LGTM!

PR #260, opened 2023-06-07
Fix profiling in GaudiTrainer

regisss:
# What does this PR do?

The profiler is currently stopped at the end of the 1st epoch. This should be done after the last training iteration.
I also added a test to be able in the future to catch this kind of error before merging.

## Before submitting


regisss:
@ZhaiFeiyue I think there is an issue with the way the profiler is stopped in the trainer: it is currently stopped after the 1st epoch, but it should actually be done after the last epoch no? I got an error on the main branch profiling a run with the trainer.

ZhaiFeiyue (approved):
LGTM

PR #261, opened 2023-06-08
fix ds inference with user prompt for text-generation

ZhaiFeiyue:
# What does this PR do?

as title

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/gaudi_spawn.py
+++ b/examples/gaudi_spawn.py
@@ -80,7 +80,9 @@ def main():
 
     # Patch sys.argv
     sys.argv = [args.training_script] + args.training_script_args
-    command_list = [" ".join(sys.argv)]
+    # handling case that argument contains whitespace
+    argv = ['"{}"'.format(arg) if " " in arg else arg for arg in sys.argv]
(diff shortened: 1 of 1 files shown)


regisss:
@ZhaiFeiyue I pushed a commit to handle the case where args are already given in the right format after `gaudi_spawn.py`: `'"..."'`. It's necessary because there are a few places in READMEs where the example command line contains such args (see [here](https://github.com/huggingface/optimum-habana/tree/main/examples/summarization#multi-card-training) for instance).

regisss (approved):
Nice PR, I should have handle this that way from the beginning!

ZhaiFeiyue:
> @ZhaiFeiyue I pushed a commit to handle the case where args are already given in the right format after `gaudi_spawn.py`: `'"..."'`. It's necessary because there are a few places in READMEs where the example command line contains such args (see [here](https://github.com/huggingface/optimum-habana/tree/main/examples/summarization#multi-card-training) for instance).

it always is '"..."' and won't be "'...'" right?

regisss:
> "

Yep

PR #274, opened 2023-06-21
Enable BartForConditionalGeneration inference with greedy search

bhargaveede:
# What does this PR do?
Enable BartForConditionalGeneration  inference on Habana


regisss:
Thanks for this PR @bhargaveede, I'll review it by the end of this week, sorry for the delay

optimum/habana/transformers/training_args.py
```diff
@@ -155,6 +161,21 @@ class GaudiTrainingArguments(TrainingArguments):
...
+    use_kv_cache: bool = field(
+        default=True,
+        metadata={"help": "Use KV caching."},
+    )
+
+    ignore_eos: bool = field(
+        default=False,
+        metadata={"help": "Ignore eos token."},
+    )
+
```
regisss, line 178: These should not be in the training arguments, let's use the argument `generation_config`: https://github.com/huggingface/optimum-habana/blob/1eef41c53ff60a12f6466c2775df03cbc0b3ebea/optimum/habana/transformers/training_args_seq2seq.py#L83

`use_cache` is already in Transformers' `GenerationConfig`: https://github.com/huggingface/transformers/blob/04ab5605fbb4ef207b10bf2772d88c53fc242e83/src/transformers/generation/configuration_utils.py#L38

Let's define a `GaudiGenerationConfig` to add `use_token_idx` and `ignore_eos`. I should probably do it because I'm not sure yet about how to do this properly.

regisss (commented):
@bhargaveede It's not an easy PR because I must think about how to integrate `token_idx` and `ignore_eos` in `GenerationConfig`. I'll try to have something merged in the next few days.

Meanwhile, I encourage you to only put the pieces of code you modified in `modeling_bart.py`. There are many things that are exactly the same as in Transformers and we don't want to have them here to facilitate maintainability.

optimum/habana/transformers/models/bart/modeling_bart.py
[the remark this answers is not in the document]
bhargaveede: @regisss I'll do the changes as prescribed above for modeling_bart.py

optimum/habana/transformers/training_args.py
bhargaveede: @regisss the default configuration we are using in Habana is use_token_idx :True and ignore_eos: False.
So, should I extend GenerationConfig to GaudiGenerationConfig and add use_token_idx and ignore_eos?
(or)
By "I'll try to have something merged in the next few days.", do you mean you'll get those changes merged?

regisss: Yes, I think I'm going to do it because to be honest I'm still not sure of how exactly I would like to have this in the library. I'll have to think about it a little bit, but I'll do it in the next few days :slightly_smiling_face:

regisss:
@bhargaveede Thanks for the changes! I'm going to push a new PR for extending `GenerationConfig` and then I'll review this one.

Regarding your questions:
1. You can just post here how the Gaudi config looks like and I'll upload it, it will be easier that way.
2. Ideally we should rely on the generation config to pass this kind of parameter, but it's true that the seq2seq trainer seems to not use it. Let me check with the Transformers team.

bhargaveede:
@regisss 
1) config is same as t5 https://huggingface.co/Habana/t5/blob/main/gaudi_config.json
So, you can create https://huggingface.co/Habana/Bart/blob/main/gaudi_config.json with the above config itself

regisss:
@bhargaveede I just created it here: https://huggingface.co/Habana/bart/blob/main/gaudi_config.json
I'll update the model card when this PR is merged. But you can already use it of course :slightly_smiling_face:

bhargaveede:
@regisss I saw the PR for extending GenerationConfig. Is it still in draft mode? 
When can I expect you to review this PR?

regisss:
@bhargaveede It's almost done, I'm just going to run a few more tests to validate it. I'll let you know when this is done :slightly_smiling_face: 
So I think I'll review this PR tomorrow.

examples/summarization/README.md
```diff
@@ -225,3 +225,28 @@ python run_summarization.py \
...
+    --dataset_name cnn_dailymail \
+    --dataset_config "3.0.0" \
+    --output_dir /tmp/tst-summarization \
+    --per_device_eval_batch_size 2 \
+    --overwrite_output_dir \
+    --predict_with_generate \
+    --use_habana \
+    --use_lazy_mode \
+    --use_hpu_graphs_for_inference \
+    --gaudi_config_name Habana/t5 \
```
regisss, line 243:
```suggestion
    --gaudi_config_name Habana/bart \
```

examples/summarization/README.md
```diff
@@ -225,3 +225,28 @@ python run_summarization.py \
...
+    --ignore_pad_token_for_loss False \
+    --pad_to_max_length \
+    --num_beams 1 \
+    --use_token_idx
```
regisss, line 247:
```suggestion
    --num_beams 1
```
`token_idx` should be automatically used after adding BART to this list of optimized models: https://github.com/huggingface/optimum-habana/blob/7204bbbb1e28e11c3b99ff6e3b5a2bc9930886dc/optimum/habana/transformers/generation/utils.py#L60

examples/summarization/README.md
```diff
@@ -225,3 +225,28 @@ python run_summarization.py \
...
+```
+
+
+Enable `--use_token_idx` in order to run optimized version of greedy search for Habana
+Only `--num_beams 1` is supported in BART
```
regisss, line 252:
```suggestion
Only `--num_beams 1` is supported for BART.
```

examples/summarization/README.md
```diff
@@ -18,7 +18,7 @@ limitations under the License.
 
 This directory contains examples for finetuning and evaluating transformers on summarization tasks.
 
-`run_summarization.py` is a lightweight example of how to download and preprocess a dataset from the [🤗 Datasets](https://github.com/huggingface/datasets) library or use your own files (jsonlines or csv), then fine-tune T5 on it.
+`run_summarization.py` is a lightweight example of how to download and preprocess a dataset from the [🤗 Datasets](https://github.com/huggingface/datasets) library or use your own files (jsonlines or csv), then fine-tune T5 (or) predict using BART on it.
```
regisss, line 21:
```suggestion
`run_summarization.py` is a lightweight example of how to download and preprocess a dataset from the [🤗 Datasets](https://github.com/huggingface/datasets) library or use your own files (jsonlines or csv), then fine-tune and evaluate T5 (or predict using BART) on it.
```

examples/summarization/run_summarization.py
```diff
@@ -439,6 +439,16 @@ def main():
...
     )
+    is_bart = True if model_args.model_name_or_path in [
+            "facebook/bart-large-mnli",
+            "facebook/bart-large-cnn",
+            "facebook/bart-large",
+            "facebook/bart-base",
+            "facebook/bart-large-xsum",
+        ] else False
+    if is_bart and not (training_args.do_predict or training_args.do_eval):
+        raise ValueError("Training is not yet supported for BART. Eval or predict can be enabled")
```
regisss, line 450:
```suggestion
    is_bart = model.config.model_type == "bart"
    if is_bart and training_args.do_train:
        raise ValueError("Training is not yet supported for BART. Eval or predict can be enabled with `--do_eval` and `--do_predict`.")
```

Let's move this after model initialization at line 467 to have a more robust way of assessing whether the current model type is BART or not.

examples/summarization/run_summarization.py
```diff
@@ -682,7 +692,10 @@ def compute_metrics(eval_preds):
     training_args.generation_num_beams = (
         data_args.num_beams if data_args.num_beams is not None else training_args.generation_num_beams
     )
-
+    if is_bart:
+        training_args.generation_max_new_tokens = (
+            model.generation_config.max_new_tokens if model.generation_config.max_new_tokens is not None else model.generation_config.max_length
+        )
```
regisss, line 698: We can remove this after #293.

optimum/habana/transformers/training_args.py
regisss: We can remove all the modifications in this file.
`ignore_eos` will be automatically set to `True` if lazy mode is on, but you can explicitly set it to `False` in the generation config here: https://github.com/huggingface/optimum-habana/blob/7204bbbb1e28e11c3b99ff6e3b5a2bc9930886dc/examples/summarization/run_summarization.py#L677

optimum/habana/transformers/trainer_seq2seq.py
```diff
@@ -151,7 +151,11 @@ def evaluate(
 
         gen_kwargs = gen_kwargs.copy()
         if gen_kwargs.get("max_length") is None and gen_kwargs.get("max_new_tokens") is None:
-            gen_kwargs["max_length"] = self.args.generation_max_length
+            if self.args.generation_max_new_tokens:
+                gen_kwargs["max_new_tokens"] = self.args.generation_max_new_tokens
+                self.model.generation_config.max_new_tokens = self.args.generation_max_new_tokens
+            else:
+                gen_kwargs["max_length"] = self.args.generation_max_length
```
regisss, line 158: I don't think this is necessary anymore after #293.

optimum/habana/transformers/trainer_seq2seq.py
```diff
@@ -201,7 +205,11 @@ def predict(
...
```
regisss, line 212: Same

optimum/habana/transformers/trainer_seq2seq.py
```diff
@@ -263,12 +271,28 @@ def prediction_step(
...
+        gen_kwargs["use_cache"] = (
+            gen_kwargs["use_cache"] if gen_kwargs.get("use_cache") is not None else self.args.use_kv_cache
+        )
+        gen_kwargs["generation_config"] = (
+            gen_kwargs["generation_config"] if gen_kwargs.get("generation_config") is not None else self.model.generation_config
+        )
+        if self.args.use_token_idx:
+            gen_kwargs["token_idx"] = (
+                torch.tensor(1, device='hpu:0')
+            )
```
regisss, line 286: Same

optimum/habana/transformers/trainer_seq2seq.py
```diff
@@ -285,7 +309,7 @@ def prediction_step(
         # Retrieves GenerationConfig from model.generation_config
         gen_config = self.model.generation_config
         # in case the batch is shorter than max length, the output should be padded
-        if generated_tokens.shape[-1] < gen_config.max_length:
+        if gen_config.max_length is not None and generated_tokens.shape[-1] < gen_config.max_length:
```
regisss, line 312: Same

optimum/habana/transformers/trainer_seq2seq.py
```diff
@@ -320,7 +344,7 @@ def prediction_step(
 
         if has_labels:
             labels = inputs["labels"]
-            if labels.shape[-1] < gen_config.max_length:
+            if gen_config.max_length is not None and labels.shape[-1] < gen_config.max_length:
```
regisss, line 347: Same

optimum/habana/transformers/generation/utils.py
```diff
@@ -170,6 +170,50 @@ def _update_model_kwargs_for_generation(
...
+    def _prepare_decoder_attention_mask(
+        self,
+        max_steps: int,  #current stopping criteria
+        batch_size: int,
+        pad_token_id: int,
+        device: str,
+        dtype: str = torch.int64,
+    ) -> torch.Tensor:
+        x = torch.zeros((batch_size, max_steps), device=device, dtype=dtype)
+        return x.index_fill(1, torch.tensor([0]), pad_token_id)    # First the position with pad_token_id
```
regisss, line 182: I think it's equivalent to what is done [here](https://github.com/huggingface/optimum-habana/blob/7204bbbb1e28e11c3b99ff6e3b5a2bc9930886dc/optimum/habana/transformers/generation/utils.py#L364) in #293 no?

optimum/habana/transformers/generation/utils.py
```diff
@@ -170,6 +170,50 @@ def _update_model_kwargs_for_generation(
...
+        if model_kwargs is not None and "decoder_input_ids" in model_kwargs:
+            return model_kwargs.pop("decoder_input_ids")
+        else:
+            token_idx = model_kwargs.get("token_idx", None)
+            decoder_start_token_id = self._get_decoder_start_token_id(decoder_start_token_id, bos_token_id)
+            if device is None:
+                device = self.device
+            if token_idx is None:
+                max_new_tokens = 1
+            return torch.ones((batch_size, max_new_tokens), dtype=torch.long, device=device) * decoder_start_token_id
```
regisss, line 215: I think that's equivalent to what is done [here](https://github.com/huggingface/optimum-habana/blob/7204bbbb1e28e11c3b99ff6e3b5a2bc9930886dc/optimum/habana/transformers/generation/utils.py#L361) in #293 no?

optimum/habana/transformers/generation/utils.py
```diff
@@ -170,6 +170,50 @@ def _update_model_kwargs_for_generation(
...
+        max_steps: int,  #current stopping criteria
+        batch_size: int,
+        device: str,
+    ) -> Tuple[torch.LongTensor, torch.LongTensor, Optional[torch.Tensor], Optional[torch.Tensor]]:
+        n_heads = self.config.decoder_attention_heads
+        embed_size_per_head = self.config.d_model // n_heads
+        key_states_ = torch.zeros((self.config.decoder_layers, batch_size, n_heads, max_steps, embed_size_per_head), device=device)
+        value_states_ = torch.zeros((self.config.decoder_layers, batch_size, n_heads, max_steps, embed_size_per_head), device=device)
+        past_key_values = tuple((key_states_[i, :, :, :, :], value_states_[i, :, :, :, :], None, None) for i in range(self.config.decoder_layers))
+        return past_key_values
```
regisss, line 195: Past key values should have the right shape since `input_ids` and `attention_mask` are padded no?

optimum/habana/transformers/generation/utils.py
```diff
@@ -346,6 +390,16 @@ def generate(
                 inputs_tensor, generation_config.pad_token_id, generation_config.eos_token_id
             )
 
+        token_idx = model_kwargs.get("token_idx", None)
+        if token_idx is not None:
```
regisss, line 394:
```suggestion
        if generation_config.static_shapes:
```

regisss (commented):
Here is a review that covers everything except the modeling files. I'll take a look at them tomorrow.

For `generation/utils.py`, it may be that I misunderstood a few things that are specific to encoder-decoders. Don't hesitate to let me know if that's the case :slightly_smiling_face:

optimum/habana/transformers/trainer_seq2seq.py
bhargaveede: Sorry for the confusion. 
Are we expecting gen_config to have max_new_tokens (or) max_length?
Current patch works with max_new_tokens and if we remove this then the gen_config won't have max_length and generated_tokens.shape[-1] < gen_config.max_length fails due to NoneType comparison.

If we are expecting max_length to be in gen_config, then changes need be made to make the tensors static according to max_length.

optimum/habana/transformers/trainer_seq2seq.py
bhargaveede: Again, couldn't find anything that propagates max_new_tokens #293

examples/summarization/run_summarization.py
bhargaveede: The Default config doesn't have max_new_tokens. I couldn't find anything in #293 which adds this to config. So, we need a way to provide that.
Either as part of GaudiGenerationConfig (by adding max_new_tokens based on max_length) or the current way

optimum/habana/transformers/training_args.py
bhargaveede: I rebased the changes on top of GaudiGenerationConfig. It's an involved change and I've responded in the comments for clarification. Meanwhile, I'm debugging issues related to accuracy as results are way off when I compare without GaudiGenerationConfig changes. 
Will push the new changes once I get it working with accuracy

examples/summarization/README.md
bhargaveede: Done

examples/summarization/README.md
bhargaveede: Done

optimum/habana/transformers/generation/utils.py
bhargaveede: The one you did is for inputs_tensor and attention_mask which are already padded in pre-processing in tokenzing itself.
So, we might actually doing that padding.

optimum/habana/transformers/generation/utils.py
bhargaveede: No, As far as I understand. Past_key_values are concatened for every token generation. Hence, this is needed for us to get specifical value from index without change in shape

bhargaveede:
@regisss  I resolved the perf and accuracy issues. However, I have to keep few of the max_new_tokens changes.

optimum/habana/transformers/generation/utils.py
```diff
@@ -173,6 +173,50 @@ def _update_model_kwargs_for_generation(
...
+        if model_kwargs is not None and "decoder_input_ids" in model_kwargs:
+            return model_kwargs.pop("decoder_input_ids")
+        else:
+            token_idx = model_kwargs.get("token_idx", None)
+            decoder_start_token_id = self._get_decoder_start_token_id(decoder_start_token_id, bos_token_id)
+            if device is None:
+                device = self.device
+            if token_idx is None:
+                max_new_tokens = 1
+            return torch.ones((batch_size, max_new_tokens), dtype=torch.long, device=device) * decoder_start_token_id
```
regisss, line 218: Could you rebase it on Transformers 4.31? https://github.com/huggingface/transformers/blob/e42587f596181396e1c4b63660abf0c736b10dae/src/transformers/generation/utils.py#L648

optimum/habana/transformers/generation/utils.py
```diff
@@ -380,7 +424,7 @@ def generate(
 
         if generation_config.static_shapes:
             # token_idx is the current index in the generation process, it is incremented each time a new token is generated
-            model_kwargs["token_idx"] = torch.tensor(inputs_tensor.shape[-1], device=inputs_tensor.device)
+            model_kwargs["token_idx"] = torch.tensor(1, device=inputs_tensor.device)
```
regisss, line 427: Why doing this? It will work only if the input is one token.

optimum/habana/transformers/generation/utils.py
```diff
@@ -411,6 +462,7 @@ def generate(
         if self.config.is_encoder_decoder:
             input_ids = self._prepare_decoder_input_ids_for_generation(
                 batch_size,
+                generation_config.max_new_tokens or generation_config.max_length,
```
regisss, line 465: It should only be `max_length` here I think

optimum/habana/transformers/models/bart/modeling_bart.py
```diff
@@ -0,0 +1,936 @@
...
+    shift_tokens_right,
+    _expand_mask,
+)
+logger = logging.get_logger(__name__)
+
+
+# Copied from modeling_bart.py: https://raw.githubusercontent.com/huggingface/transformers/648d0deb1dd28a5d9956e63d8cf8c18f96a6a2aa/src/transformers/models/bart/modeling_bart.py
+# The difference is: modified dynamic shapes to static shapes with `mark_step` for performance improvement.
+
+class gaudi_BartLearnedPositionalEmbedding(nn.Embedding):
```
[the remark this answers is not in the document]
regisss: Feel free to ignore those comments as I just saw that name formats are not consistent in other modeling files. I'll do a PR later to uniformize everything.

optimum/habana/transformers/generation/utils.py
bhargaveede: @regisss  
Current static_shapes implementation always start with token_idx starting with 1.
inputs_tensor is already padded according to max_length.

So, everytime we increment token_idx with +1 and use index_copy and index_add for eliminating cat

optimum/habana/transformers/generation/utils.py
bhargaveede: Done

bhargaveede:
@regisss As informed over email, I've made the changes and updated the PR. Please review the changes

optimum/habana/transformers/generation/utils.py
```diff
@@ -173,6 +186,92 @@ def _update_model_kwargs_for_generation(
...
+
+    def _prepare_decoder_input_ids_for_generation(
+        self,
+        batch_size: int,
+        model_input_name: str,
+        decoder_start_token_id: int = None,
+        bos_token_id: int = None,
+        max_length: int = 1,
+        model_kwargs: Optional[Dict[str, torch.Tensor]] = None,
+        device: torch.device = None,
```
regisss, line 228: Can you reorder the args in the same order as in Transformers please? https://github.com/huggingface/transformers/blob/v4.31.0/src/transformers/generation/utils.py#L648
And we can put `max_length` at the end.

optimum/habana/transformers/generation/utils.py
```diff
@@ -408,12 +518,15 @@ def generate(
...
-            input_ids = self._prepare_decoder_input_ids_for_generation(
-                batch_size,
+            input_ids, _ = self._prepare_decoder_input_ids_for_generation(
+                batch_size=batch_size,
+                model_input_name=model_input_name,
+                model_kwargs=model_kwargs,
                 decoder_start_token_id=generation_config.decoder_start_token_id,
                 bos_token_id=generation_config.bos_token_id,
-                model_kwargs=model_kwargs,
+                max_length=generation_config.max_length or generation_config.max_new_tokens,
```
regisss, line 529: Hmm I have a doubt here. I think `max_length` will be updated according to the value of `max_new_tokens`. I'm going to check.

regisss (commented):
@bhargaveede I left a few minor comments.

I'll test it quickly on my side and we should be able to merge it in the beginning of next week :slightly_smiling_face:

regisss:
@bhargaveede CI failed here: https://github.com/huggingface/optimum-habana/actions/runs/5834012844/job/15822744581?pr=274#step:5:860
You can probably reproduce it locally running:
```py
pip install .[tests]
pytest tests/test_trainer_seq2seq.py -v -s
```

optimum/habana/transformers/generation/utils.py
bhargaveede: @regisss  max_length will be updated but that happens after this part of the code. [https://github.com/huggingface/optimum-habana/pull/274/files/6147008361d20bce0883ec1989d658ddb039f2a9#diff-284d6c109e6788a4405e302113203f6b7e98be68afc7dd3d85b89c6329b9275eR553]

bhargaveede:
> 

I've done the changes and also fixed the CI failure by limited changing stopping_criteria for greedy search

bhargaveede:
@regisss How to trigger run-test. I cannot remove/edit the label to re add it to trigger CI.
Can you review and trigger it?

regisss:
@bhargaveede I just triggered the CI. I'll review your changes later this week.

bhargaveede:
@regisss resolved the latest conflicts.
Could you please re-trigger and review the code once you get some time.

bhargaveede:
@regisss I've changed is_greedy_gen_mode flag to suit latest utils.py.
Earlier test was failing as the flag wasn't changed.

bhargaveede:
@regisss Is there anything pending to be done. If not, When can we expect it to get merged?

regisss:
@bhargaveede I just need to review and test it. Sorry for the delay, I know this must be frustrating but there have been a lot of pretty urgent things to do lately. I'm going to do my best to have it merged by the beginning of next week.

vivekgoe:
@regisss gentle reminder to review and merge this PR :-). We have another one related to T5 evaluation performance in the queue which is dependent on this one. Thanks.

vivekgoe:
@bhargaveede GPT2 fails in local testing. Change not be merged until issue is fixed.

optimum/habana/transformers/generation/utils.py
```diff
@@ -87,6 +89,18 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwa
...
+def _get_stopping_criteria(
+    self, generation_config: GaudiGenerationConfig, stopping_criteria: Optional[StoppingCriteriaList]
+) -> StoppingCriteriaList:
+    criteria = StoppingCriteriaList()
+    if generation_config.max_length is not None:
+        criteria.append(StaticMaxLengthCriteria(generation_config.max_length))
+    if generation_config.max_time is not None:
+        criteria.append(MaxTimeCriteria(max_time=generation_config.max_time))
+    criteria = self._merge_criteria_processor_list(criteria, stopping_criteria)
+    return criteria
```
regisss, line 101: I think this should be defined in `GaudiGenerationMixin` below

optimum/habana/transformers/generation/utils.py
```diff
@@ -87,6 +89,18 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwa
         return self.cur_step >= self.max_steps
 
 
+def _get_stopping_criteria(
+    self, generation_config: GaudiGenerationConfig, stopping_criteria: Optional[StoppingCriteriaList]
+) -> StoppingCriteriaList:
+    criteria = StoppingCriteriaList()
+    if generation_config.max_length is not None:
+        criteria.append(StaticMaxLengthCriteria(generation_config.max_length))
```
regisss, line 97: We should rather have
```suggestion
    if generation_config.max_length is not None:
        if generation_config.static_shapes:
            criteria.append(StaticMaxLengthCriteria(generation_config.max_length))
        else:
            max_position_embeddings = getattr(self.config, "max_position_embeddings", None)
            criteria.append(
                MaxLengthCriteria(
                    max_length=generation_config.max_length,
                    max_position_embeddings=max_position_embeddings,
                )
            )
```
to be able to manage the non-static-shape case.

optimum/habana/transformers/generation/utils.py
```diff
@@ -189,6 +203,122 @@ def _update_model_kwargs_for_generation(
...
+    def _prepare_decoder_attention_mask(
+        self,
+        max_steps: int,  # current stopping criteria
+        batch_size: int,
+        pad_token_id: int,
+        device: str,
+        dtype: str = torch.int64,
+    ) -> torch.Tensor:
+        x = torch.zeros((batch_size, max_steps), device=device, dtype=dtype)
+        return x.index_fill(1, torch.tensor([0]), pad_token_id)  # First the position with pad_token_id
```
regisss, line 215: Very naive question and I'm sure I'm missing something, why should the 1st column of the decoder attention mask be equal to `pad_token_id`? The attention mask should contain only 0s and 1s no?

optimum/habana/transformers/generation/utils.py
```diff
@@ -189,6 +203,122 @@ def _update_model_kwargs_for_generation(
...
+        decoder_start_token_id = self._get_decoder_start_token_id(decoder_start_token_id, bos_token_id)
+        if device is None:
+            device = self.device
+        if token_idx is None:
+            decoder_input_ids_start = (
+                torch.ones((batch_size, 1), dtype=torch.long, device=device) * decoder_start_token_id
+            )
+        else:
+            decoder_input_ids_start = (
+                torch.ones((batch_size, max_length), dtype=torch.long, device=device) * decoder_start_token_id
```
regisss, line 270: Why `(batch_size, max_length)` and not keeping `(batch_size, 1)`?

optimum/habana/transformers/generation/utils.py
```diff
@@ -189,6 +203,122 @@ def _update_model_kwargs_for_generation(
...
+            model_args |= set(inspect.signature(self.forward).parameters)
+        for key, value in model_kwargs.items():
+            if value is not None and key not in model_args:
+                unused_model_args.append(key)
+
+        if unused_model_args:
+            raise ValueError(
+                f"The following `model_kwargs` are not used by the model: {unused_model_args} (note: typos in the"
+                " generate arguments will also show up in this list)"
+            )
```
regisss, line 320: Do we need this? We removed it after Transformers v4.31 was released as the workaround for Llama was not needed anymore.

optimum/habana/transformers/generation/utils.py
```diff
@@ -331,11 +461,10 @@ def generate(
         if generation_config.static_shapes is None:
             generation_config.static_shapes = self.config.model_type in MODELS_OPTIMIZED_WITH_STATIC_SHAPES
         if generation_config.ignore_eos is None:
-            generation_config.ignore_eos = lazy_mode
+            generation_config.ignore_eos = not lazy_mode
```
regisss, line 464: I don't understand this. `ignore_eos` is used to generate tokens till `max_length` is reached, so this should be done in lazy mode. Why inverting it?

optimum/habana/transformers/generation/utils.py
```diff
@@ -425,6 +566,7 @@ def generate(
                 decoder_start_token_id=generation_config.decoder_start_token_id,
                 bos_token_id=generation_config.bos_token_id,
                 device=inputs_tensor.device,
+                max_length=generation_config.max_length or generation_config.max_new_tokens,
```
regisss, line 569:
```suggestion
                max_length=generation_config.max_length,
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -382,16 +511,28 @@ def generate(
                 inputs_tensor, generation_config.pad_token_id, generation_config.eos_token_id
             )
 
-        if generation_config.static_shapes:
+        if self.config.is_encoder_decoder and generation_config.static_shapes:
```
regisss, line 514: Inputs and attention mask for decoder-only models will not be padded if you do this, is it intended?

optimum/habana/transformers/generation/utils.py
```diff
@@ -480,16 +622,14 @@ def generate(
...
-                    "You need to set `max_new_tokens` in your generation configuration to use static shapes."
-                )
+        if self.config.is_encoder_decoder and generation_mode == GenerationMode.GREEDY_SEARCH:
+            stopping_criteria = _get_stopping_criteria(
+                self, generation_config=generation_config, stopping_criteria=stopping_criteria
+            )
+        else:
+            stopping_criteria = self._get_stopping_criteria(
+                generation_config=generation_config, stopping_criteria=stopping_criteria
+            )
```
regisss, line 632: We should not have two different methods with the same name. All custom logic should be implemented in `self._get_stopping_criteria`.

regisss (commented):
@bhargaveede I left several comments:
- Regarding generation, I'm unsure about a few changes and asked for clarification. 
- BART modeling changes look good to me, related comments are very minor.

I'm also going to run the regression tests on your branch and will let you know tomorrow if there is anything.

regisss:
@bhargaveede Could you rebase your branch on main so that the latest CI tests are available please? I added regression tests for generation with all the models we optimized and I would like to run them :slightly_smiling_face:

bhargaveede:
@regisss, I'll rebase and address review comments whichever are trivial. You can run regression tests once that's done

bhargaveede:
@regisss  I've rebased it. Can you trigger regression tests

regisss:
I see a few throughput regressions for some models. I'll look into it tomorrow or Friday.
If you want to run the tests, you can do it with:
```
pytest tests/test_text_generation_example.py -v -s
```

bhargaveede:
@

> I see a few throughput regressions for some models. I'll look into it tomorrow or Friday. If you want to run the tests, you can do it with:
> 
> ```
> pytest tests/test_text_generation_example.py -v -s
> ```

I ran these tests locally. I got only 5/9 passed in main
I faced an issue with Falcon model and after fixing it 5/9 are passing in dev/bart_inference branch.

Is there any specific requirement we need to ensure before running the above tests?

optimum/habana/transformers/generation/utils.py
bhargaveede: I will move the custom logic to _gaudi_get_stopping_criteria and add it to GaudiGenerationMixin

optimum/habana/transformers/generation/utils.py
bhargaveede: No. I've made the change to avoid inputs and attention_mask padding in case of encoder_decoder models as inputs are already padded. Where as in decoder models, Inputs are not padded

optimum/habana/transformers/generation/utils.py
bhargaveede: What should be the default behavior? 
I inverted it as I was checking the perf and in earlier case when ignore_eos wasn't there in generation config. Ignore_eos is False.

optimum/habana/transformers/generation/utils.py
bhargaveede: Not needed

optimum/habana/transformers/generation/utils.py
bhargaveede: Will do the modifications.

optimum/habana/transformers/generation/utils.py
bhargaveede: It should be 1. In case of bart, pad_token is equal to 1. Will change it.

optimum/habana/transformers/generation/utils.py
bhargaveede: We're padding and using index_copy to copy the token instead of cat operation after the generation of token.
https://github.com/bhargaveede/optimum-habana/blob/dev/bart_inference/optimum/habana/transformers/generation/utils.py#L1274

optimum/habana/transformers/generation/utils.py
```diff
@@ -383,16 +495,32 @@ def generate(
...
+                        generation_config.max_length,
+                        inputs_tensor.shape[0],
+                        generation_config.pad_token_id,
+                        inputs_tensor.device,
+                    )
+
+            # if model_kwargs.get("past_key_values", None) is None and generation_config.use_cache:
+            #     model_kwargs["past_key_values"] = self._prepare_past_key_values(
+            #         generation_config.max_length, inputs_tensor.shape[0], inputs_tensor.device
+            #     )
```
regisss, line 523: To remove?
Also, do we still need `_prepare_past_key_values` if this is commented? I didn't find any other place where it's called.

optimum/habana/transformers/generation/utils.py
```diff
@@ -189,6 +192,116 @@ def _update_model_kwargs_for_generation(
 
         return model_kwargs
 
+    def _prepare_decoder_attention_mask(
+        self,
+        max_steps: int,  # current stopping criteria
+        batch_size: int,
+        pad_token_id: int,
+        device: str,
+        dtype: str = torch.int64,
```
regisss, line 201: `bool` should be enough for an attention mask no?

optimum/habana/transformers/generation/utils.py
```diff
@@ -189,6 +192,116 @@ def _update_model_kwargs_for_generation(
...
+    def _prepare_decoder_input_ids_for_generation(
+        self,
+        batch_size: int,
+        model_input_name: str,
+        model_kwargs: Dict[str, torch.Tensor],
+        decoder_start_token_id: int = None,
+        bos_token_id: int = None,
+        device: torch.device = None,
+        max_length: int = 1,
+    ) -> Tuple[torch.LongTensor, Dict[str, torch.Tensor]]:
```
regisss, line 235: Since this method overrides the one of Transformers, could you write a short docstring that tells the difference with the Transformers one? It can be just one sentence. It seems to be linked to the initialization of `decoder_input_ids_start` if I'm not mistaken.

optimum/habana/transformers/generation/utils.py
```diff
@@ -189,6 +192,116 @@ def _update_model_kwargs_for_generation(
...
+        self, generation_config: GaudiGenerationConfig, stopping_criteria: Optional[StoppingCriteriaList]
+    ) -> StoppingCriteriaList:
+        criteria = StoppingCriteriaList()
+        if generation_config.max_length is not None:
+            if (
+                generation_config.static_shapes
+                and self.config.is_encoder_decoder
+                and self._get_generation_mode(generation_config) == GenerationMode.GREEDY_SEARCH
+            ):
+                criteria.append(StaticMaxLengthCriteria(generation_config.max_length))
```
regisss, line 291: Do we really need this? We already add the same criteria here: https://github.com/huggingface/optimum-habana/blob/8d1a432f99f2838a91cea6aa87052ef42dba077c/optimum/habana/transformers/generation/utils.py#L505

optimum/habana/transformers/generation/utils.py
regisss: Ah okay I see. So, now, `ignore_eos` should be initialized to `None` in the generation config if not specified, and here we should set it to the same value as `lazy_mode` because in lazy mode we don't want to stop when sequences are finished.
This was actually the reason why regression tests failed. Setting it back to `lazy_mode` gives the expected numbers.

optimum/habana/transformers/generation/utils.py
```diff
@@ -383,16 +495,32 @@ def generate(
...
-            )
-            if model_kwargs["attention_mask"] is not None:
-                model_kwargs["attention_mask"] = torch.nn.functional.pad(
-                    model_kwargs["attention_mask"], (0, generation_config.max_new_tokens), value=0
+            # In encoder_decoder models, Inputs are already padded
+            if not self.config.is_encoder_decoder:
+                # token_idx is the current index in the generation process, it is incremented each time a new token is generated
+                model_kwargs["token_idx"] = torch.tensor(inputs_tensor.shape[-1], device=inputs_tensor.device)
+                inputs_tensor = torch.nn.functional.pad(
+                    inputs_tensor, (0, generation_config.max_length), value=generation_config.pad_token_id
```
regisss, line 504:
```suggestion
                    inputs_tensor, (0, generation_config.max_new_tokens), value=generation_config.pad_token_id
```
If we pad with `max_length`, the size of the tensor will be `2*input_length + max_new_tokens`.

optimum/habana/transformers/generation/utils.py
```diff
@@ -383,16 +495,32 @@ def generate(
...
                 )
+                if model_kwargs["attention_mask"] is not None:
+                    model_kwargs["attention_mask"] = torch.nn.functional.pad(
+                        model_kwargs["attention_mask"], (0, generation_config.max_length), value=0
```
regisss, line 508:
```suggestion
                        model_kwargs["attention_mask"], (0, generation_config.max_new_tokens), value=0
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -189,6 +192,116 @@ def _update_model_kwargs_for_generation(
...
+        decoder_start_token_id = self._get_decoder_start_token_id(decoder_start_token_id, bos_token_id)
+        if device is None:
+            device = self.device
+        if token_idx is None:
+            decoder_input_ids_start = (
+                torch.ones((batch_size, 1), dtype=torch.long, device=device) * decoder_start_token_id
+            )
+        else:
+            decoder_input_ids_start = (
+                torch.ones((batch_size, max_length), dtype=torch.long, device=device) * decoder_start_token_id
```
regisss, line 259:
```suggestion
                torch.ones((batch_size, self.generation_config.max_length), dtype=torch.long, device=device) * decoder_start_token_id
```
This should work, and that way we can remove the extra `max_length` argument to this method.

regisss (commented):
@bhargaveede So the regression is due to `generation_config.ignore_eos = not lazy_mode` as `not` should not be there. Here are several comments.
After they are taken care of, I think we should be able to merge!

optimum/habana/transformers/generation/utils.py
bhargaveede: Also, Shouldn't the behavior be controlled from command line (or) generation_config by setting ignore_eos value rather than depending on lazy mode value? 
As the behavior is now conflicting with CPU/GPU runs.

regisss: Sure, you can add a condition to set it to False for encoder-decoder models. Could you provide some throughput numbers for both configurations (True and False).
It's already possible to set it within the generation config. This is just a default initialization if it has not been set.

regisss (approved):
LGTM! Thanks for this big PR @bhargaveede. Sorry for being a bit slow to review, it's a big piece, encoder-decoders have their own particularities and I didn't want to have any regression for decoders.

CI failed because it cannot launch an instance in AWS us-east-1. Let's see if tomorrow it's better, otherwise I'll run it locally and merge after.

We will have to add a regression test for BART too, but let's do that in another PR.

regisss:
I ran the CI on an instance of mine and it passed, merging.

@bhargaveede Could you add a test for generation with BART in another PR please? Maybe in [this file](https://github.com/huggingface/optimum-habana/blob/main/tests/test_text_generation_example.py), or another one if it's more convenient.

regisss:
@bhargaveede We have a regression on `test_run_seq2seq_qa_t5-small_multi_card`. The F1 score is much lower than before:
```
AssertionError: 48.609943492751356 not greater than or equal to 64.0629 : for metric eval_f1.
```

Maybe some padding is done for non-optimized models like T5 at some point?

bhargaveede:
@regisss Do you have the test command for the same. I'll check it

bhargaveede:
@regisss Only thing I could think of without running is probably ignore_eos as it's set to false in case of encoder decoder model which t5-small is.
https://github.com/huggingface/optimum-habana/pull/274/files#diff-284d6c109e6788a4405e302113203f6b7e98be68afc7dd3d85b89c6329b9275eR428

regisss:
@bhargaveede I'm on my phone but you can get the name of the test with: pytest tests/test_examples.py - k "t5" --co
This will return 2 tests, the one with regression is the seq2seq QA test.

PR #275, opened 2023-06-21
Remove use_fused_adam config in clip

x574chen:
# What does this PR do?

I am verifying CLIP using latest optimum habana. Disable fused adam is no long necessary after enabling hpu graph for training and media dataloader.
image


regisss (approved):
LGTM!

PR #279, opened 2023-06-25
Gaudi2 support opt-66b(DS mode)

ZhaiFeiyue:
# What does this PR do?

current text-generation only support Bloom-176b, but not support opt-66b, since opt-66b can not fit into 1 gaudi2 device(96GB). 
with this PR opt-66B could run with 2 Gaudi2 cards

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/checkpoint_utils.py
+++ b/examples/text-generation/checkpoint_utils.py
@@ -59,13 +59,6 @@ def write_checkpoints_json(model_name_or_path, local_rank, checkpoints_json):
             json.dump(data, fp)
 
 
-def model_is_bloom(config):
-    """
-    Checks if the given config belongs to a BLOOM-like model.
-    """
-    return config.model_type == "bloom"
-
-
 def get_optimized_model_name(config):
     model_names = ["bloom", "gpt2", "opt", "gptj", "gpt_neox"]
     for model_name in model_names:
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -180,7 +180,7 @@ def main():
             with deepspeed.OnDevice(dtype=model_dtype, device="meta"):
                 model = AutoModelForCausalLM.from_config(config, torch_dtype=model_dtype)
         else:
-            with deepspeed.OnDevice(dtype=model_dtype, device=args.device):
+            with deepspeed.OnDevice(dtype=model_dtype, device="cpu"):
                 model = AutoModelForCausalLM.from_pretrained(args.model_name_or_path, torch_dtype=model_dtype)
         model = model.eval()
 
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -151,11 +151,6 @@ def main():
 
     world_size, rank, args.local_rank = initialize_distributed_hpu()
 
-    # Tweak generation so that it runs faster on Gaudi
-    from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
-
-    adapt_transformers_to_gaudi()
-
     if use_deepspeed:
         # Check if DeepSpeed is installed
         from transformers.deepspeed import is_deepspeed_available
@@ -172,6 +167,11 @@ def main():
         logger.info("DeepSpeed is enabled.")
     else:
         logger.info("Single-device run.")
+
+    # Tweak generation so that it runs faster on Gaudi
+    from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
+
+    adapt_transformers_to_gaudi()
 
     # Set seed before initializing model.
     from optimum.habana.utils import set_seed
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -151,6 +151,11 @@ def main():
 
     world_size, rank, args.local_rank = initialize_distributed_hpu()
 
+    # Tweak generation so that it runs faster on Gaudi
+    from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
+
+    adapt_transformers_to_gaudi()
+
     if use_deepspeed:
         # Check if DeepSpeed is installed
         from transformers.deepspeed import is_deepspeed_available
@@ -167,11 +172,6 @@ def main():
         logger.info("DeepSpeed is enabled.")
     else:
         logger.info("Single-device run.")
-
-    # Tweak generation so that it runs faster on Gaudi
-    from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
-
-    adapt_transformers_to_gaudi()
 
     # Set seed before initializing model.
     from optimum.habana.utils import set_seed
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -25,7 +25,7 @@ import time
 
 import torch
 import torch.nn.functional as F
-from checkpoint_utils import get_ds_injection_policy, model_is_bloom, model_is_optimized, write_checkpoints_json
+from checkpoint_utils import get_ds_injection_policy, model_is_optimized, write_checkpoints_json
 from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
 from transformers.generation import GenerationConfig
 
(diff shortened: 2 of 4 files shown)


ZhaiFeiyue:
@regisss opt-66B cannot run on 1 Gaudi2 and also cannot run with DS 8x, since in [here](https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/run_generation.py#L183C61-L183C61) model weights will be transferred to device before weights sharding in `deepspeed.init_inference()`, I have tried to  change opt path to bloom path, but there are confliction, such as 
- deepspeed will hijack nn.Embedding` which is also be done in `modeling_opt.py` 
- in `_attn`, the device of `caucal_mask`  is `meta` which lead to a runtime mismatch of tensors

so I change the device from `args.device` to `cpu`, means load weight to host and then do weights sharding in `deepspeed.init_inference()`, which works well when running opt-66b on 2x
But see [here](https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/run_generation.py#L183C61-L183C61), if device is `cpu` and when running with 8x, means loading weights from disk to memory 8 times, waste lots of host memory, right?

regisss:
Thanks for this PR @ZhaiFeiyue, I'll review this PR by the end of this week!

regisss:
@ZhaiFeiyue So I investigated this a bit, and with a few changes to DeepSpeed I got it to work:
- replace `torch.half` by `torch.bfloat16` [here](https://github.com/HabanaAI/DeepSpeed/blob/410b3dbb74c7d266eba71aadf26ee616c2882040/deepspeed/module_inject/layers.py#L28) and [there](https://github.com/HabanaAI/DeepSpeed/blob/410b3dbb74c7d266eba71aadf26ee616c2882040/deepspeed/module_inject/layers.py#L62)
- replacing the forward in [`OPTEmbedding`](https://github.com/HabanaAI/DeepSpeed/blob/410b3dbb74c7d266eba71aadf26ee616c2882040/deepspeed/module_inject/layers.py#L84) with more or less the same as we have [here](https://github.com/huggingface/optimum-habana/blob/1eef41c53ff60a12f6466c2775df03cbc0b3ebea/optimum/habana/transformers/models/opt/modeling_opt.py#L18)

Do you think we can push these changes to Habana's DeepSpeed fork?
Otherwise we can do some monkey patching in a similar way to what we do to override Transformers modeling. WDYT?

ZhaiFeiyue:
> 

@regisss thanks for your investigation :smile:, with the changes above you can run text-generation with opt-66b ds 8x in the way like Bloom right,? since bloom is special handled in `run_generation.py` (I guess you have another PR that makes changes in `run_generation.py`).
or base on this PR with your changes, you can run opt-66b ds 8x in your side?

regisss:
> > 
> 
> @regisss thanks for your investigation 😄, with the changes above you can run text-generation with opt-66b ds 8x in the way like Bloom right,? since bloom is special handled in `run_generation.py` (I guess you have another PR that makes changes in `run_generation.py`). or base on this PR with your changes, you can run opt-66b ds 8x in your side?

Yes, it follows the same path as BLOOM in the script. I'm going to push a new commit so that you can test it.
But the question is: should these changes be directly in Habana's DeepSpeed fork or just in Optimum Habana?

ZhaiFeiyue:
cool @regisss, for your question I prefer we do these in optimum-habana side, since Habana DS is a fork version and  should not involve too much model special changes.

ZhaiFeiyue:
@regisss very clean changes 👍
but there is a error in my side, seems that the `Linearlayer`  has not been replaced.
image

any ideas or I missed something?

regisss:
@ZhaiFeiyue Yeah I have the same error, I need to see where it comes from exactly. When doing the changes directly in DeepSpeed it was working.
I've not got time to investigate this much yet, I'll let you know when it is fixed :slightly_smiling_face:

regisss:
@ZhaiFeiyue The model is now loaded correctly but results are weird:
```
Input/outputs:
----------------------------------------------------------------------
input 1: ('DeepSpeed is a machine learning framework',)
output 1: ('DeepSpeed is a machine learning framework',)
----------------------------------------------------------------------
```
It's like it is not doing anything.

ZhaiFeiyue:
@regisss  yes, same in my side

ZhaiFeiyue:
outputs are always 50272
image

ZhaiFeiyue:
weights seems that not loaded correctly
 the red part is dummy weights
image

regisss:
Weird :thinking: 
I'll be off for a few hours, feel free to keep investigating @ZhaiFeiyue. Otherwise I'll get back to it later.

regisss:
Weird :thinking: 
I'll be off for a few hours, feel free to keep investigating @ZhaiFeiyue. Otherwise I'll get back to it later.

regisss:
@ZhaiFeiyue There must be something wrong in the injection policy. Maybe because of the way we write the JSON checkpoint here: https://github.com/huggingface/optimum-habana/blob/b1f0bd7e436be6d57a893eee7e34647a5a86ca3f/examples/text-generation/checkpoint_utils.py#L59
I also see that auto-injection will be merged at some point (see [here](https://github.com/HabanaAI/DeepSpeed/blob/410b3dbb74c7d266eba71aadf26ee616c2882040/deepspeed/inference/engine.py#L67)), so if it's planned for v1.11.0 maybe it's better to just wait for it and keep your initial change?

ZhaiFeiyue:
@regisss agree with you, I will open a new PR and we could keep your changes in this PR

ZhaiFeiyue:
new PR is #285

ZhaiFeiyue:
@regisss I finally got time to debug OPT-66b now 😄, the weights is not loaded correctly because the name mismatch, your injection works well. I add new changes here
image

if with opt-125m and 66b the weights name starting from `model.` but 13b starting from `decoder.` 

see [here](https://github.com/HabanaAI/DeepSpeed/blob/410b3dbb74c7d266eba71aadf26ee616c2882040/deepspeed/module_inject/load_checkpoint.py#L347C25-L347C25) the prefix with level=0 will be stripped, which lead to the name mismatch [here](https://github.com/HabanaAI/DeepSpeed/blob/410b3dbb74c7d266eba71aadf26ee616c2882040/deepspeed/module_inject/load_checkpoint.py#L38)

I have tested 125m 13b, will test 66b later when resource available.

regisss:
@ZhaiFeiyue Nice! Let me know if OPT-66b works  :slightly_smiling_face:

ZhaiFeiyue:
@regisss  bad news 😢  66b does not work, but  comment this [line](https://github.com/HabanaAI/DeepSpeed/blob/410b3dbb74c7d266eba71aadf26ee616c2882040/deepspeed/module_inject/load_checkpoint.py#L347) it works need do more investigation.

ZhaiFeiyue:
@regisss I have checked all the weights' name of bloom models(from 560m to 176b), all of them are same.
but for OPT, they are different as I mentioned before
- OPT-125m OPT-66b prefix=`model.` and  name starting with `model.`
- OPT-13b, OPT-30b prefix = '' and name starting with `decoder.`

transformers `from_pretrained()` could handle both the above cases [from here](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L3029) but in DS, the prefix are always removed I think [from here](https://github.com/HabanaAI/DeepSpeed/blob/410b3dbb74c7d266eba71aadf26ee616c2882040/deepspeed/module_inject/load_checkpoint.py#L347), because ckpt_type should be `pp` and level should be `0`, so when running OPT-125m and 66b the name is mismatch(in the checkpints bin file it starting with `model.` but in DS the name is starting `decoder.` since `model.` is removed) and weight cannot be loaded.

base on the above analysis, what I think is `is there any possible that we create a offline weights converter tool that to remove prefix for 125m and 66b? `

correct me if some wrong.

ZhaiFeiyue:
@regisss  that's great if could fix in Transformers team.

ZhaiFeiyue:
@regisss let's close this PR, because the changes for OPT should be in Habana DS like [official DS](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/module_inject/replace_module.py#L573C1-L588C1)

regisss:
> @regisss let's close this PR, because the changes for OPT should be in Habana DS like [official DS](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/module_inject/replace_module.py#L573C1-L588C1)

Sounds good!

PR #285, opened 2023-07-04
support opt-66b

ZhaiFeiyue:
# What does this PR do?

temporary enable opt-66b(ds mode).

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -198,6 +198,7 @@ def main():
             with deepspeed.OnDevice(dtype=model_dtype, device="meta"):
                 model = AutoModelForCausalLM.from_config(config, torch_dtype=model_dtype)
         else:
+            # TODO: revisit placement on CPU when auto-injection is possible
             with deepspeed.OnDevice(dtype=model_dtype, device="cpu"):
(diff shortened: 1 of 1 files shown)


examples/text-generation/run_generation.py
```diff
@@ -198,7 +198,7 @@ def main():
             with deepspeed.OnDevice(dtype=model_dtype, device="meta"):
                 model = AutoModelForCausalLM.from_config(config, torch_dtype=model_dtype)
         else:
-            with deepspeed.OnDevice(dtype=model_dtype, device=args.device):
+            with deepspeed.OnDevice(dtype=model_dtype, device="cpu"):
```
regisss, line 201:
```suggestion
            # TODO: revisit placement on CPU when auto-injection is possible
            with deepspeed.OnDevice(dtype=model_dtype, device="cpu"):
```

regisss (approved):
@ZhaiFeiyue LGTM! I just left one comment to add a "TODO" comment so that we don't forget to revisit this in the future.

FYI, it runs well on 2 devices but on 8 devices all my processes are killed because I guess they ran out of host memory XD

PR #296, opened 2023-07-14
Add llama model change

yeonsily:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/habana/transformers/generation/utils.py
```diff
@@ -179,6 +180,35 @@ def _update_model_kwargs_for_generation(
 
         return model_kwargs
 
+    def _validate_model_kwargs(self, model_kwargs: Dict[str, Any]):
```
regisss, line 183: The issue with `token_type_ids` returned by Llama's tokenizer has been solved in https://github.com/huggingface/transformers/pull/24042, which will be part of the next release of Transformers. So let's add the following comment to not forget it:
```suggestion
    # TODO: remove this method when Transformers v4.31 is released since it solves the issue with Llama
    def _validate_model_kwargs(self, model_kwargs: Dict[str, Any]):
```

regisss (commented):
@yeonsily LGTM!

I just have a minor remark about adding a comment in the code that will have to be updated when Transformers v4.31 is released.

Also, I see that the code style check failed. Could you run
```bash
pip install black ruff
make style
```
from the root of the repo please?

optimum/habana/transformers/generation/utils.py
yeonsily: @regisss  Thank you for the update! Will add that.
Btw I saw 'make style' update other code format not just my changes in my files. Let me try again.

regisss (approved):
LGTM!

PR #302, opened 2023-07-20
remove kv cache wa

ZhaiFeiyue:
# What does this PR do?

as title, `past_key.index_add_(2, token_idx - 1, key - torch.index_select(past_key, 2, token_idx - 1))` is used to WA HPU bug and has been fixed from  1.10.0, so could use `index_copy_` directly now

Fixes # (issue)

## Before submitting


ZhaiFeiyue:
@regisss HabanaAI release 1.11.0 is coming soon and we have lots of optimizations for bloom, which could be also applied to other TG model, so let's remove the old WA first.

regisss (approved):
LGTM!

PR #313, opened 2023-08-01
internal

fxmarty:

PR #317, opened 2023-08-02
Trigger workflows for external PRs on label

fxmarty:
Revert https://github.com/huggingface/optimum-habana/pull/313 as it is not a good solution given that it is not possible to disable notifications for review requests in github: https://github.com/orgs/community/discussions/14564. We do not want to get spammed by review requests at every commit.

The `authorize` job is run only when:
* The PR does not come from a fork ==> runs automatically ==> anybody with write access can still run workflows automatically
* The workflow is triggered not from `pull_request_target` event (e.g. rather `workflow_dispatch` that only people with write access can trigger)
* The PR is labeled `run-tests`, which triggers a `labeled` action type.

Note: this workflow is not susceptible to race condition on commits being added after the PR being labeled, given that we use `if: (github.event.action == 'labeled' && github.event.label.name == 'run-test')` and NOT `if: contains(github.event.pull_request.labels.*.name, 'safe to test')` (which is dangerous). Ref: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

All following jobs require the `authorize` job.

I think it is a decent enough compromise to generalize this to other subpackages.

### DIFF
--- a/.github/workflows/fast_tests.yml
+++ b/.github/workflows/fast_tests.yml
@@ -2,11 +2,11 @@ name: Unit and integration tests
 
 
 # WARNING: As this workflow supports the pull_request_target event, please exercise extra care when editing it.
-# Reference: https://gist.github.com/fxmarty/fba1ea3b2ebbf9a75cca71c08f05a85b
 on:
   workflow_dispatch:
   pull_request_target:
     branches: [ main ]
+    types: [ opened, synchronize, reopened, labeled ]
     paths:
       - "optimum/**.py"
       - "tests/**.py"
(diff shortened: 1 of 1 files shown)


fxmarty:
@regisss Let me know what you think.

regisss (approved):
@fxmarty Thanks, it looks good!

So, if I understood correctly, to trigger this workflow for external contributors I'll have to add the label "run-test" right?

regisss:
@fxmarty And I guess we don't need the internal and external environments anymore right?

fxmarty:
@regisss Yes exactly!

PR #324, opened 2023-08-09
Fixes for streaming dataset mode

MohitIntel:
Fixes for streaming dataset mode
 
A customer reported seeing errors when using --streaming.
Made a fix to get rid of TypeError: object of type 'IterableDataset' has no len().
Updated Readme with an example command that demonstrates correct use of --streaming. with --max_steps specified.

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/language-modeling/README.md
+++ b/examples/language-modeling/README.md
@@ -69,6 +69,27 @@ python run_clm.py \
     --throughput_warmup_steps 3
 ```
 
+To run with streaming dataset, use the '--streaming' flag with '--max_steps' and optionally'--max_eval_samples' (only when evaluation is enabled) specified like so:
+
+```bash
+python run_clm.py \
+    --model_name_or_path gpt2 \
+    --dataset_name wikitext \
+    --dataset_config_name wikitext-2-raw-v1 \
+    --per_device_train_batch_size 4 \
+    --per_device_eval_batch_size 4 \
+    --do_train \
+    --output_dir /tmp/test-clm \
+    --gaudi_config_name Habana/gpt2 \
+    --use_habana \
+    --use_lazy_mode \
+    --use_hpu_graphs_for_inference \
+    --throughput_warmup_steps 3 \
+    --streaming \
+    --max_steps 1000 \
+    --do_eval \
+    --max_eval_samples 200
+```
 
 ### Multi-card Training
 
--- a/examples/language-modeling/README.md
+++ b/examples/language-modeling/README.md
@@ -69,28 +69,6 @@ python run_clm.py \
     --throughput_warmup_steps 3
 ```
 
-To run with streaming dataset, use the '--streaming' flag with '--max_steps' and optionally'--max_eval_samples' (only when evaluation is enabled) specified like so:
-
-```bash
-python run_clm.py \
-    --model_name_or_path gpt2 \
-    --dataset_name wikitext \
-    --dataset_config_name wikitext-2-raw-v1 \
-    --per_device_train_batch_size 4 \
-    --per_device_eval_batch_size 4 \
-    --do_train \
-    --output_dir /tmp/test-clm \
-    --gaudi_config_name Habana/gpt2 \
-    --use_habana \
-    --use_lazy_mode \
-    --use_hpu_graphs_for_inference \
-    --throughput_warmup_steps 3 \
-    --streaming \
-    --max_steps 1000 \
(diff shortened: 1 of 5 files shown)


examples/language-modeling/README.md
```diff
@@ -69,6 +69,27 @@ python run_clm.py \
...
+    --gaudi_config_name Habana/gpt2 \
+    --use_habana \
+    --use_lazy_mode \
+    --use_hpu_graphs_for_inference \
+    --throughput_warmup_steps 3 \
+    --streaming \
+    --max_steps 1000 \
+    --do_eval \
+    --max_eval_samples 200
+```
```
regisss, line 92: Let's move this paragraph into the "Streaming" section of the README: https://github.com/huggingface/optimum-habana/tree/main/examples/language-modeling#streaming

examples/language-modeling/run_clm.py
```diff
@@ -547,15 +547,15 @@ def group_texts(examples):
...
             train_dataset = train_dataset.select(range(max_train_samples))
 
     if training_args.do_eval:
         if "validation" not in tokenized_datasets:
             raise ValueError("--do_eval requires a validation dataset")
         eval_dataset = lm_datasets["validation"]
-        if data_args.max_eval_samples is not None:
+        if data_args.max_eval_samples is not None and not data_args.streaming:
             max_eval_samples = min(len(eval_dataset), data_args.max_eval_samples)
             eval_dataset = eval_dataset.select(range(max_eval_samples))
```
regisss, line 560: `max_eval_samples` will actually not have any effect in streaming mode since this line cannot be computed with an iterable dataset.

examples/language-modeling/run_clm.py
```diff
@@ -602,10 +602,13 @@ def compute_metrics(eval_preds):
...
-            data_args.max_train_samples if data_args.max_train_samples is not None else len(train_dataset)
-        )
-        metrics["train_samples"] = min(max_train_samples, len(train_dataset))
+        if data_args.streaming:
+            train_samples = training_args.max_steps * training_args.per_device_train_batch_size
+        else:
+            train_samples = len(train_dataset)
+
+        max_train_samples = data_args.max_train_samples if data_args.max_train_samples is not None else train_samples
+        metrics["train_samples"] = min(max_train_samples, train_samples)
```
regisss, line 611: Since `max_train_samples` cannot be used in streaming mode, I think we should rather have:
```suggestion
        if data_args.streaming:
            metrics["train_samples"] = training_args.max_steps * training_args.per_device_train_batch_size
        else:
            max_train_samples = data_args.max_train_samples if data_args.max_train_samples is not None else len(train_dataset)
            metrics["train_samples"] = min(max_train_samples, len(train_dataset))
```

examples/language-modeling/run_clm.py
```diff
@@ -614,11 +617,16 @@ def compute_metrics(eval_preds):
...
 
-        max_eval_samples = data_args.max_eval_samples if data_args.max_eval_samples is not None else len(eval_dataset)
-        metrics["eval_samples"] = min(max_eval_samples, len(eval_dataset))
+        if data_args.streaming:
+            eval_samples = training_args.max_steps * training_args.per_device_eval_batch_size
+        else:
+            eval_samples = len(eval_dataset)
+
+        max_eval_samples = data_args.max_eval_samples if data_args.max_eval_samples is not None else eval_samples
+        metrics["eval_samples"] = min(max_eval_samples, eval_samples)
```
regisss, line 628: Unfortunately the evaluation loop will loop over the whole evaluation set independently of `max_steps` or `max_eval_samples` (`max_steps` only affects the training loop). So we cannot compute `metrics["eval_samples"]` in streaming mode.

regisss (commented):
@MohitIntel Good catch! I see the behavior is the same in Transformers but I agree with you it should be clearer :slightly_smiling_face: 

I left a few comments which are related to `max_train_samples` and `max_eval_samples` that cannot be used in streaming mode. In particular, the whole evaluation set will be looped over independently of `max_eval_samples` in streaming mode.

I think we should raise an error if `streaming==True` and `max_steps is None` with a message suggesting to set `max_steps` to get training working in streaming mode. Not all people will read the README.

regisss (approved):
LGTM!

I pushed a couple of commits to apply the same changes to `run_mlm.py` and run `make style`.

PR #312, opened 2023-07-28
Upgrade to Transformers v4.31

regisss:
# What does this PR do?

Following #306 since it was accidentally merged.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

PR #326, opened 2023-08-10
Enable Falcon-7b

schoi-habana:
This PR enables Habana optimized Falcon-7b. This version of the model uses 
1) static input shape along with token_idx and position_ids to avoid repeated graph compilation
2) fused RoPE kernel

This change requires transformers=4.32.0.dev0

## Before submitting


optimum/habana/transformers/generation/utils.py
```diff
@@ -183,7 +183,7 @@ def _validate_model_kwargs(self, model_kwargs: Dict[str, Any]):
         if self.config.is_encoder_decoder:
             for key in ["decoder_input_ids"]:
                 model_kwargs.pop(key, None)
-        if self.config.model_type == "llama":
+        if self.config.model_type == "llama" or self.config.model_type == "falcon":
```
regisss, line 186: Actually I removed the whole `_validate_model_kwargs` method as the issue with Llama has been fixed in Transformers v4.31 (I merged #312).
I tried Falcon without this and it worked on my side. Could you let me know if you get the same?

regisss (commented):
@schoi-habana Very clean PR! I left a couple of comments.

Also, I merged #312 to upgrade to Transformers v4.31 and #331 to be able to use Falcon in the text-generation example. Could you rebase this branch and let me know if that works well on your side please?

optimum/habana/transformers/generation/utils.py
schoi-habana: Thanks for upgrading Transformers version, it worked on my side as well. Marking it resolved

regisss:
There is an issue with the CI, it shouldn't fail. At least not with this error, I'm going to take a look.

schoi-habana:
Thanks @regisss. I thought the error `ModuleNotFoundError: No module named 'transformers.models.falcon'` is from using the old Transformers version. However, CI still fails with your patch to upgrade Transformers. If you find any issue with my change, please let me know.

regisss:
> Thanks @regisss. I thought the error `ModuleNotFoundError: No module named 'transformers.models.falcon'` is from using the old Transformers version. However, CI still fails with your patch to upgrade Transformers. If you find any issue with my change, please let me know.

Your branch is fine and `main` too. The issue is with the way the CI GitHub workflow is configured, it didn't trigger the tests at the right commit: so basically it skipped my commit of yesterday to upgrade Transformers...
What's weird is that it didn't happen with other PRs :thinking: I'll take a closer look at it!

regisss:
@schoi-habana Have you also tested Falcon-40B?

schoi-habana:
> @schoi-habana Have you also tested Falcon-40B?

No, this patch was targeted only for Falcon-7B. I was told you are going to work on Falcon-40B

regisss:
Well the modeling code is the same in Transformers, so not sure there is anything else to do. I'll test it :+1:

regisss:
@schoi-habana I just piushed a commit to add Falcon in this table: https://github.com/huggingface/optimum-habana#validated-models

You may also see some other commits to see if I can fix this CI issue.

regisss (approved):
Now CI passed. It's still weird that it failed yesterday :thinking: 

Anyway, I'm going to merge this one!

PR #328, opened 2023-08-11
Modify flan-t5 deepspeed configuration

yeonsily:
# What does this PR do?

Fixes # (issue)

## Before submitting


libinta (approved):

regisss:
@yeonsily @libinta Evaluation is quite slow (that was probably the case before this PR), I guess it's due to ZeRO-3 being use at inference. Do you observe the same? For example, I got 7 evaluation steps performed in 4min46s, which is approximately the same time needed to run 20 training steps.

PR #330, opened 2023-08-11
Enable hpu graph

htang2012:
# Enable hpu graph 

Fixes # (HPU graph is not enabled)

## Before submitting

### DIFF
--- a/text-generation-inference/Dockerfile
+++ b/text-generation-inference/Dockerfile
@@ -36,7 +36,7 @@ COPY --from=tgi /tmp/text-generation-inference/launcher launcher
 RUN cargo build --release
 
 # Text Generation Inference base image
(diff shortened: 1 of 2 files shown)


regisss (commented):
@htang2012 Good catch! Let's rather use `self.htcore` [here](https://github.com/huggingface/optimum-habana/blob/d3f6c069dcca41efc51d6105ca3aab545c80991c/text-generation-inference/server/text_generation_server/utils/logits_process.py#L51) and [there](https://github.com/huggingface/optimum-habana/blob/d3f6c069dcca41efc51d6105ca3aab545c80991c/text-generation-inference/server/text_generation_server/utils/logits_process.py#L53) please. Importing `htcore` too early led to some issues in other parts of the code if I remember correctly.

Regarding the switch to SynapseAI v1.11, I'll first release a new version of Optimum Habana and then we can do it.

regisss (approved):
LGTM

PR #342, opened 2023-08-16
Fix inference with Llama-2-70B

regisss:
# What does this PR do?

Text generation with Llama2-70B currently fails because of a dimension issue in key/value layers when doing tensor parallelism.

You can test it on Gaudi2 with:
```bash
python ../gaudi_spawn.py --use_deepspeed --world_size 2 run_generation.py --model_name_or_path meta-llama/Llama-2-70b-hf --batch_size 1 --use_hpu_graphs --use_kv_cache --max_new_tokens 100
```

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/llama/modeling_llama.py
+++ b/optimum/habana/transformers/models/llama/modeling_llama.py
@@ -1,4 +1,5 @@
 import math
+import os
 from typing import List, Optional, Tuple, Union
 
 import torch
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/llama/modeling_llama.py
```diff
@@ -58,9 +59,13 @@ def gaudi_llama_attention_forward(
         key_states = self.k_proj(hidden_states)
         value_states = self.v_proj(hidden_states)
 
+    tp_world_size = 1 if self.training else int(os.environ.get("WORLD_SIZE", 1))
+
     query_states = query_states.view(bsz, q_len, self.num_heads, self.head_dim).transpose(1, 2)
-    key_states = key_states.view(bsz, q_len, self.num_key_value_heads, self.head_dim).transpose(1, 2)
-    value_states = value_states.view(bsz, q_len, self.num_key_value_heads, self.head_dim).transpose(1, 2)
+    key_states = key_states.view(bsz, q_len, self.num_key_value_heads // tp_world_size, self.head_dim).transpose(1, 2)
+    value_states = value_states.view(bsz, q_len, self.num_key_value_heads // tp_world_size, self.head_dim).transpose(
```
ZhaiFeiyue, line 66: how about leave dim of `self.num_key_value_heads` to `-1`  because other three are not `-1`?

optimum/habana/transformers/models/llama/modeling_llama.py
```diff
@@ -58,9 +59,13 @@ def gaudi_llama_attention_forward(
         key_states = self.k_proj(hidden_states)
         value_states = self.v_proj(hidden_states)
 
+    tp_world_size = 1 if self.training else int(os.environ.get("WORLD_SIZE", 1))
```
ZhaiFeiyue, line 62: could we add a TODO here, since MS DS has fixed [here](https://github.com/microsoft/DeepSpeed/blob/341cefd2a406f578ee05b3a9d0686553926a7e61/deepspeed/module_inject/auto_tp.py#L383)), and I have checked the code of Habana internal DS, same fixed is applied([here](https://github.com/HabanaAI/DeepSpeed/blob/main/deepspeed/module_inject/replace_module.py#L440)) but seems is not  released in 1.11.

optimum/habana/transformers/models/llama/modeling_llama.py
regisss: You're absolutely right! Much easier :grin:

ZhaiFeiyue (approved):

PR #347, opened 2023-08-17
add warmup for evalution

ZhaiFeiyue:
# What does this PR do?

when measure the performance, we use `throughput_warmup_steps` to skip compilation steps for training, but for inference(evaluation and prediction) the compilation steps (first 3 steps and the last one batch if the dataset is not divided by BS) are not skipped . Absolutely we could follow [the same way](https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/trainer.py#L775) in training side, but sometime the dataset for evaluation is quite small and BS for evaluation is large, means too few steps for evaluation, in this case we cannot skip too much steps(4 include first 3 steps and the last one if with `--dataloader_drop_last True`), so this PR is a proposal to address the above issue by pre-running warmup_steps (reuse `throughput_warmup_steps`) to do the compilation first then do real evaluation or prediction like [text_generation](https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/run_generation.py#L390). 

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer_seq2seq.py
+++ b/optimum/habana/transformers/trainer_seq2seq.py
@@ -18,7 +18,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 import torch
-from torch.utils.data import Dataset
+from torch.utils.data import DataLoader, Dataset
 from transformers.deepspeed import is_deepspeed_zero3_enabled
 
 from optimum.utils import logging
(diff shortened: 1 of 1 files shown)


ZhaiFeiyue:
@regisss with is PR T5 large performance is 
before
image
after  set throughput_warmup_steps to 3
image

the eval_sample_per_second is improved from 2.7 to 4.0.

ZhaiFeiyue:
closed, since we decided to enable static shape for T5

PR #356, opened 2023-08-23
Fix graph compilation error from Falcon when batch size>1

schoi-habana:
This change fixes the size mismatch between position_ids and cos when batch size > 1.

## Before submitting


schoi-habana:
@regisss would you be able to check why the CI test keeps failing? The error message says some checkpoint is missing but that seems irrelevant to my change.

PR #360, opened 2023-08-24
Fix beam search output

puneeshkhanna:
Also update cur_len with token_idx.item() and use that to slice the input_ids when calling beam_scorer.process().
The logic will work when token idx is None too.

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -1678,6 +1678,10 @@ class GaudiGenerationMixin(GenerationMixin):
         num_beams = beam_scorer.num_beams
 
         batch_beam_size, cur_len = input_ids.shape
+        token_idx = model_kwargs.get("token_idx", None)
+        if token_idx is not None:
+            # Update cur_len in case of static shapes
+            cur_len = token_idx.item()
 
         if num_beams * batch_size != batch_beam_size:
             raise ValueError(
@@ -1778,7 +1782,7 @@ class GaudiGenerationMixin(GenerationMixin):
 
             # stateless
             beam_outputs = beam_scorer.process(
-                input_ids,
+                input_ids[:, :cur_len],
                 next_token_scores,
                 next_tokens,
                 next_indices,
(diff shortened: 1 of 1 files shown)


puneeshkhanna:
@regisss  - please review this PR. Also "make style" passed so CI should pass now.

regisss:
Thanks @puneeshkhanna! I'll review it shortly

puneeshkhanna:
All checks passed.

optimum/habana/transformers/generation/utils.py
```diff
@@ -2371,7 +2380,6 @@ def constrained_beam_search(
                 cur_len = cur_len + 1
                 continue  # don't waste resources running the code we don't need
 
-            token_idx = model_kwargs.get("token_idx", None)
```
regisss, line 2374: Why removing this line? We need to update `token_idx` at each iteration if I'm not mistaken.

regisss (commented):
@puneeshkhanna Good catch!
I just have one comment regarding removing `token_idx = model_kwargs.get("token_idx", None)` in `constrained_beam_search`.

optimum/habana/transformers/generation/utils.py
puneeshkhanna: We get it once out side the loop and set cur_len to it. 
cur_len is incremented inside the loop and use cur_len to slice the tensor.
Note that the logic will now work even if we don't set static shapes as earlier code was forcefully using token_idx.item() inside the loop..

<<
 token_idx = model_kwargs.get("token_idx", None)
 if token_idx is not None:
     # Update cur_len in case of static shapes
     cur_len = token_idx.item()
>>

puneeshkhanna: Logic is now same in both constrained_beam_search() as well as beam_search()

puneeshkhanna: token_idx is still updated though by calling update_model_kwargs(). It is just we don't need to read it here since we can use cur_len.

regisss: Ah yes you're fully right, I missed that `token_idx` is updated anyway. Thanks for the explanation!

regisss (approved):
LGTM!

PR #359, opened 2023-08-24
Load llama ckpt to meta to work around OOM issue on CPU

mandy-li:
When generate text with DeepSpeed library, this PR loads llama ckpt to meta to work around OOM issue on CPU.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #361, opened 2023-08-24
enable checkpoint loading for llama, which will benefit llama2 70b an…

sywangyi:
…d greatly save host memory

### DIFF
--- a/examples/text-generation/checkpoint_utils.py
+++ b/examples/text-generation/checkpoint_utils.py
@@ -3,8 +3,8 @@ import os
 from pathlib import Path
 
 import torch
-from huggingface_hub import list_repo_files, snapshot_download
-from transformers.utils import is_offline_mode, is_safetensors_available
+from huggingface_hub import snapshot_download
+from transformers.utils import is_offline_mode
 
 
 def get_repo_root(model_name_or_path, local_rank=-1):
@@ -22,11 +22,6 @@ def get_repo_root(model_name_or_path, local_rank=-1):
 
         # Only download PyTorch weights by default
         allow_patterns = ["*.bin"]
-        # If the model repo contains any .safetensors file and
-        # safetensors is installed, only download safetensors weights
-        if is_safetensors_available():
-            if any(".safetensors" in filename for filename in list_repo_files(model_name_or_path)):
-                allow_patterns = ["*.safetensors"]
 
         # Download only on first process
         if local_rank in [-1, 0]:


sywangyi:
should work with https://github.com/microsoft/DeepSpeed/pull/4206 for llama2 70b

regisss:
Isn't it a duplicate of #359 ?

sywangyi:
yes. duplicated, I have not noticed that, I will close this PR.

sywangyi:
but https://github.com/microsoft/DeepSpeed/pull/4206 should be ported to habana deepspeed, or else llama 70b output is incorrect after enabling the shard load

sywangyi:
also is_safetensors_available() should be removed which stop the download of *.bin, since allow_patterns is set to ["*.safetensors"]

regisss:
> but [microsoft/DeepSpeed#4206](https://github.com/microsoft/DeepSpeed/pull/4206) should be ported to habana deepspeed, or else llama 70b output is incorrect after enabling the shard load

I think @mandy-li is looking into it

regisss:
> also is_safetensors_available() should be removed which stop the download of _.bin, since allow_patterns is set to ["_.safetensors"]

Can you provide a command to reproduce it?
If `safetensors` is not installed, it should download the `.bin` checkpoint. Or do you want to download the `.bin` checkpoint even though `safetensors` is installed?

regisss:
> [https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/checkpoint_utils.py#L63，](https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/checkpoint_utils.py#L63%EF%BC%8C) this line only filter the *.bin, and write it to checkpoint.json

Good catch! Can you keep just this change in this PR so that we merge it while @mandy-li can keep investigating Llama meta loading with DeepSpeed?

sywangyi:
> > [https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/checkpoint_utils.py#L63，](https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/checkpoint_utils.py#L63%EF%BC%8C) this line only filter the *.bin, and write it to checkpoint.json
> 
> Good catch! Can you keep just this change in this PR so that we merge it while @mandy-li can keep investigating Llama meta loading with DeepSpeed?

done

regisss (approved):
LGTM!

PR #379, opened 2023-09-01
Fix transformer version to be installed

vivekgoe:
# What does this PR do?
Fix transformer version to be installed (should be updated to a newer version only after ensuring new version does not cause regressions). This is to prevent unexplained failures for users when a new version of transformer is released.

Fixes # (issue)

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ except Exception as error:
 
 
 INSTALL_REQUIRES = [
-    "transformers >= 4.32.0",
+    "transformers == 4.32.0",
     "optimum",
     "torch",
     "accelerate >= 0.22.0",


regisss:
Unfortunately, we would also need to do it in [Optimum](https://github.com/huggingface/optimum/blob/7450ca30e295abc9e20d56d0aa741402322def0f/setup.py#L69) (the parent library) as we advertise to install Optimum Habana with `pip install optimum[habana]`. Which means that I need to release a patch for Optimum every time we upgrade Transformers.
While this is feasible, the nightly CI currently relies on Transformers' main (same for Accelerate and Diffusers) so we should be able to catch errors before release and have patches ready. It didn't happen lately because a big change in Transformers occurred but it should be the case now. So I would be in favor of keeping it that way for now and, if that doesn't prove to be a good solution, switching to a stricter strategy where versions are pinned here and in Optimum.

vivekgoe:
@regisss ok lets drop this PR for now.

PR #386, opened 2023-09-06
accelerate the inference when input prompt length changes in TGI

sywangyi:
also if there's pad left in input token, token_idx is wrong calculated.

fix issue that's found in multi-request enabling


regisss (approved):
LGTM, thanks for catching this!

PR #389, opened 2023-09-08
support static shape in concatenate and filter in TGI

sywangyi:
# What does this PR do?
add static shape support in concatenate and filter

## Before submitting

### DIFF
--- a/text-generation-inference/server/text_generation_server/models/causal_lm.py
+++ b/text-generation-inference/server/text_generation_server/models/causal_lm.py
@@ -153,7 +153,7 @@ class CausalLMBatch(Batch):
         )
 
     @tracer.start_as_current_span("filter")
-    def filter(self, request_ids: List[int]) -> Optional["CausalLMBatch"]:
+    def filter(self, request_ids: List[int], is_optimized_for_gaudi: bool = False) -> Optional["CausalLMBatch"]:
         if len(request_ids) == 0:
             raise ValueError("Batch must have at least one request")
         if len(request_ids) == len(self):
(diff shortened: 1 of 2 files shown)


text-generation-inference/server/text_generation_server/models/causal_lm.py
```diff
@@ -202,37 +202,51 @@ def filter(self, request_ids: List[int]) -> Optional["CausalLMBatch"]:
...
+            self.attention_mask = self.attention_mask[
+                keep_indices,
+                -(self.padding_right_offset + max_input_length) : (
+                    self.attention_mask.shape[1] - self.padding_right_offset
+                )
+                + new_padding_right_offset,
+            ]
 
         # Ensure that past_key_values tensors can be updated in-place
+        kv_tuple = False
```
regisss, line 217: Why is this necessary now? Are layers automatically converted into lists at some point?

text-generation-inference/server/text_generation_server/models/causal_lm.py
```diff
@@ -348,10 +362,12 @@ def concatenate(cls, batches: List["CausalLMBatch"]) -> "CausalLMBatch":
             # BLOOM Keys:   [batch_size * num_heads, head_dim, seq_length]
             # BLOOM Values: [batch_size * num_heads, seq_length, head_dim]
             # And ensure that we can update tensors in-place
+            kv_tuple = False
```
regisss, line 365: Same

sywangyi: in https://github.com/huggingface/optimum-habana/pull/389/files#diff-244871853752a7a019853c852803a797a9a14dddb3d9d859ecbae4b316ed6974R457

text-generation-inference/server/text_generation_server/models/causal_lm.py
sywangyi: in https://github.com/huggingface/optimum-habana/pull/389/files#diff-244871853752a7a019853c852803a797a9a14dddb3d9d859ecbae4b316ed6974R247. the past_key_value type should be exactly same with the case in single batch. or else hpu graph compile will meet type check error.

regisss (approved):
LGTM

PR #390, opened 2023-09-08
add warmup arg and move stats printing to the end

polisettyvarma:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -66,6 +66,7 @@ def main():
     )
     parser.add_argument("--max_new_tokens", type=int, default=100, help="Number of tokens to generate.")
     parser.add_argument("--batch_size", type=int, default=1, help="Input batch size.")
+    parser.add_argument("--warmup", type=int, default=5, help="Number of warmup iterations for benchmarking.")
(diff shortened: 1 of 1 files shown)


polisettyvarma:
please review this @regisss @vivekgoe @libinta

regisss (commented):
@polisettyvarma Why moving memory stats and the graph compilation duration to the end far from the throughput? They should all be in the same place IMO.

polisettyvarma:
> @polisettyvarma Why moving memory stats and the graph compilation duration to the end far from the throughput? They should all be in the same place IMO.

@regisss reason for keeping throughput there is for separator variable used in multiple places for alignment mostly. we created/updated the string but not printing throughput there.

regisss:
> > @polisettyvarma Why moving memory stats and the graph compilation duration to the end far from the throughput? They should all be in the same place IMO.
> 
> @regisss reason for keeping throughput there is for separator variable used in multiple places for alignment mostly. we created/updated the string but not printing throughput there.

Okay, but I maintain that throughput, graph compilation duration and memory stats should all be printed together and before the outputs. Otherwise it may be very misleading and some users will just not see the throughput.

polisettyvarma:
please review now @regisss

regisss (approved):
LGTM

PR #392, opened 2023-09-11
pad to max_total_tokens in case max_new_token changes per request and…

sywangyi:
… new hpu graph will be created

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/text-generation-inference/server/text_generation_server/models/causal_lm.py
+++ b/text-generation-inference/server/text_generation_server/models/causal_lm.py
@@ -91,6 +91,15 @@ class CausalLMBatch(Batch):
         max_truncation = 0
         padding_right_offset = 0
         max_decode_tokens = 0
+
+        # TODO: this should be set to rust side `max_total_tokens`,
+        # (see https://github.com/huggingface/text-generation-inference/blob/main/launcher/src/main.rs#L177)
+        # but TGI does not offer an API to expose this variable to python, as this variable
+        # is handled by the client but it appears the model is initialized by the server.
+        # An alternative could be to initialize the buffers during warmup.
+        # Dummy
+        max_total_tokens = 2048
+
         for i, r in enumerate(pb.requests):
             requests_idx_mapping[r.id] = i
             inputs.append(r.inputs)
@@ -118,6 +127,9 @@ class CausalLMBatch(Batch):
 
         max_input_length = max(input_lengths)
         max_tokens = len(inputs) * max_input_length + max_decode_tokens
+        if is_optimized_for_gaudi:
+            # pad to max_total_tokens in case max_new_token changes per request and triggers new hpu graph generation
+            padding_right_offset = max_total_tokens - max_input_length
 
         input_ids = tokenized_inputs["input_ids"]
         attention_mask = tokenized_inputs["attention_mask"]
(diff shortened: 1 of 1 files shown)


regisss (commented):
So the idea is to compile the model once for size 2048 to avoid extra compilations when different lengths are requested right?

sywangyi:
> So the idea is to compile the model once for size 2048 to avoid extra compilations when different lengths are requested right?

yes.

regisss:
Okay, no extra compilation will be needed for all sequences > 2048 tokens? Or just for some multiples of 8 (or another number) that are smaller than 2048?
Do we have an idea of how much slower latency is with this change for 7B models for instance?

Just trying to understand if this is good enough to keep this for a while or if it is just meant to be a short-term fix.

sywangyi:
@regisss , I see there's  `--max-total-tokens` in the text-generation-launcher command line. whose description is" This is the most important value to set as it defines the "memory budget" of running clients requests. Clients will send input sequences and ask to generate `max_new_tokens` on top. with a value of `1512` users can send either a prompt of `1000` and ask for `512` new tokens, or send a prompt of `1` and ask for `1511` max_new_tokens. The larger this value, the larger amount each request will be in your RAM and the less effective batching can be [env: MAX_TOTAL_TOKENS=] [default: 2048]", currently this parameter is not passed to python, as what I describe in the code. I would like to utilize this parameter and set the padding length. so that if we set max-total-tokens in launcher to 2048. the sequence > 2048 will not passed to server python code.
for 7B llama. I test the perf. when max-new-tokens set to 64, if pad out length to 1024 (2048-1024(input pad to 1024)), the time_per_token is 44ms, if not pad to 1024, just pad to new token number(64), the time_per_token is 33ms. I test it in AWS DL1 instance.

sywangyi:
The change of three dims could lead to the hpu graph recompile. one is input length, one is output length, one is batch number. the https://github.com/huggingface/optimum-habana/pull/386 is for input length. this PR is for output length. the most complicated is for batch dim. Should we get the max batch in warmup and use the max batch for each inference？ generating too much graph may lead to memory problem.

regisss:
> @regisss , I see there's `--max-total-tokens` in the text-generation-launcher command line. whose description is" This is the most important value to set as it defines the "memory budget" of running clients requests. Clients will send input sequences and ask to generate `max_new_tokens` on top. with a value of `1512` users can send either a prompt of `1000` and ask for `512` new tokens, or send a prompt of `1` and ask for `1511` max_new_tokens. The larger this value, the larger amount each request will be in your RAM and the less effective batching can be [env: MAX_TOTAL_TOKENS=] [default: 2048]", currently this parameter is not passed to python, as what I describe in the code. I would like to utilize this parameter and set the padding length. so that if we set max-total-tokens in launcher to 2048. the sequence > 2048 will not be passed to server python code. for 7B llama. I test the perf. when max-new-tokens set to 64, if pad out length to 1024 (2048-1024(input pad to 1024)), the time_per_token is 44ms, if not pad to 1024, just pad to new token number(64), the time_per_token is 33ms. I test it in AWS DL1 instance.

Okay, could you link to this comment about `--max-total-tokens` in the comment you added at line 95 please?

regisss:
> The change of three dims could lead to the hpu graph recompile. one is input length, one is output length, one is batch number. the #386 is for input length. this PR is for output length. the most complicated is for batch dim. Should we get the max batch in warmup and use the max batch for each inference？some batch contains dummy data that's not a valid request. however generating too much graph may lead to memory problem. Or shall we generate graph for each batch num (< max_batch) in warmup. Any idea?

IMO it depends whether we have more or less constant latency with max_batch (which we probably don't).
Generating several graphs for various batch sizes during warmup could make sense, but this may require a lot of memory no? I don't know the memory taken by a static graph for such a model.

regisss (approved):
LGTM

PR #396, opened 2023-09-12
Add support for attention softmax in BF16 such as for llama.

puneeshkhanna:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/configuration_utils.py
+++ b/optimum/habana/transformers/generation/configuration_utils.py
@@ -14,6 +14,9 @@ class GaudiGenerationConfig(GenerationConfig):
     ignore_eos (`bool`, *optional*):
         Whether to ignore finished sequences (faster in lazy mode and with HPU graphs) or not (eager mode).
         If not specified, it will automatically be set to `True` if lazy mode is on.
+    attn_softmax_bf16 (`bool`, *optional*):
+        Whether to run attention softmax layer in lower precision provided that the model supports it and
+        is also running in lower precision.
     """
 
     def __init__(self, **kwargs):
--- a/optimum/habana/transformers/generation/configuration_utils.py
+++ b/optimum/habana/transformers/generation/configuration_utils.py
@@ -21,3 +21,4 @@ class GaudiGenerationConfig(GenerationConfig):
 
         self.static_shapes = kwargs.get("static_shapes", None)
         self.ignore_eos = kwargs.get("ignore_eos", None)
+        self.attn_softmax_bf16 = kwargs.get("attn_softmax_bf16", None)
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -447,9 +447,7 @@ class GaudiGenerationMixin(GenerationMixin):
         self._validate_generated_length(generation_config, input_ids_length, has_default_max_length)
 
         # determine whether attention softmax needs to execute in lower precision
-        model_kwargs["attn_softmax_bf16"] = (
-            generation_config.attn_softmax_bf16 if hasattr(generation_config, "attn_softmax_bf16") else None
-        )
+        model_kwargs["attn_softmax_bf16"] = generation_config.attn_softmax_bf16
 
         # 7. determine generation mode
         generation_mode = self._get_generation_mode(generation_config, assistant_model)
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -446,6 +446,11 @@ class GaudiGenerationMixin(GenerationMixin):
             generation_config.max_length = generation_config.max_new_tokens + input_ids_length
         self._validate_generated_length(generation_config, input_ids_length, has_default_max_length)
 
+        # determine whether attention softmax needs to execute in lower precision
+        model_kwargs["attn_softmax_bf16"] = (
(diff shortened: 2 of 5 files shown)


dvarshney-habana (approved):

polisettyvarma (approved):

optimum/habana/transformers/models/llama/modeling_llama.py
```diff
@@ -209,11 +218,13 @@ def gaudi_llama_model_forward(
     output_hidden_states: Optional[bool] = None,
     return_dict: Optional[bool] = None,
     token_idx: Optional[torch.Tensor] = None,
+    attn_softmax_bf16: Optional[bool] = None,
```
regisss, line 221:
```suggestion
    attn_softmax_bf16: Optional[bool] = False,
```

optimum/habana/transformers/models/llama/modeling_llama.py
```diff
@@ -352,6 +365,7 @@ def forward(
         output_hidden_states: Optional[bool] = None,
         return_dict: Optional[bool] = None,
         token_idx: Optional[torch.Tensor] = None,
+        attn_softmax_bf16: Optional[bool] = None,
```
regisss, line 368:
```suggestion
        attn_softmax_bf16: Optional[bool] = False,
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -446,6 +446,11 @@ def generate(
             generation_config.max_length = generation_config.max_new_tokens + input_ids_length
         self._validate_generated_length(generation_config, input_ids_length, has_default_max_length)
 
+        # determine whether attention softmax needs to execute in lower precision
+        model_kwargs["attn_softmax_bf16"] = (
+            generation_config.attn_softmax_bf16 if hasattr(generation_config, "attn_softmax_bf16") else None
+        )
```
regisss, line 452: I think we should define a `attn_softmax_bf16` field in: https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/generation/configuration_utils.py
That way, we won't have to check here if it is in the generation config, and that will allow users to easily enable it in other contexts (e.g. model evaluation during training as in https://github.com/huggingface/optimum-habana/blob/4bf8a224b8838850e67ac85ca7fdbca91cc09d2d/optimum/habana/transformers/trainer_seq2seq.py#L44

regisss (commented):
@puneeshkhanna Nice PR!
I left a couple of comments, mainly to add this new arg to `GaudiGenerationConfig`.

optimum/habana/transformers/generation/utils.py
puneeshkhanna: Got it. Will push patch soon.

puneeshkhanna: Done

optimum/habana/transformers/models/llama/modeling_llama.py
puneeshkhanna: Done

optimum/habana/transformers/models/llama/modeling_llama.py
puneeshkhanna: Done

puneeshkhanna:
@regisss - I hope I have addressed all the nice review comments. Please check now.

optimum/habana/transformers/generation/configuration_utils.py
```diff
@@ -21,3 +21,4 @@ def __init__(self, **kwargs):
 
         self.static_shapes = kwargs.get("static_shapes", None)
         self.ignore_eos = kwargs.get("ignore_eos", None)
+        self.attn_softmax_bf16 = kwargs.get("attn_softmax_bf16", None)
```
regisss, line 24: Could you also add it in the docstring of the class?  You can reuse the help text you wrote in `run_generation.py`.

regisss (commented):
Thanks for updating the PR @puneeshkhanna ! Let's also add `attn_softmax_bf16` in the docstring of the `GaudiGenerationConfig` class and then it will be good to merge.

optimum/habana/transformers/generation/configuration_utils.py
puneeshkhanna: So silly of me that I missed it. Done now.

regisss (approved):

mandy-li:
@puneeshkhanna I am wondering why we use "--attn_softmax_bf16" arg to control if to run attn_softmax in bf16 or not? why not run attn_softmax_bf16 directly if the model is in bf16/hmp? why we let user control from cmd arg? Thanks

puneeshkhanna:
> @puneeshkhanna I am wondering why we use "--attn_softmax_bf16" arg to control if to run attn_softmax in bf16 or not? why not run attn_softmax_bf16 directly if the model is in bf16/hmp? why we let user control from cmd arg? Thanks

@mandy-li  - Model code by default has explicit upcast to float32 (torch.nn.functional.softmax(attn_weights, dim=-1, dtype=torch.float32)) so even though model is run in bf16, softmax will always run in FP32. Usually that is the case with softmax in attention block in most of the models because of exponent operations. 
Hence we need a user control for it and if desired accuracy and perf can be achieved via softmax in BF16 only then this option can be enabled.

PR #397, opened 2023-09-12
skip hpugraph usage for first token to save memory

polisettyvarma:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -77,7 +77,7 @@ Here are a few settings you may be interested in:
 - `--batch_size` to specify the batch size
 - `--bf16` to run generation in bfloat16 precision (or to be specified in your DeepSpeed configuration if using DeepSpeed)
 - `--use_hpu_graphs` to use [HPU graphs](https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_HPU_Graphs.html) to speed up generation
-- `--limit_hpu_graphs` to skip HPUGraph usage for first token to save memory
+- `--limit_hpu_graphs` to skip HPU Graph usage for first token to save memory
 - `--use_kv_cache` to use the [key/value cache](https://huggingface.co/docs/transformers/main/en/main_classes/text_generation#transformers.GenerationConfig.use_cache) to speed up generation
 - `--do_sample` or `--num_beams` to generate new tokens doing sampling or beam search (greedy search is the default)
 - `--prompt` to benchmark the model on a prompt of your choice
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -77,6 +77,7 @@ Here are a few settings you may be interested in:
 - `--batch_size` to specify the batch size
 - `--bf16` to run generation in bfloat16 precision (or to be specified in your DeepSpeed configuration if using DeepSpeed)
 - `--use_hpu_graphs` to use [HPU graphs](https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_HPU_Graphs.html) to speed up generation
+- `--limit_hpu_graphs` to skip HPUGraph usage for first token to save memory
 - `--use_kv_cache` to use the [key/value cache](https://huggingface.co/docs/transformers/main/en/main_classes/text_generation#transformers.GenerationConfig.use_cache) to speed up generation
 - `--do_sample` or `--num_beams` to generate new tokens doing sampling or beam search (greedy search is the default)
 - `--prompt` to benchmark the model on a prompt of your choice
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -172,6 +172,11 @@ def main():
         type=str,
         help="Output directory to store results in.",
     )
+    parser.add_argument(
+        "--limit_hpu_graphs",
+        action="store_true",
+        help="Skip HPUGraph usage for first token to save memory",
+    )
 
     args = parser.parse_args()
 
@@ -185,6 +190,8 @@ def main():
         os.environ.setdefault("PT_HPU_LAZY_ACC_PAR_MODE", "0")
         os.environ.setdefault("PT_HPU_ENABLE_LAZY_COLLECTIVES", "true")
 
+    if not args.use_hpu_graphs:
+        args.limit_hpu_graphs = False
     # Device is HPU
     args.device = "hpu"
     import habana_frameworks.torch.hpu as torch_hpu
@@ -360,6 +367,7 @@ def main():
     generation_config.force_words_ids = force_words_ids
     generation_config.num_return_sequences = args.num_return_sequences
     generation_config.attn_softmax_bf16 = args.attn_softmax_bf16
+    generation_config.limit_hpu_graphs = args.limit_hpu_graphs
 
     if args.dataset_name is None:
         # Benchmark over the prompts below
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -175,7 +175,7 @@ def main():
     parser.add_argument(
         "--limit_hpu_graphs",
         action="store_true",
-        help="Skip HPUGraph usage for first token to save memory",
+        help="Skip HPU Graph usage for first token to save memory",
     )
 
     args = parser.parse_args()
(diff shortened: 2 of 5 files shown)


puneeshkhanna (approved):

polisettyvarma:
please review this @dvarshney-habana

examples/text-generation/README.md
```diff
@@ -77,6 +77,7 @@ Here are a few settings you may be interested in:
 - `--batch_size` to specify the batch size
 - `--bf16` to run generation in bfloat16 precision (or to be specified in your DeepSpeed configuration if using DeepSpeed)
 - `--use_hpu_graphs` to use [HPU graphs](https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_HPU_Graphs.html) to speed up generation
+- `--limit_hpu_graphs` to skip HPUGraph usage for first token to save memory
```
regisss, line 80:
```suggestion
- `--limit_hpu_graphs` to skip HPU Graph usage for first token to save memory
```

examples/text-generation/run_generation.py
```diff
@@ -166,6 +166,11 @@ def main():
         type=str,
         help="Output directory to store results in.",
     )
+    parser.add_argument(
+        "--limit_hpu_graphs",
+        action="store_true",
+        help="Skip HPUGraph usage for first token to save memory",
```
regisss, line 172:
```suggestion
        help="Skip HPU Graph usage for first token to save memory",
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -432,6 +434,11 @@ def generate(
         if streamer is not None:
             streamer.put(input_ids.cpu())
 
+        # determine whether limit_hpu_graphs needs to be used
+        model_kwargs["limit_hpu_graphs"] = (
+            generation_config.limit_hpu_graphs if hasattr(generation_config, "limit_hpu_graphs") else False
+        )
```
regisss, line 440: Let's define `limit_hpu_graphs` in: https://github.com/huggingface/optimum-habana/blob/4bf8a224b8838850e67ac85ca7fdbca91cc09d2d/optimum/habana/transformers/generation/configuration_utils.py#L19
so that we don't have to check here if it's in the generation config.

optimum/habana/transformers/generation/utils.py
```diff
@@ -1096,12 +1103,20 @@ def greedy_search(
             # prepare model inputs
             model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
 
+            hpu_graphs_kwargs = {}
+            if model_kwargs["limit_hpu_graphs"]:
+                hpu_graphs_kwargs.update({"bypass_hpu_graphs": False})
+                if "first_token" not in model_kwargs.keys():
+                    model_kwargs["first_token"] = True
+                    hpu_graphs_kwargs.update({"bypass_hpu_graphs": True})
```
regisss, line 1111: Maybe we could define a `get_hpu_graphs_kwargs` method in this class to factorize this block which is called in every decoding method?
So that we have:
```py
hpu_graphs_kwargs = get_hpu_graphs_kwargs(model_kwargs)
```
It will be easier to maintain.

optimum/habana/transformers/generation/utils.py
```diff
@@ -1408,12 +1423,20 @@ def sample(
...
```
regisss, line 1431:
```suggestion
            hpu_graphs_kwargs = get_hpu_graphs_kwargs(model_kwargs)
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -1726,11 +1749,19 @@ def beam_search(
 
             model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
 
+            hpu_graphs_kwargs = {}
+            if model_kwargs["limit_hpu_graphs"]:
+                hpu_graphs_kwargs.update({"bypass_hpu_graphs": False})
+                if "first_token" not in model_kwargs.keys():
+                    model_kwargs["first_token"] = True
+                    hpu_graphs_kwargs.update({"bypass_hpu_graphs": True})
```
regisss, line 1757:
```suggestion
            hpu_graphs_kwargs = get_hpu_graphs_kwargs(model_kwargs)
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -2370,11 +2401,19 @@ def constrained_beam_search(
...
```
regisss, line 2409:
```suggestion
            hpu_graphs_kwargs = get_hpu_graphs_kwargs(model_kwargs)
```

optimum/habana/transformers/generation/utils.py
polisettyvarma: image

i did actually but when i run style it changes the code which is wrong and i need to add on top of it something
i can unpack tuple then dict

optimum/habana/transformers/generation/utils.py
polisettyvarma: same as above

optimum/habana/transformers/generation/utils.py
polisettyvarma: same as above

optimum/habana/transformers/generation/utils.py
polisettyvarma: same as above

optimum/habana/transformers/generation/utils.py
regisss: Oh that's weird! What do you return in `_get_hpu_graphs_kwargs`?

polisettyvarma:
please review now @regisss, @dvarshney-habana and @puneeshkhanna

optimum/habana/transformers/generation/configuration_utils.py
puneeshkhanna, line 1: add doc string too similar to attn softmax bf16

puneeshkhanna: Rest all looks good to me

dvarshney-habana (approved):

regisss (commented):
CI failed because `_get_hpu_graphs_kwargs` should be linked to Transformers' `generate here: https://github.com/huggingface/optimum-habana/blob/af3da936c338907c446b4612e649eafee147c5a3/optimum/habana/transformers/modeling_utils.py#L103

Or, maybe easier, you can just declare this method outside of this class and use it that way without `self`.

polisettyvarma:
> CI failed because `_get_hpu_graphs_kwargs` should be linked to Transformers' `generate here:
> 
> https://github.com/huggingface/optimum-habana/blob/af3da936c338907c446b4612e649eafee147c5a3/optimum/habana/transformers/modeling_utils.py#L103
> 
> Or, maybe easier, you can just declare this method outside of this class and use it that way without `self`.

I took the first approach, please review @regisss

optimum/habana/transformers/generation/utils.py
```diff
@@ -131,6 +131,14 @@ def _expand_dict_for_generation(dict_to_expand):
 
         return input_ids, model_kwargs
 
+    def _get_hpu_graphs_kwargs(self, model_kwargs):
+        hpu_graphs_kwargs = {}
+        if model_kwargs["limit_hpu_graphs"]:
+            hpu_graphs_kwargs.update({"bypass_hpu_graphs": False})
+            if "first_token" not in model_kwargs.keys():
+                model_kwargs["first_token"] = True
+                hpu_graphs_kwargs.update({"bypass_hpu_graphs": True})
```
regisss, line 140: We also need to return `hpu_graphs_kwargs` at the end

regisss (commented):
@polisettyvarma CI failed, we need to return `hpu_graphs_kwargs` at the end of `_get_hpu_graphs_kwargs`.
There is also one merge conflict to solve, probably due to the trim logit PR that I just merged. There shouldn't be much to do I think.

polisettyvarma:
please review now @regisss

regisss (approved):

yafshar:
@regisss, @polisettyvarma which model is using this feature? I do not see `bypass_hpu_graphs` in any models' arguments. Would you please point me to a model which is using this?

regisss:
> @regisss, @polisettyvarma which model is using this feature? I do not see `bypass_hpu_graphs` in any models' arguments. Would you please point me to a model which is using this?

@yafshar You won't see it in the code of any model. It's managed by the HPU graph API which wraps the forward of the model.
For example, in a Docker container running the SynapseAI 1.15 Docker image, you can find it at the beginning of the `wrapped_hpugraph_forward` method in `/usr/local/lib/python3.10/dist-packages/habana_frameworks/torch/hpu/graphs.py`.

PR #408, opened 2023-09-17
Fix first token

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/text-generation-inference/server/text_generation_server/utils/logits_process.py
+++ b/text-generation-inference/server/text_generation_server/utils/logits_process.py
@@ -43,8 +43,7 @@ class StaticWarper:
         self.static_next_logprob = None
 
     def __call__(self, scores):
-        # use hpu graph only for > 1st token
-        if self.hpu_graph is None and scores.shape[-2] == 1:
+        if self.hpu_graph is None:
             self.static_scores = scores.clone().contiguous()
             self.static_warped_scores = scores.clone().contiguous()
             self.static_next_logprob = scores.clone().contiguous()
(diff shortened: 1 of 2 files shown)


text-generation-inference/server/text_generation_server/utils/logits_process.py
```diff
@@ -59,16 +58,14 @@ def __call__(self, scores):
...
-        else:
-            # 1st token disposal
-            for warper in self.warpers:
-                scores = warper(None, scores)
-            return scores, torch.log_softmax(scores, -1)
+        return self.static_warped_scores, self.static_next_logprob
+        # cpu branch
+        for warper in self.warpers:
+            scores = warper(None, scores)
+        return scores, torch.log_softmax(scores, -1)
```
regisss, line 68: This part of the code will never be reached as we return before, is it expected?

sywangyi: yes, the 1st token score also has shape of [1, xxxx], same with the other token. as in https://github.com/huggingface/optimum-habana/pull/408/files#diff-244871853752a7a019853c852803a797a9a14dddb3d9d859ecbae4b316ed6974R658

regisss: So why keeping these lines then?

sywangyi: should I remove it. I see https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/utils/logits_process.py#L66 . Is original code left here for debug?

regisss: I think so given this code will never be reached (the method always returns before reaching there)

regisss (approved):

PR #418, opened 2023-09-22
makes the record_shapes of the profiler changeable.

inkcherry:
The default argument of profiler ```record_shapes``` is ```True```. Enabling this may improve memory usage and may lead to some workload crashes with OOM when profiling is enabled. Setting ```record_shapes``` to ```False``` when it is not used could avoid these crashes and enable profiling.  
  
This pr makes the  ```record_shapes``` changeable.


regisss (approved):
LGTM!

The code style check failed, could you run the following from the root of the repo to make sure it passes please?
```
pip install --upgrade black ruff
make style
```

inkcherry:
@regisss  fixed, thanks for the review!

PR #426, opened 2023-09-26
Add a configurable max input tokens parameter

### LINKED ISSUE #423: Does it make sense to also provide an option of max input tokens for text generation ?
### Feature request

Right now we tokenize as below and based upon input sentences in the batch, the input ids are padded.
input_tokens = tokenizer.batch_encode_plus(input_sentences, return_tensors="pt", padding=True)

What if the next batch of data exceeds the above input tokens length which will trigger recompilations.

Would it not be better if we also introduce an argument of max input length and tokenize as below using padding to max length and truncation to True:

input_tokens = tokenizer.batch_encode_plus(batch_sentences, padding='max_length', truncation=True, max_length=args.max_input_length)

### Motivation

Avoid graph recompilations.

### Your contribution

Yes if we agree.

puneeshkhanna:
# What does this PR do?

Fixes #423 

## Before submitting

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -65,6 +65,12 @@ def main():
         help="Whether to perform generation in bf16 precision.",
     )
     parser.add_argument("--max_new_tokens", type=int, default=100, help="Number of tokens to generate.")
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


regisss (approved):
LGTM!

PR #430, opened 2023-09-27
No need to wrap DDP when using Fast DDP

ikurtchen:
# What does this PR do?

When I enable fast ddp with `--distribution_strategy fast_ddp`, the model is wrapped with `torch.nn.parallel.DistributedDataParallel` from `accelerator.prepare_model()`, this could be incorrect since fast ddp handles gradients all-reduce by itself. In this patch, it checks if it's not fast ddp before wrap with `torch.nn.parallel.DistributedDataParallel` in `accelerator`.

## Before submitting
  - no
  - no need update doc 
  - no need new tests


regisss (approved):
Good catch!

PR #440, opened 2023-10-03
Update GPTNEOX command 

ankurneog:
Remove world size argument from from GPTNEOX Multi node execution command , since the argument is redundant in case of Multinode execution, which relies on the hostfile for IP and device/card information.

@regisss  : please help with the review and merge

### DIFF
--- a/docs/source/package_reference/stable_diffusion_pipeline.mdx
+++ b/docs/source/package_reference/stable_diffusion_pipeline.mdx
@@ -38,3 +38,13 @@ To get the most out of it, it should be associated with a scheduler that is opti
 
 [[autodoc]] diffusers.schedulers.scheduling_ddim.GaudiDDIMScheduler
     - all
+
+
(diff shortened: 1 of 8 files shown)


regisss (approved):
LGTM, good catch!

regisss:
However, that's independent form this PR, but the doc build workflow fails because it cannot find the Habana Docker image:
```
ERROR: failed to solve: vault.habana.ai/gaudi-docker/1.11.0/ubuntu20.04/habanalabs/pytorch-installer-2.0.1:latest: vault.habana.ai/gaudi-docker/1.11.0/ubuntu20.04/habanalabs/pytorch-installer-2.0.1:latest: not found
```
Indeed, there is no `latest` tag here: https://vault.habana.ai/ui/native/gaudi-docker/1.11.0/ubuntu20.04/habanalabs/pytorch-installer-2.0.1/

PR #444, opened 2023-10-03
option to not crop in dataset run

ssarkar2:
# What does this PR do?
Currently for dataset run, the dataset will be cropped to fixed len. Add option to not crop

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -594,7 +594,6 @@ def main():
                 if torch.is_tensor(batch[t]):
                     batch[t] = batch[t].to(args.device)
             # Generate new sequences
-            print(batch['input_ids'].shape)
             outputs = model.generate(
                 **batch,
                 generation_config=generation_config,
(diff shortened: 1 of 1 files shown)


libinta (approved):

examples/text-generation/run_generation.py
```diff
@@ -544,7 +546,7 @@ def generate():
         raw_dataset = raw_dataset.remove_columns([name for name in raw_dataset.column_names if name != column_name])
 
         # Set the prompt length to args.max_input_tokens if > 0 else default to 16
```
regisss, line 548: Can you also update this comment for the -1 case?

regisss (commented):
I just left a very minor comment

examples/text-generation/run_generation.py
ssarkar2: thanks, fixed

regisss (approved):
LGTM!

PR #457, opened 2023-10-09
make sure a match between logits and logits_dtype in evaluation_loop

xuwh-intel:
# What does this PR do?

In the trainer.py/evaluation_loop function, `logits = self.preprocess_logits_for_metrics(logits, labels)` will change the type of logits from tuple to tensor.  But the logits_dtype obtained from `get_dtype(logits)` is still a list type and does not match the shape of the new logits. this will cause the following error in `all_preds = convert_into_dtypes(all_preds, logits_dtype)`.  So I re-get dtype of logits after `logits = self.preprocess_logits_for_metrics(logits, labels)` to ensure a match between logits and logits_dtype.

image

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -1556,6 +1556,7 @@ class GaudiTrainer(Trainer):
                 logits = self.accelerator.pad_across_processes(logits, dim=1, pad_index=-100)
                 if self.preprocess_logits_for_metrics is not None:
                     logits = self.preprocess_logits_for_metrics(logits, labels)
+                    logits_dtype = get_dtype(logits)
                 logits = self.accelerator.gather_for_metrics((logits))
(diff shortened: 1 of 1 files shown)


regisss:
@xuwh-intel Ooh interesting! Could you share a command so that I reproduce this issue please? I would like to understand why it doesn't happen in the CI examples.

xuwh-intel:
> @xuwh-intel Ooh interesting! Could you share a command so that I reproduce this issue please? I would like to understand why it doesn't happen in the CI examples.

It looks like the model I tested does not set `keys_to_ignore_at_inference = ["past_key_values"]` in configuration, which caused this error.

regisss:
@xuwh-intel Is it a model from the Transformers library and that has not been optimized for Gaudi in Optimum Habana?

xuwh-intel:
> @xuwh-intel Is it a model from the Transformers library and that has not been optimized for Gaudi in Optimum Habana?

Yes, it's chatglm2.

PR #459, opened 2023-10-11
Add gradient_checkpointing field to loading configuration for audio-classification

vivekgoe:
# What does this PR do?
https://huggingface.co/facebook/wav2vec2-base/blob/main/config.json model has gradient_checkpointing set to True. Ideally we would expect this to be overriden by training_args.gradient_checkpointing value (default is False), but this is not happening when executing run_audio_classification.py (perhaps because directly setting gradient_checkpointing from training_args has been deprecated in transformers). 
To fix this issue, added gradient_checkpoiting as part of fields being updated  when loading config (using AutoConfig.from_pretrained). Confirmed from PT profiler traces that gradient_checkpointing is not getting activated with this change.  As expected small improvement in performance is also seen with this change.    

Fixes # (issue)

## Before submitting


vivekgoe:
@regisss please help review and submit (if it looks good).

vivekgoe:
@regisss I will try out change you have suggested in GaudiTrainer and upload a new patch if that works. Thanks.

vivekgoe:
Discarding this PR and moving changes to new one https://github.com/huggingface/optimum-habana/pull/484

PR #461, opened 2023-10-12
Fixing the relative path in config read

vineethanandh:
# What does this PR do?
This change will fix the relative path in gaudi config. Instead uses the absolute path in the current execution env.

Fixes # (issue)
Errors caused by relative gaudi config path when test is triggered from a different path.


vineethanandh (commented):
@regisss 
Can you please review this change?

vineethanandh:
@vivekgoe can you please review this change?

regisss (approved):
LGTM!

regisss:
@vineethanandh Can you also run `make style` please?

PR #465, opened 2023-10-13
Enable `torch.compile`

regisss:
# What does this PR do?

As per title.

## Before submitting

### DIFF
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -625,7 +625,6 @@ class GaudiTrainingArguments(TrainingArguments):
             from .modeling_utils import adapt_transformers_to_gaudi
 
             adapt_transformers_to_gaudi()
-            version = get_habana_frameworks_version()
             if self.use_lazy_mode:
                 assert (os.getenv('PT_HPU_LAZY_MODE') == '1' or os.getenv('PT_HPU_LAZY_MODE') == None), \
                     "Argument --use_lazy_mode used, but PT_HPU_LAZY_MODE={os.getenv('PT_HPU_LAZY_MODE')}. For lazy mode, set PT_HPU_LAZY_MODE to 1"
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -440,7 +440,6 @@ class GaudiTrainingArguments(TrainingArguments):
         if (self.torch_compile_mode is not None or self.torch_compile_backend is not None) and not self.torch_compile:
             assert get_habana_frameworks_version().minor > 12, "Torch compile is not available"
             self.torch_compile = True
-            assert int(torch.__version__.split(".")[0]) >= 2, "Graph mode is available only in PyTorch 2.x."
             assert not os.getenv("PT_HPU_LAZY_MODE", "1") != "0", "Dynamo and lazy are mutually exclusive."
             # Note: PT_HPU_LAZY_MODE=0 needs to be set before library is loaded,
             #       setting it here would be too late - hence assertion.
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -628,6 +628,8 @@ class GaudiTrainingArguments(TrainingArguments):
 
             if self.use_lazy_mode:
                 logger.info("Enabled lazy mode.")
+            # TODO: remove the block below when upgrade to SynapseAI 1.13 is done
+            # as eager mode will not be available anymore
             elif not self.torch_compile:
                 if os.getenv("PT_HPU_LAZY_MODE", "1") != "0":
                     os.environ["PT_HPU_LAZY_MODE"] = "2"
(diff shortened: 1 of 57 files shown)


optimum/habana/transformers/training_args.py
```diff
@@ -610,7 +628,7 @@ def _setup_devices(self) -> "torch.device":
 
             if self.use_lazy_mode:
                 logger.info("Enabled lazy mode.")
-            else:
+            elif not self.torch_compile:
                 os.environ["PT_HPU_LAZY_MODE"] = "2"
                 logger.info("Enabled eager mode because use_lazy_mode=False.")
```
HBN-MichalSzy, line 633: Please rewrite this part, LAZY_MODE=2 is no longer supported. We should make sure at this point that it's 1 for Lazy (we checked above 0 for compile, so it's no need here).

```
if self.use_lazy_mode:
           assert os.getenv('PT_HPU_LAZY_MODE') == '1' or os.getenv('PT_HPU_LAZY_MODE') == None, f"Argument --use_lazy_mode used, but PT_HPU_LAZY_MODE={os.getenv('PT_HPU_LAZY_MODE')}. For lazy mode, set PT_HPU_LAZY_MODE to 1"
           logger.info("Enabled lazy mode.")
```

optimum/habana/transformers/training_args.py
```diff
@@ -607,10 +625,11 @@ def _setup_devices(self) -> "torch.device":
...
 
             adapt_transformers_to_gaudi()
-
             if self.use_lazy_mode:
+                assert (os.getenv('PT_HPU_LAZY_MODE') == '1' or os.getenv('PT_HPU_LAZY_MODE') == None), \
+                    "Argument --use_lazy_mode used, but PT_HPU_LAZY_MODE={os.getenv('PT_HPU_LAZY_MODE')}. For lazy mode, set PT_HPU_LAZY_MODE to 1"
                 logger.info("Enabled lazy mode.")
-            else:
+            elif not self.torch_compile:
                 os.environ["PT_HPU_LAZY_MODE"] = "2"
```
HBN-MichalSzy, line 633: Now that there is assert in use_lazy_mode, we no longer can work with new eager mode. Please change this code to do the following:
```
if os.getenv("PT_HPU_LAZY_MODE", "1") != "0":
    os.environ["PT_HPU_LAZY_MODE"] = "2"
```

HBN-MichalSzy (approved):

optimum/habana/transformers/training_args.py
```diff
@@ -278,8 +279,7 @@ def __post_init__(self):
             raise ValueError(
                 "`--use_lazy_mode`, `--use_hpu_graphs_for_inference`, `--use_hpu_graphs_for_training` and `--gaudi_config_name` cannot be used without `--use_habana`."
             )
-
-        if use_hpu_graphs and not self.use_lazy_mode:
+        if use_hpu_graphs and (not self.use_lazy_mode and not self.torch_compile_backend):
```
regisss, line 282: We should change the message below as it's mentioning eager mode, shouldn't we?

optimum/habana/transformers/training_args.py
```diff
@@ -437,6 +437,24 @@ def __post_init__(self):
             if version.parse(version.parse(torch.__version__).base_version) < version.parse("2.0.0"):
                 raise ValueError("--optim adamw_torch_fused requires PyTorch 2.0 or higher")
 
+        if (self.torch_compile_mode is not None or self.torch_compile_backend is not None) and not self.torch_compile:
+            assert get_habana_frameworks_version().minor > 12, "Torch compile is not available"
+            self.torch_compile = True
+            assert int(torch.__version__.split(".")[0]) >= 2, "Graph mode is available only in PyTorch 2.x."
```
regisss, line 443: Can we not assume that the version of torch will necessarily be >= 2.x and remove this test? The last Docker image with a version of torch < 2.x was released quite a long time ago I think.

optimum/habana/transformers/training_args.py
```diff
@@ -610,8 +628,9 @@ def _setup_devices(self) -> "torch.device":
 
             if self.use_lazy_mode:
                 logger.info("Enabled lazy mode.")
-            else:
-                os.environ["PT_HPU_LAZY_MODE"] = "2"
+            elif not self.torch_compile:
+                if os.getenv("PT_HPU_LAZY_MODE", "1") != "0":
+                    os.environ["PT_HPU_LAZY_MODE"] = "2"
```
regisss, line 633: If I understood correctly, this will have to be removed when SynapseAI 1.13 is released as eager mode won't be supported anymore right?

regisss (commented):
I pushed a new commit to fix the merge conflict and the style. I also left a few comments.

optimum/habana/transformers/training_args.py
HBN-MichalSzy: yes.

optimum/habana/transformers/training_args.py
regisss: Sounds good

PR #475, opened 2023-10-19
Upgrade to Transformers 4.34

regisss:
# What does this PR do?

As per title.

## Before submitting


optimum/habana/transformers/models/falcon/modeling_falcon.py
```diff
@@ -51,10 +51,29 @@ def gaudi_falcon_rotary_embedding_forward(self, query, key, seq_len, position_id
     """
     cos, sin = self.cos_sin(seq_len, past_key_values_length, position_ids, query.device, query.dtype)
 
+    # Query and key's shapes are [bs * num_heads, seq_len, dim], might need manual expansion. Ifs and elses used to
```
schoi-habana, line 54: @regisss I think we should discard this change. It leads to a significant performance drop. Also, it can handle cases where the expansion_factor > 1 even without this change

schoi-habana (changes requested):

optimum/habana/transformers/models/falcon/modeling_falcon.py
regisss: Sounds good

regisss: I just removed it

schoi-habana: @regisss From internal tests it turned out this is needed for some configurations. Without this they got graph compilation error. Please bring this change back.

regisss: Done.
What about the performance drop? Was it when doing text generation or during training?

regisss: @schoi-habana We have to remove it because the performance drop is quite important for both training and inference. I don't seen any compilation error with SynapseAI 1.12.0.
@libinta Maybe something to monitor to make sure everything works as expected for the release of 1.13?

libinta (approved):

PR #477, opened 2023-10-19
Update GaudiNIC multi-node-training dockerfile and setup

yeonsily:
# What does this PR do?

Fixes # (issue)

## Before submitting


libinta (commented):
since we give example of 4 nodes, should we update the hostfile template with 4 nodes?

examples/multi-node-training/README.md
```diff
@@ -44,8 +46,26 @@ docker run -it --runtime=habana -e HABANA_VISIBLE_DEVICES=all -e OMPI_MCA_btl_va
 
 > For AWS DL1 instances, `--privileged` must be passed to the `docker run` command so that EFA interfaces are visible.
 
-Finally, you will have to copy the public key of the leader node in the `~/.ssh/authorized_keys` file of all other nodes to enable password-less SSH.
 
+You will have to copy id_rsa.pub contents from every node's docker to every other node's docker's ~/.ssh/authorized_keys (all public keys need to be in all hosts' authorized_keys) to enable password-less SSH.
```
regisss, line 50: Weird I never had to do that on AWS DL1 instances with EFA

regisss (commented):
I think we should also have a `multi-node-training/efa/` folder where we store the other Dockerfile and `.deepspeed_env` and clearly state in the README that we provide images for both Gaudi NIC and EFA cases.

examples/multi-node-training/README.md
yeonsily: Created EFA folder and move AWS dockerfile and deepspeed_env to there. 
But this step is mentioned previously also and I just explained step by step details.

regisss: Yes but in my understanding we only need to copy the key of the leader node to every other bode's authorized keys, not every key.

regisss (commented):
We also need to update this link: https://github.com/huggingface/optimum-habana/blob/0fb6bb6b7bbfb06b85a6c10f113b8283aa028b02/docs/source/usage_guides/multi_node_training.mdx#L126

examples/multi-node-training/README.md
yeonsily: I had to follow the step from https://github.com/HabanaAI/Model-References/tree/master/PyTorch/nlp/DeepSpeedExamples/deepspeed-bert#set-up-password-less-ssh for GaudiNIC.

Let me try just leader node key also.

yeonsily: I confirmed that we only need to copy leader node's key. I updated the README.

regisss (approved):

PR #484, opened 2023-10-25
Enable/disable gradient_checkpointing as per training_args.gradient_checkpointing value

vivekgoe:
# What does this PR do?
https://huggingface.co/facebook/wav2vec2-base/blob/main/config.json  model has gradient_checkpointing set to True. Ideally we would expect this to be overriden by training_args.gradient_checkpointing value (default is False), but this is not happening when executing run_audio_classification.py
To fix this issue, explicitly call model.gradient_checkpoint_disable() if training_args.gradient_checkpointing = False. Confirmed from PT profiler traces that gradient_checkpointing is not getting activated with this change.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -622,6 +622,8 @@ class GaudiTrainer(Trainer):
                 import transformers.models.t5.modeling_t5 as modeling_t5
 
                 modeling_t5.checkpoint = torch.utils.checkpoint.checkpoint
+        else:
+            self.model.gradient_checkpointing_disable()
 
         model = self._wrap_model(self.model_wrapped)
 
--- a/notebooks/AI_HW_Summit_2022.ipynb
+++ b/notebooks/AI_HW_Summit_2022.ipynb
@@ -177,10 +177,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "training_args[\"use_habana\"] = True                  # Whether to use HPUs or not\n",
-    "training_args[\"use_lazy_mode\"] = True               # Whether to use lazy or eager mode\n",
+    "training_args[\"use_habana\"] = True  # Whether to use HPUs or not\n",
+    "training_args[\"use_lazy_mode\"] = True  # Whether to use lazy or eager mode\n",
     "training_args[\"gaudi_config_name\"] = \"Habana/gpt2\"  # Gaudi configuration to use\n",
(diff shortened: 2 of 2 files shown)


vivekgoe:
@regisss this PR is continuation of https://github.com/huggingface/optimum-habana/pull/459 (closed because I messed it up during rebase).  Have taken care of your review comments from #459, please review again. Thanks.

regisss (commented):
LGTM!
Can you remove the changes of the notebook please?

regisss:
I see there's also an error in CI:
```
AttributeError: 'RegressionDictModel' object has no attribute 'gradient_checkpointing_disable'
```
Let me take a closer look.

optimum/habana/transformers/trainer.py
```diff
@@ -622,6 +622,8 @@ def hpu_deepspeed_checkpointing(function, *checkpoint_args):
                 import transformers.models.t5.modeling_t5 as modeling_t5
 
                 modeling_t5.checkpoint = torch.utils.checkpoint.checkpoint
+        else:
+            self.model.gradient_checkpointing_disable()
```
regisss, line 626:
```suggestion
            # Hack because `RegressionModel` in test_trainer.py doesn't have `gradient_checkpointing_disable`
            if hasattr(self.model, "gradient_checkpointing_disable"):
                self.model.gradient_checkpointing_disable()
```

regisss (commented):
Many tests actually use a toy model that inherits from `torch.nn.Module` and not from `PreTrainedModel` so it doesn't have the `gradient_checkpointing_disable` method.
I propose a simple fix below, it's quite hacky but reworking the tests would be quite tedious and I prefer keeping them as close as possible to Transformers' tests.

optimum/habana/transformers/trainer.py
vivekgoe: Updated as per suggestion, thanks.

regisss (commented):
LGTM!
Can you also remove the formating changes to the notebook?

vivekgoe:
@regisss I updated as per your comments, CI passes too.  Running make style is always making formatting changes to notebook in my workspace (even after rebasing to latest commit on main). Since style check in CI is passing with this I assume its ok, please let me know if that is not correct assumption.

regisss (approved):
There are rare occurrences where `black` or `ruff` produce different results on different environments, which is weird. But okay let's merge.

PR #489, opened 2023-10-27
Fix split validation dataset problem

mandy-li:
If a dataset doesn't have validation subset, partial of data will be split into tokenized_datasets["validation"]. but later tokenized_datasets["test"] was used for eval.


sywangyi (approved):

regisss (approved):
LGTM

PR #496, opened 2023-10-30
add mistral support for generation

sywangyi:
add mistral static shape support for generation


sywangyi:
based the implementation of mistral in transformers v4.34.1

regisss:
Thanks for adding support for this popular model! I'll review this in the beginning of next week :slightly_smiling_face:

regisss (commented):
LGTM!
Could you check if the model runs faster when removing all mentions to `padding_mask` please? I observed that with Llama.

sywangyi:
Hi, @regisss , I use https://huggingface.co/mistralai/Mistral-7B-v0.1 for test, and _flash_attn_2_enabled is not set in config, so padding_mask is not used.

regisss (approved):
@sywangyi I just added CI baselines for Gaudi and Gaudi2 and I'll merge once the CI is done

PR #513, opened 2023-11-09
Add infra to enable/disable dynamic shapes feature through gaudi_config

vivekgoe:
# What does this PR do?
Adds code to enable/disable dynamic shapes support (in Synapse SW) through gaudi_config.json

Fixes # (issue)

## Before submitting


regisss (commented):
It looks good to me! I think we should also had a test: either a regression test (if it's relevant to monitor speed and accuracy numbers) or just an integration test to make sure a very basic training runs without any issues, WDYT?

vivekgoe:
@regisss I am trying to add a integration test into test_trainer.py, for this I need a dataset with variable length input tensors. I was not able to modify existing dataset in test file to generate variable length inputs, do you have any suggestions on what will be quickest way of doing it? Will really appreciate your help.

regisss:
> @regisss I am trying to add a integration test into test_trainer.py, for this I need a dataset with variable length input tensors. I was not able to modify existing dataset in test file to generate variable length inputs, do you have any suggestions on what will be quickest way of doing it? Will really appreciate your help.

Another possibility would be to run an example without padding examples. For instance, the [text-classification example](https://github.com/huggingface/optimum-habana/tree/main/examples/text-classification) with the argument `--pad_to_max_length False`.
Do we want to just check that it works or to also monitor the throughput?

vivekgoe:
@regisss that is a good idea, let me try it out. I just want to check that feature works to catch any potential regressions.

vivekgoe:
@regisss I have added a test, please review again and merge if it looks ok. Thanks.

tests/test_trainer.py
```diff
@@ -670,6 +690,50 @@ def setUp(self):
...
+        # Run training with variable length inputs and enable dynamic shapes support
+        train_dataset = RegressionDatasetDynamic(length=256)
+        gaudi_config = get_gaudi_config()
+        gaudi_config.use_dynamic_shapes = True
+        args = GaudiTrainingArguments(
+            "./regression", use_habana=True, use_lazy_mode=True, per_device_train_batch_size=1, num_train_epochs=1
+        )
+        model = RegressionModel()
+        optimizer = torch.optim.SGD(model.parameters(), lr=1.0)
+        lr_scheduler = torch.optim.lr_scheduler.LambdaLR(optimizer, lr_lambda=lambda x: 1.0)
```
regisss, line 703: I think we don't need that, default ones should be fine

tests/test_trainer.py
```diff
@@ -670,6 +690,50 @@ def setUp(self):
...
+        # Run training again with variable length inputs and disable dynamic shapes support
+        train_dataset = RegressionDatasetDynamic(length=256)
+        gaudi_config = get_gaudi_config()
+        gaudi_config.use_dynamic_shapes = False
+        args = GaudiTrainingArguments(
+            "./regression", use_habana=True, use_lazy_mode=True, per_device_train_batch_size=1, num_train_epochs=1
+        )
+        model = RegressionModel()
+        optimizer = torch.optim.SGD(model.parameters(), lr=1.0)
+        lr_scheduler = torch.optim.lr_scheduler.LambdaLR(optimizer, lr_lambda=lambda x: 1.0)
```
regisss, line 721: Same

regisss (commented):
Thanks @vivekgoe, it looks good!
I just left a few additional very minor comments and then we can merge :slightly_smiling_face:

vivekgoe:
@regisss updated as per your review comments, please check. Thanks.

regisss (approved):
LGTM!

PR #524, opened 2023-11-14
fixing update method - avoid copy idx to cpu which splitting the graph

bgoldberg-habana:
fixing update method - avoid copy idx to cpu which splitting the graph
issue description:
in the update method when calling to 
prev[:, :, :idx, :].copy_(cur)

the split method is using scalars and not tensors, so pytorch convert the tensor to scalar and only then call to split..
so once we need to convert the tensor to scalar, the bridge is doing mark step and splitting the graph until this point..


libinta (approved):

dvarshney-habana (approved):

regisss (approved):
LGTM

PR #525, opened 2023-11-14
Sarkar/dyn prompt

ssarkar2:
# What does this PR do?

1. Fix issue when prompt_len is a multiple of bucket
2. add `--simulate_dynamic_prompt`, which lets us simulate dynamic input. `adjust_batch` function adjusts the batch lengths based on inputs from `--simulate_dynamic_prompt`
3. Add appropriate warmup for dynamic inputs based on min and max of `--simulate_dynamic_prompt`
4. add `--reduce_recompilation` for extra optimizations for time-to-first-token in dynamic case. This does the following optimizations:
  - Preprocess on CPU (pad on CPU)
  - compute position_idx on CPU

## Before submitting

### DIFF
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -199,3 +199,5 @@ python run_generation.py \
 ```
 
 `--bucket_size` option is especially useful when processing an input stream with varying lengths, that is when you have something like `--dataset_name squad --column_name context --max_input_tokens -1`. `--max_input_tokens -1` specifies no truncation of input prompt in the dataset.
+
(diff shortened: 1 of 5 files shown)


ssarkar2 (commented):
Add readme, something like:

Another way to simulate dynamic input is to use `--simulate_dyn_prompt`. For example `--simulate_dyn_prompt 25,35,45` will extend or crop the default prompt (or the prompt passed in using `--prompt`) to sizes 25, 35, and 45, and throughput will be measured for these 3 lengths. If `--simulate_dyn_prompt` is used, the min and max input lengths from it are computed to perform warmup as well. One final optimization that can be used in case of dynamic inputs is `--reduce_recompile`. Thus the suggested configuration to simulate dynamicity after warmup is to use all three arguments: `--simulate_dyn_prompt 25,35,45 --reduce_recompile --bucket 30`

ssarkar2 (commented):
run `make style`

ssarkar2:
closing in favor of **https://github.com/huggingface/optimum-habana/pull/543**

PR #529, opened 2023-11-15
Add load_meta_device option to reduce host RAM

jiminha:
# What does this PR do?
Add --load_meta_device option to reduce host RAM for falcon40b

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/language-modeling/run_lora_clm.py
+++ b/examples/language-modeling/run_lora_clm.py
@@ -134,6 +134,16 @@ class ModelArguments:
         default=False,
         metadata={"help": ("Whether to run attention softmax layer in bf16 precision",)},
     )
+    load_meta_device: bool = field(
+        default=False,
+        metadata={
+           "help": (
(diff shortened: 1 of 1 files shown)


regisss:
Nice! Would that work for Llama too?

jiminha:
> Nice! Would that work for Llama too?

regisss:
Perf degradation is weird, that should not happen. I'm going to take a look at it, I'm suspecting something with PEFT.

libinta (approved):

regisss:
Actually perf degradation is expected. What happens here, when using `device_map`, is that the model is split across devices and then naive pipeline parallelism is performed. Hence the drop in performance.
@jiminha Have you tried other device map policies like `"balanced"` and `"sequential"`?

regisss:
"balanced" and "sequential" both gives better throughput but I didn't monitor the host memory usage

regisss:
Hmm okay nevermind, using "balanced" and "sequential" doesn't help in terms of memory

regisss (approved):
LGTM!

PR #531, opened 2023-11-16
Fix for attn_softmax_bf16 when generation_config is None

schoi-habana:
This addresses an error from the attn_softmax_bf16 change in cases where self.model.generation_config exists but is 'None'.

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -853,8 +853,8 @@ class GaudiTrainer(Trainer):
                     self.control = self.callback_handler.on_step_begin(args, self.state, self.control)
 
                 # attn_softmax_bf16 is enabled only for llama
-                if self.model.config.model_type == "llama":
-                    if self.model.generation_config.attn_softmax_bf16:
+                if hasattr(self.model, "generation_config"):
+                    if self.model.config.model_type == "llama" and self.model.generation_config.attn_softmax_bf16:
                         inputs["attn_softmax_bf16"] = True
 
                 # TODO: keep syncs for fast DDP?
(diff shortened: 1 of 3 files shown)


optimum/habana/transformers/trainer.py
```diff
@@ -853,7 +853,7 @@ def hpu_deepspeed_checkpointing(function, *checkpoint_args):
                     self.control = self.callback_handler.on_step_begin(args, self.state, self.control)
 
                 # attn_softmax_bf16 is enabled only for llama
-                if hasattr(self.model, "generation_config"):
+                if hasattr(self.model, "generation_config") and self.model.generation_config is not None:
```
regisss, line 856: Not sure we need to add this. If `self.model.generation_config.attn_softmax_bf16` is None, the condition `self.model.generation_config.attn_softmax_bf16` in the if-statement right below will not be True and thus nothing is done no?

schoi-habana: We met an error saying:  'NoneType' object has no attribute 'attn_softmax_bf16' so had to check if self.model.generation_config is not None

regisss: Ah my bad I read the code too quickly!
How did you get that error? What was the command or use case?

mandy-li: We got the error why FT llama2-7b

mandy-li (approved):

optimum/habana/transformers/trainer.py
schoi-habana: for sequence classification

PR #532, opened 2023-11-16
Improve llama performance and reduce memory consumption by updating sin/cos cache when inferring more than max position embeddings (4096) 

puneeshkhanna:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -679,9 +679,7 @@ class GaudiGenerationMixin(GenerationMixin):
                     )
             if self.config.model_type in ["llama"]:
                 if self.config.max_position_embeddings < calculated_max_length:
-                    unwrap_deepspeed_model(self).update_sincos_cache(
-                        seq_len=calculated_max_length
-                    )
+                    unwrap_deepspeed_model(self).update_sincos_cache(seq_len=calculated_max_length)
 
         # 7. determine generation mode
         generation_mode = self._get_generation_mode(generation_config, assistant_model)
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -677,6 +677,11 @@ class GaudiGenerationMixin(GenerationMixin):
                     unwrap_deepspeed_model(self).allocate_kv_cache(
                         bs * generation_config.num_beams, calculated_max_length
                     )
+            if self.config.model_type in ["llama"]:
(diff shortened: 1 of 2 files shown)


puneeshkhanna:
All testing results with the fix highlighted in below snapshot:

image

To summarize:
1. Bloom 7B (when seq len > 4096) 1x perf improves by ~7% and memory consumption reduces by ~11 GB in some cases.
2. Bloom 13B (when seq len > 4096) 1x perf improves by ~3.5-5% and memory consumption reduces by ~14 GB in some cases.
3. When seq len < 4096, no change in perf as well as memory which is expected. Minor run to run variations.
4. Bloom 70B (when seq len > 4096) 8x perf improves by ~6-11% and memory consumption reduces by ~500-1000 MB.

puneeshkhanna:
Just to further highlight the actual issue via below code of LlamaRotaryEmbedding,  self.max_seq_len_cached is initially set to 4096 during model init() and when actual seq len > 4096 then self._set_cos_sin_cache can always get executed as part of HPU graphs model forward pass causing increase in memory as well as reduced performance. 
So via the fix in this PR, we are updating cos sin cache beforehand based upon actual seq len.

    def _set_cos_sin_cache(self, seq_len, device, dtype):
        self.max_seq_len_cached = seq_len
        t = torch.arange(self.max_seq_len_cached, device=device, dtype=self.inv_freq.dtype)

        freqs = torch.einsum("i,j->ij", t, self.inv_freq)
        # Different from paper, but it uses a different permutation in order to obtain the same calculation
        emb = torch.cat((freqs, freqs), dim=-1)
        self.register_buffer("cos_cached", emb.cos()[None, None, :, :].to(dtype), persistent=False)
        self.register_buffer("sin_cached", emb.sin()[None, None, :, :].to(dtype), persistent=False)

    def forward(self, x, seq_len=None):
        # x: [bs, num_attention_heads, seq_len, head_size]
        if seq_len > self.max_seq_len_cached:
            self._set_cos_sin_cache(seq_len=seq_len, device=x.device, dtype=x.dtype)
        return (
            self.cos_cached[:, :, :seq_len, ...].to(dtype=x.dtype),
            self.sin_cached[:, :, :seq_len, ...].to(dtype=x.dtype),
        )

puneeshkhanna:
@regisss - please trigger CI and review.

regisss (approved):
LGTM, nice improvement!

PR #537, opened 2023-11-19
Enable Falcon 180B

hlahkar:
# What does this PR do?

This PR enables Falcon 180B Finetuning.

### DIFF
--- a/examples/language-modeling/README.md
+++ b/examples/language-modeling/README.md
@@ -350,7 +350,8 @@ python run_clm.py \
 ## PEFT
 
 To run LoRA finetuning, you can use `run_lora_clm.py`.
-Here are single-/multi-device command examples for Llama1-7B, Falcon-40B and Llama2-70B:
+Here are single-/multi-device command examples for Llama1-7B, Falcon-40B and Llama2-70B.
+You can also use multicard version for Falcon-180B:
 
 - Single-card finetuning of Llama1-7B:
 ```bash
@@ -522,7 +523,7 @@ python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
 
 - Multi-card finetuning of Falcon-180B:
 ```bash
-DEEPSPEED_HPU_ZERO3_SYNC_MARK_STEP_REQUIRED=1 HABANA_PROFILE=1 LOWER_LIST=ops_bf16.txt python3 ../gaudi_spawn.py \
+DEEPSPEED_HPU_ZERO3_SYNC_MARK_STEP_REQUIRED=1 LOWER_LIST=ops_bf16.txt python3 ../gaudi_spawn.py \
     --world_size 8 --use_deepspeed run_lora_clm.py \
     --model_name_or_path tiiuae/falcon-180B \
     --dataset_name timdettmers/openassistant-guanaco \
--- a/examples/language-modeling/README.md
+++ b/examples/language-modeling/README.md
@@ -522,8 +522,8 @@ python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
 ````
 
 - Multi-card finetuning of Falcon-180B:
-- Falcon-180B example command saves only the LoRA parameters at end
-- For inference we need to merge the pretrained model and LoRA weights
+  - Falcon-180B example command saves only the LoRA parameters at end
+  - For inference we need to merge the pretrained model and LoRA weights
 ```bash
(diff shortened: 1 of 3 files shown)


examples/language-modeling/README.md
```diff
@@ -453,8 +453,43 @@ LOWER_LIST=ops_bf16.txt python3 ../gaudi_spawn.py \
...
+    --dataset_name timdettmers/openassistant-guanaco \
+    --bf16 True \
+    --output_dir ./model_lora_falcon_ddp \
+    --num_train_epochs 3 \
+    --per_device_train_batch_size 1 \
+    --per_device_eval_batch_size 1 \
+    --gradient_accumulation_steps 16 \
+    --evaluation_strategy "no" \
+    --save_strategy "steps" \
+    --save_steps 195 \
```
regisss, line 471: 195 steps corresponds to one epoch?

examples/language-modeling/run_lora_clm.py
```diff
@@ -92,6 +92,10 @@ class ModelArguments:
         default=None,
         metadata={"help": "Where do you want to store the pretrained models downloaded from huggingface.co"},
     )
+    auth_token: Optional[str] = field(
+        default=None,
+        metadata={"help": "auth token for private models"},
+    )
```
regisss, line 98: Let's call it `token` as in other examples: https://github.com/huggingface/optimum-habana/blob/3a13a0141d7417b709fe22fd177228c834117bbf/examples/language-modeling/run_clm.py#L122

regisss (commented):
This only works from SynapseAI 1.13 right?

examples/language-modeling/README.md
vivekgoe: @regisss 195 steps is greater than 3 epochs for this case, we added this number as a work-around to avoid CPU OOM we see when checkpoint save is triggered at the end of training.  Now we have figured out that we can avoid full model checkpoint (by not adding --save_steps) and instead save only LoRA trained parameters from run_lora_clm.py, so we will remove --save_steps 195 altogether.

examples/language-modeling/run_lora_clm.py
vivekgoe: Will update in next patch.

examples/language-modeling/README.md
```diff
@@ -321,7 +321,7 @@ python run_clm.py \
 
 ## PEFT
 
-To run LoRA finetuning and inference. you could use `run_lora_clm.py` as an example. Multi-card examples can be simply adapted to run LoRA finetuning. Here is the CLM example with Llama1-7B and Falcon-40B:
+To run LoRA finetuning and inference. you could use `run_lora_clm.py` as an example. Multi-card examples can be simply adapted to run LoRA finetuning. Here is the CLM example with Llama1-7B, Falcon-40B and Falcon-180B:
```
vivekgoe, line 324: For Falcon-180B specify that only finetuning is validated.

regisss:
I pushed a commit because there were a few issues with the code block examples in the README.

One question, when you say "For inference we need to merge the pretrained model and LoRA weights", do you mean that users have to call `.merge_and_unload()` (see [here](https://github.com/huggingface/optimum-habana/blob/0dfda27dd0351d0753c35de2abb752309e2383b4/examples/text-generation/utils.py#L208C17-L208C36)) so that inference works? Or just that you need both checkpoints to run the LoRA model?

vivekgoe:
@regisss from the code below in run_lora_clm.py it looks like full model is being saved (not only LoRA parameters), what do you think?  @hlahkar please check this on your side.

https://github.com/huggingface/optimum-habana/blob/0dfda27dd0351d0753c35de2abb752309e2383b4/examples/language-modeling/run_lora_clm.py#L648C66-L648C66

libinta:
I remembered when we tested falcon40b lora training checkpoint, we have to have both checkpoints to run inference. I remembered we tried load_and_merged call, it doesn't work well at that time

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

hlahkar:
> I pushed a commit because there were a few issues with the code block examples in the README.
> 
> One question, when you say "For inference we need to merge the pretrained model and LoRA weights", do you mean that users have to call `.merge_and_unload()` (see [here](https://github.com/huggingface/optimum-habana/blob/0dfda27dd0351d0753c35de2abb752309e2383b4/examples/text-generation/utils.py#L208C17-L208C36)) so that inference works? Or just that you need both checkpoints to run the LoRA model?

@regisss Yes we are only storing the LoRA weights in this case; and the user will have to use '.merge_and_unload' to make inference work.

regisss (approved):

PR #540, opened 2023-11-20
Fix split by whitespaces not a single space

oelayan7:
Example for a run:
====================================
output.stdout:
'habana-torch-plugin     1.14.0.71         \n'
====================================
output.stdout.split("\n")[0].split(" "):
['habana-torch-plugin', '', '', '', '', '1.14.0.71', '', '', '', '', '', '', '', '', '']
so taking the -1 will result in an empty string
====================================
output.stdout.split("\n")[0].split():
['habana-torch-plugin', '1.14.0.71']
so taking -1 will result in the version

### DIFF
--- a/optimum/habana/utils.py
+++ b/optimum/habana/utils.py
@@ -215,7 +215,7 @@ def get_habana_frameworks_version():
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
     )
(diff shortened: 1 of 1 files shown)


regisss (approved):
LGTM!
I don't see this issue with Synapse AI 1.12 but your suggestion looks more robust.

PR #542, opened 2023-11-20
Add llamav2 fp8 inference

bgoldberg-habana:
Add llamav2 fp8 inference
adding kv cache in fp8 option

### DIFF
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -99,7 +99,7 @@ Here are a few settings you may be interested in:
 - `--attn_softmax_bf16` to run attention softmax layer in bfloat16 precision provided that the model (such as Llama) supports it
 - `--trim_logits` to calculate logits only for the last token in the first time step provided that the model (such as Llama) supports it
 - `--kv_cache_fp8` Store kv-cache in float8 when kv-cache is used
-- `--enable_quantization` Enable Quantization to fp8
+- `--enable_synapse_quantization` Enable Synapse Quantization to fp8
 
 For example, you can reproduce the results presented in [this blog post](https://huggingface.co/blog/habana-gaudi-2-bloom) with the following command:
 ```bash
@@ -216,7 +216,7 @@ python run_generation.py \
 `--bucket_size` option is especially useful when processing an input stream with varying lengths, that is when you have something like `--dataset_name squad --column_name context --max_input_tokens -1`. `--max_input_tokens -1` specifies no truncation of input prompt in the dataset.
 
 ### Running with FP8
-llama-7b in FP8 is enabled. use --enable_quantization to enable synapse quantization in fp8.
+llama-7b in FP8 is enabled. use --enable_synapse_quantization to enable synapse quantization in fp8.
 Add the `--kv_cache_fp8` argument to run the model with a KV cache allocated in fp8.
 More information on enabling fp8 in Synapse is here:
 https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html
@@ -235,6 +235,6 @@ python run_generation.py \
 --max_new_tokens 200 \
 --batch_size=2 \
 --kv_cache_fp8 \
---enable_quantization \
+--enable_synapse_quantization \
 ```
 
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -99,7 +99,7 @@ Here are a few settings you may be interested in:
 - `--attn_softmax_bf16` to run attention softmax layer in bfloat16 precision provided that the model (such as Llama) supports it
 - `--trim_logits` to calculate logits only for the last token in the first time step provided that the model (such as Llama) supports it
 - `--kv_cache_fp8` Store kv-cache in float8 when kv-cache is used
-- `--enable_synapse_quantization` Enable Synapse Quantization to fp8
+- `--fp8` Enable Quantization to fp8
 
 For example, you can reproduce the results presented in [this blog post](https://huggingface.co/blog/habana-gaudi-2-bloom) with the following command:
 ```bash
@@ -217,7 +217,7 @@ python run_generation.py \
 
 ### Running with FP8
 
-Llama2-7b in FP8 is enabled. Use `--enable_synapse_quantization` to enable quantization in fp8.
+Llama2-7b in FP8 is enabled. Use `--fp8` to enable quantization in fp8.
 Add the `--kv_cache_fp8` argument to run the model with a KV cache allocated in fp8.
 More information on enabling fp8 in SynapseAI is available here:
 https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html
@@ -229,13 +229,12 @@ python run_generation.py \
 --model_name_or_path meta-llama/Llama-2-7b-hf \
 --use_hpu_graphs \
 --use_kv_cache \
---bf16 \
 --reuse_cache \
 --trim_logits \
 --attn_softmax_bf16 \
 --max_new_tokens 200 \
 --batch_size=2 \
 --kv_cache_fp8 \
---enable_synapse_quantization \
+--fp8 \
 ```
 
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -99,7 +99,7 @@ Here are a few settings you may be interested in:
 - `--attn_softmax_bf16` to run attention softmax layer in bfloat16 precision provided that the model (such as Llama) supports it
 - `--trim_logits` to calculate logits only for the last token in the first time step provided that the model (such as Llama) supports it
 - `--kv_cache_fp8` Store kv-cache in float8 when kv-cache is used
-- `--enable_quantization_file` Enable Quantization to fp8
+- `--enable_quantization` Enable Quantization to fp8
 
 For example, you can reproduce the results presented in [this blog post](https://huggingface.co/blog/habana-gaudi-2-bloom) with the following command:
 ```bash
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -216,9 +216,10 @@ python run_generation.py \
 `--bucket_size` option is especially useful when processing an input stream with varying lengths, that is when you have something like `--dataset_name squad --column_name context --max_input_tokens -1`. `--max_input_tokens -1` specifies no truncation of input prompt in the dataset.
 
 ### Running with FP8
-llama-7b in FP8 is enabled. use --enable_synapse_quantization to enable synapse quantization in fp8.
+
+Llama2-7b in FP8 is enabled. Use `--enable_synapse_quantization` to enable quantization in fp8.
 Add the `--kv_cache_fp8` argument to run the model with a KV cache allocated in fp8.
-More information on enabling fp8 in Synapse is here:
+More information on enabling fp8 in SynapseAI is available here:
 https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html
 
 Here is an example:
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -224,7 +224,7 @@ https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP
 
 Here is an example:
 ```bash
-UPDATE_MME_OUTPUT_PRECISION_FILTER="v_proj,matmul_av" USE_DEFAULT_QUANT_PARAM=true UPDATE_GRAPH_OUTPUT_MME=false ENABLE_CALC_DYNAMIC_RANGE=false ENABLE_EXPERIMENTAL_FLAGS=true
+USE_DEFAULT_QUANT_PARAM=true UPDATE_GRAPH_OUTPUT_MME=false ENABLE_CALC_DYNAMIC_RANGE=false ENABLE_EXPERIMENTAL_FLAGS=true
 python run_generation.py \
 --model_name_or_path meta-llama/Llama-2-7b-hf \
 --use_hpu_graphs \
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -98,6 +98,8 @@ Here are a few settings you may be interested in:
 - `--prompt` to benchmark the model on one or several prompts of your choice
 - `--attn_softmax_bf16` to run attention softmax layer in bfloat16 precision provided that the model (such as Llama) supports it
 - `--trim_logits` to calculate logits only for the last token in the first time step provided that the model (such as Llama) supports it
+- `--kv_cache_fp8` Store kv-cache in float8 when kv-cache is used
+- `--enable_quantization_file` Enable Quantization to fp8
 
 For example, you can reproduce the results presented in [this blog post](https://huggingface.co/blog/habana-gaudi-2-bloom) with the following command:
 ```bash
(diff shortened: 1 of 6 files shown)


examples/text-generation/run_generation.py
```diff
@@ -206,6 +206,13 @@ def setup_parser(parser):
     )
     parser.add_argument("--verbose_workers", action="store_true", help="Enable output from non-master workers")
 
+    parser.add_argument(
+        "--kv_cache_fp8",
+        action="store_true",
+        help="Store kv-cache in float8 when kv-cache is used",
+    )
+    parser.add_argument("--enable_quantization", "-eq", action="store_true", help="Enable Quantization to fp8")
```
regisss, line 214: Having 2 arguments seems a bit redundant, can't we keep only `--kv_cache_fp8`? Or do they have different meaning?

examples/text-generation/README.md
```diff
@@ -212,3 +214,25 @@ python run_generation.py \
 ```
 
 `--bucket_size` option is especially useful when processing an input stream with varying lengths, that is when you have something like `--dataset_name squad --column_name context --max_input_tokens -1`. `--max_input_tokens -1` specifies no truncation of input prompt in the dataset.
+
+### Running with FP8
+llama-7b in FP8 is enabled using these instructions to enable FP8 in synapse:https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html
+use --kv_cache_fp8 to run with kv cache allocated in fp8
```
regisss, line 220:
```suggestion

To use Llama2-7b in FP8, use these instructions to enable FP8 in SynapseAI: https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html
Add the `--kv_cache_fp8` argument to run the model with a KV cache allocated in fp8.
```

regisss (commented):
I left a couple of comments.
Additional question: does this only work with the 7B-parameter checkpoint of Llama2 or are the 13B/70B checkpoints compatible too?

examples/text-generation/run_generation.py
bgoldberg-habana: they have different meaning,
enable_quantization - enable syanpse quantization in the bridge, setting all gemm operations to fp8.
kv_cache_fp8 - is only to store kv cache in fp8 - should be used when enable_quantization is turn on but it's not a must. it's saves memory and has a better perf.

examples/text-generation/README.md
bgoldberg-habana: with your suggestion, it seems like they need to do the instructions to enable fp8.
but we already done that.. they only need to run with the flag..

will fix this - 
Add the `--kv_cache_fp8` argument to run the model with a KV cache allocated in fp8.

bgoldberg-habana:
> I left a couple of comments.
> Additional question: does this only work with the 7B-parameter checkpoint of Llama2 or are the 13B/70B checkpoints compatible too?

It should work but we didn't commit to it.

examples/text-generation/run_generation.py
regisss: Okay I see, in that case I would rename `enable_quantization` by `fp8`. Or will it also enable fp4 in the future?

examples/text-generation/README.md
regisss: I'm still not convinced by the phrasing here. It makes it sound like the user should visit the Gaudi doc to make it work, which is not the case. We should rather have something like: "More information here: [link]".
Also, I would add that `--enable quantization` enables to run matrix multiplications in fp8.

bgoldberg-habana: fixed.

examples/text-generation/run_generation.py
bgoldberg-habana: let's stick with enable_quantization

regisss: IMO `enable_quantization` is misleading as there are tons of different quantization format and strategies available in the HF ecosystem. People may think that int8 or fp4 quantization is available for instance.
We also use "fp8" in Accelerate for instance: https://huggingface.co/docs/accelerate/v0.24.0/en/package_reference/kwargs#accelerate.utils.FP8RecipeKwargs
So, unless there is a strong argument for sticking to `enable_quantization`, I would prefer to have a more explicit name.

bgoldberg-habana: fixed

examples/text-generation/README.md
```diff
@@ -212,3 +214,27 @@ python run_generation.py \
 ```
 
 `--bucket_size` option is especially useful when processing an input stream with varying lengths, that is when you have something like `--dataset_name squad --column_name context --max_input_tokens -1`. `--max_input_tokens -1` specifies no truncation of input prompt in the dataset.
+
+### Running with FP8
+llama-7b in FP8 is enabled. use --enable_synapse_quantization to enable synapse quantization in fp8.
+Add the `--kv_cache_fp8` argument to run the model with a KV cache allocated in fp8.
+More information on enabling fp8 in Synapse is here:
+https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html
```
regisss, line 222:
```suggestion

Llama2-7b in FP8 is enabled. Use `--enable_synapse_quantization` to enable quantization in fp8.
Add the `--kv_cache_fp8` argument to run the model with a KV cache allocated in fp8.
More information on enabling fp8 in SynapseAI is available here:
https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html
```

regisss (commented):
Still not convinced by this arg name but anyway we can rediscuss that later.
I left a small nit and then I'll wait for the 1.13 AMI to be released on AWS to quickly test it before merging.

examples/text-generation/run_generation.py
```diff
@@ -206,6 +206,13 @@ def setup_parser(parser):
     )
     parser.add_argument("--verbose_workers", action="store_true", help="Enable output from non-master workers")
 
+    parser.add_argument(
+        "--kv_cache_fp8",
+        action="store_true",
+        help="Store kv-cache in float8 when kv-cache is used",
+    )
+    parser.add_argument("--enable_synapse_quantization", "-esq", action="store_true", help="Enable Quantization to fp8")
```
polisettyvarma, line 214: how about enable_fp8_quantization ?

regisss: Yes, it's more explicit. Or less verbose `--fp8_quantization`, or even `--fp8` that I suggested before (but that's okay if you want to keep the term "quantization").
Ideally, argument names should be self-explanatory and as little verbose as possible, which is why I proposed changing it.

bgoldberg-habana: but we are going to enable different quantization approach in the next sprint so we can differ between synapse quantization and Habana quantization tool (this is the name of the new tool.. )

bgoldberg-habana: yes and the synapse one will be deprecated when we move to 1.14

regisss: Okay I see. Would it be acceptable to have an arg name that enables synapse fp8 quantization now with 1.13 and then the same arg that enables the other tool when we switch to 1.14?

examples/text-generation/README.md
```diff
@@ -212,3 +214,28 @@ python run_generation.py \
...
+### Running with FP8
+
+Llama2-7b in FP8 is enabled. Use `--enable_synapse_quantization` to enable quantization in fp8.
+Add the `--kv_cache_fp8` argument to run the model with a KV cache allocated in fp8.
+More information on enabling fp8 in SynapseAI is available here:
+https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html
+
+Here is an example:
+```bash
+UPDATE_MME_OUTPUT_PRECISION_FILTER="v_proj,matmul_av" USE_DEFAULT_QUANT_PARAM=true UPDATE_GRAPH_OUTPUT_MME=false ENABLE_CALC_DYNAMIC_RANGE=false ENABLE_EXPERIMENTAL_FLAGS=true
```
MrGeva, line 227: let's talk about it. do not merge with this yet.

MrGeva (changes requested):

examples/text-generation/run_generation.py
regisss: @bgoldberg-habana @libinta Any idea regarding my question?

MrGeva:
LGTM

examples/text-generation/run_generation.py
bgoldberg-habana: we change the flag name to fp8

examples/text-generation/run_generation.py
```diff
@@ -211,9 +211,7 @@ def setup_parser(parser):
         action="store_true",
         help="Store kv-cache in float8 when kv-cache is used",
     )
-    parser.add_argument(
-        "--enable_synapse_quantization", "-esq", action="store_true", help="Enable Quantization to fp8"
-    )
+    parser.add_argument("--fp8", action="store_true", help="Enable Quantization to fp8")
```
ulivne, line 214: I think the flag name is misleading,  it seems like a data type and not like different flow,
also in future we may be able to quantize to other data types

ulivne (changes requested):

bgoldberg-habana:
@regisss can we merge the PR?

regisss:
> @regisss can we merge the PR?

I'm going to try it quickly with 1.13, it should be merged in a couple of hours

regisss:
@bgoldberg-habana I'm getting this error when running the example:
```
  File "run_generation.py", line 490, in <module>
    main()
  File "run_generation.py", line 294, in main
    generate()
  File "run_generation.py", line 275, in generate
    outputs = model.generate(
  File "/usr/local/lib/python3.8/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/generation/utils.py", line 680, in generate
    unwrap_deepspeed_model(self).allocate_kv_cache(
  File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 493, in allocate_kv_cache
    self.model.allocate_kv_cache(batch_size, max_seq_len, inp_seq_len, kv_cache_fp8)
  File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 331, in allocate_kv_cache
    layer.allocate_kv_cache(batch_size, max_seq_len, inp_seq_len, kv_cache_fp8)
  File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 267, in allocate_kv_cache
    self.self_attn.allocate_kv_cache(batch_size, max_seq_len, inp_seq_len, kv_cache_fp8)
  File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 103, in allocate_kv_cache
    self.past_key = torch.zeros(key_shape, dtype=dtype, device=device)
RuntimeError: Expected habana_helpers::is_supported_type(type) to be true, but got false.
```
It seems to be an issue with `torch.float8_e4m3fn` taht is not recognized. Do you know what could be the reason?

bgoldberg-habana:
> @bgoldberg-habana I'm getting this error when running the example:
> 
> ```
> Traceback (most recent call last):
>   File "run_generation.py", line 490, in <module>
>     main()
>   File "run_generation.py", line 294, in main
>     generate()
>   File "run_generation.py", line 275, in generate
>     outputs = model.generate(
>   File "/usr/local/lib/python3.8/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
>     return func(*args, **kwargs)
>   File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/generation/utils.py", line 680, in generate
>     unwrap_deepspeed_model(self).allocate_kv_cache(
>   File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 493, in allocate_kv_cache
>     self.model.allocate_kv_cache(batch_size, max_seq_len, inp_seq_len, kv_cache_fp8)
>   File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 331, in allocate_kv_cache
>     layer.allocate_kv_cache(batch_size, max_seq_len, inp_seq_len, kv_cache_fp8)
>   File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 267, in allocate_kv_cache
>     self.self_attn.allocate_kv_cache(batch_size, max_seq_len, inp_seq_len, kv_cache_fp8)
>   File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 103, in allocate_kv_cache
>     self.past_key = torch.zeros(key_shape, dtype=dtype, device=device)
> RuntimeError: Expected habana_helpers::is_supported_type(type) to be true, but got false.
> ```
> 
> It seems to be an issue with `torch.float8_e4m3fn` that is not recognized. Do you know what could be the reason?

'which version do you use? 1.12? 1.13?
because it's only part of 1.13 release

regisss:
Yes 1.13, I'm using this Docker image: vault.habana.ai/gaudi-docker/1.13.0/ubuntu20.04/habanalabs/pytorch-installer-2.1.0:latest

regisss:
And
```
HL-SMI Version:                           hl-1.13.0-rc-fw-47.0.0.0
Driver Version:                                     1.13.0-ee32e42
```
from `hl-smi`

bgoldberg-habana:
> is_supported_type

running on gaudi1 or 2?

regisss:
Ah that must be this! Gaudi1 (DL1 instance on AWS), my Gaudi2 instance won't get 1.13 before next week...
fp8 is only supported on Gaudi2 right?

regisss:
Okay let's merge then, could you just let me know the throughput you get with the example command line so that I can compare with my runs next week please?

regisss (approved):

bgoldberg-habana:
i'll send you all the numbers i  have through @libinta

regisss:
Thanks!

regisss:
@bgoldberg-habana When I try the command of the README with Gaudi2 on 1.13, I get the following error:
```
  File "run_generation.py", line 490, in <module>
    main()
  File "run_generation.py", line 294, in main
    generate()
  File "run_generation.py", line 275, in generate
    outputs = model.generate(
  File "/usr/local/lib/python3.8/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/generation/utils.py", line 800, in generate
    return self.greedy_search(
  File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/generation/utils.py", line 1371, in greedy_search
    model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
  File "/usr/local/lib/python3.8/dist-packages/optimum/habana/transformers/models/llama/modeling_llama.py", line 584, in prepare_inputs_for_generation
    input_ids = input_ids[:, :token_idx]
RuntimeError: Graph compile failed. synStatus=synStatus 26 [Generice failure].
```

Does it work on your side?

regisss:
@bgoldberg-habana It's actually just a `\` missing at the end of the line where the env variables are declared. Fixed in #586.

PR #543, opened 2023-11-20
Sarkar/dyn prompt afterrefactor

ssarkar2:
# What does this PR do?
1. Fix issue when prompt_len is a multiple of bucket
2. add `--simulate_dynamic_prompt`, which lets us simulate dynamic input. `adjust_batch` function adjusts the batch lengths based on inputs from `--simulate_dynamic_prompt`
3. Add appropriate warmup for dynamic inputs based on min and max of `--simulate_dynamic_prompt`
4. add `--reduce_recompilation` for extra optimizations for time-to-first-token in dynamic case. This does the following optimizations:
  - Preprocess on CPU (pad on CPU)
  - compute position_idx on CPU
  

Fixes # (issue)

## Before submitting


ssarkar2:
moving https://github.com/huggingface/optimum-habana/pull/525 to after-refactor code

examples/text-generation/run_generation.py
```diff
@@ -205,6 +206,17 @@ def setup_parser(parser):
         help="Whether to skip hash with views for HPU graphs. When skip_hash_with_views is not used, the input to HPU graphs includes both view and base tensors.",
     )
     parser.add_argument("--verbose_workers", action="store_true", help="Enable output from non-master workers")
+    parser.add_argument(
+        "--simulate_dyn_prompt",
+        default="",
+        type=str,
+        help="If empty static prompt is used. If a comma separated list of integers are passed, we warmup and use those shapes for prompt length",
```
regisss, line 213:
```suggestion
        help="If empty, static prompt is used. If a comma separated list of integers is passed, we warmup and use those shapes for prompt length.",
```

regisss, line 213: I think it should be defined similarly to the `--prompt` arg with default None and type int: https://github.com/huggingface/optimum-habana/blob/4ffb9d7d495347f4aef0d66eecca06bcef4c2b11/examples/text-generation/run_generation.py#L125-L131
It will also return a list and looks cleaner IMO.

examples/text-generation/run_generation.py
```diff
@@ -288,19 +303,49 @@ def generate():
         HabanaProfile.disable()
         # Compilation
         logger.info("Graph compilation...")
+        if len(args.simulate_dyn_prompt) > 0:
+            dyn_prompt_lens = [int(k) for k in args.simulate_dyn_prompt.split(",")]
+        else:
+            dyn_prompt_lens = None
```
regisss, line 309: This wouldn't be needed with the suggestion above for defining the simulate_dyn_prompt arg.

regisss:
@ssarkar2 Can you also rebase your branch on main so that the CI passes please?

ssarkar2:
@regisss : incorporated the suggestions and merged main

optimum/habana/transformers/generation/utils.py
```diff
@@ -573,6 +582,11 @@ def generate(
             or self._get_generation_mode(generation_config, assistant_model) == GenerationMode.BEAM_SEARCH
         )
         model_kwargs["bucket_size"] = generation_config.bucket_size if generation_config.static_shapes else -1
+        model_kwargs["reduce_recompile"] = (generation_config.reduce_recompile, False)[
+            generation_config.reduce_recompile is None
+        ]
```
regisss, line 587:
```suggestion
        model_kwargs["reduce_recompile"] = generation_config.reduce_recompile if generation_config.reduce_recompile is not None else False
```
I think it's easier to understand that way. You'll probably have to run `make style` too because this line is probably too long.

ssarkar2:
@regisss , incorporated review comments

examples/text-generation/README.md
```diff
@@ -215,6 +215,7 @@ python run_generation.py \
 
 `--bucket_size` option is especially useful when processing an input stream with varying lengths, that is when you have something like `--dataset_name squad --column_name context --max_input_tokens -1`. `--max_input_tokens -1` specifies no truncation of input prompt in the dataset.
 
+Another way to simulate dynamic input is to use `--simulate_dyn_prompt`. For example `--simulate_dyn_prompt 25,35,45` will extend or crop the default prompt (or the prompt passed in using `--prompt`) to sizes 25, 35, and 45, and throughput will be measured for these 3 lengths. If `--simulate_dyn_prompt` is used, the min and max input lengths from it are computed to perform warmup as well. One final optimization that can be used in case of dynamic inputs is `--reduce_recompile`. Thus the suggested configuration to simulate dynamicity after warmup is to use all three arguments: `--simulate_dyn_prompt 25,35,45 --reduce_recompile --bucket 30`
```
regisss, line 218:
```suggestion
Another way to simulate dynamic input is to use `--simulate_dyn_prompt`. For example `--simulate_dyn_prompt 25,35,45` will extend or crop the default prompt (or the prompt passed in using `--prompt`) to sizes 25, 35, and 45, and throughput will be measured for these 3 lengths. If `--simulate_dyn_prompt` is used, the min and max input lengths from it are computed to perform warmup as well. One final optimization that can be used in case of dynamic inputs is `--reduce_recompile`. Thus the suggested configuration to simulate dynamicity after warmup is to use all three arguments: `--simulate_dyn_prompt 25,35,45 --reduce_recompile --bucket_size 30`
```
Trying to run an example, shouldn't it be `--bucket_size` in the README?

ssarkar2: Since "bucket" uniquely identifies 'bucket_size' without ambiguity, it is enough. Ie, if the prefix uniquely identifies the option, it is enough. But yes, probably clearer to have teh actual full name

ssarkar2:
@regisss  Added readme change

ssarkar2:
@regisss , can we merge this?

libinta (approved):

examples/text-generation/run_generation.py
```diff
@@ -293,19 +309,46 @@ def generate():
...
+                def rounder(x):
+                    return int(math.ceil(x / args.bucket_size) * args.bucket_size)
+
+                min_prompt_len = rounder(mn)
+                max_sentence_len = rounder(mx)
+                for _ in range(args.warmup):
+                    lst = list(range(min_prompt_len, max_sentence_len + 1, args.bucket_size))
+                    for sz in lst:
+                        print("Warming up for shape,", sz - 1, flush=True)
+                        generate(sz - 1, args.reduce_recompile)
```
regisss, line 337: Why `sz - 1` and not `sz`? It should a multiple of `bucket_size` no?

examples/text-generation/run_generation.py
```diff
@@ -293,19 +309,46 @@ def generate():
...
         t0 = time.perf_counter()
         # Benchmark over n_iterations iterations
-        for i in range(args.n_iterations):
-            generated = generate()
+        if dyn_prompt_lens is None:
+            for i in range(args.n_iterations):
+                generated = generate(None, args.reduce_recompile)
+        else:
+            for i in range(args.n_iterations):
+                print("Generating for shape,", dyn_prompt_lens[i])
```
regisss, line 350: This will raise an error if `args.n_iterations` is larger than the length of `dyn_prompt_lens`

examples/text-generation/README.md
```diff
@@ -215,6 +215,7 @@ python run_generation.py \
 
 `--bucket_size` option is especially useful when processing an input stream with varying lengths, that is when you have something like `--dataset_name squad --column_name context --max_input_tokens -1`. `--max_input_tokens -1` specifies no truncation of input prompt in the dataset.
 
+Another way to simulate dynamic input is to use `--simulate_dyn_prompt`. For example `--simulate_dyn_prompt 25,35,45` will extend or crop the default prompt (or the prompt passed in using `--prompt`) to sizes 25, 35, and 45, and throughput will be measured for these 3 lengths. If `--simulate_dyn_prompt` is used, the min and max input lengths from it are computed to perform warmup as well. One final optimization that can be used in case of dynamic inputs is `--reduce_recompile`. Thus the suggested configuration to simulate dynamicity after warmup is to use all three arguments: `--simulate_dyn_prompt 25,35,45 --reduce_recompile --bucket_size 30`
```
regisss, line 218:
```suggestion
Another way to simulate dynamic input is to use `--simulate_dyn_prompt`. For example `--simulate_dyn_prompt 25,35,45` will extend or crop the default prompt (or the prompt passed in using `--prompt`) to sizes 25, 35, and 45, and throughput will be measured for these 3 lengths. If `--simulate_dyn_prompt` is used, the min and max input lengths from it are computed to perform warmup as well. One final optimization that can be used in case of dynamic inputs is `--reduce_recompile`. Thus the suggested configuration to simulate dynamicity after warmup is to use all three arguments: `--simulate_dyn_prompt 25 35 45 --reduce_recompile --bucket_size 30`
```

regisss (commented):
I spotted a bug running
```
python run_generation.py --model_name_or_path gpt2 --use_hpu_graphs --use_kv_cache --max_new_tokens 100 --simulate_dyn_prompt 25 35 45 --reduce_recompile --bucket_size 30
```

examples/text-generation/run_generation.py
ssarkar2: this is fine. for bucket_size>1 and most values of sz, sz-1 and sz will compile to the same size (some multiple of bucket_size)

There is some reason I forget exactly why, but its better to have sz-1 than sz.

It doesnt need to be multiple of bucket here, it will be rounded up inside

examples/text-generation/run_generation.py
ssarkar2: fixed

ssarkar2:
@regisss , fixed, please take a look

regisss (approved):

PR #545, opened 2023-11-20
Fix run generation and translation CI

yeonsily:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/translation/run_translation.py
+++ b/examples/translation/run_translation.py
@@ -400,8 +400,12 @@ def main():
         if data_args.test_file is not None:
             data_files["test"] = data_args.test_file
             extension = data_args.test_file.split(".")[-1]
+        if extension == "jsonl":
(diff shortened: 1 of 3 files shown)


yeonsily:
@regisss I see that we have more code change after https://github.com/huggingface/optimum-habana/pull/539 merged in.
Also we have difference in run_translation.py.

regisss:
@yeonsily Yep, I forgot to update the diff file for the text generation example.
For the translation one, I just pushed the new changes from Transformers.
Usually, when this test fails, it means that one example got changed [here](https://github.com/huggingface/transformers/tree/main/examples/pytorch) in Transformers so we should add it to our examples (if compatible). Only the text-generation example is very different from the Transformers one so usually I directly update the diff file without adding any change as they are often unrelated.

regisss (approved):

PR #549, opened 2023-11-21
Update trainer.py

skaulintel:
Fix ds scheduler_dict error in CI

"UnboundLocalError: local variable 'scheduler_dict' referenced before assignment"

# What does this PR do?

Fixes # (issue)

## Before submitting


skaulintel:
@regisss noticed scheduler_dict isn't defined in case of deepspeed. The above fixes the issue, but please let me know if there is a better method of getting passed the error.

regisss (approved):
LGTM! Your fix is what is done in Transformers: https://github.com/huggingface/transformers/blob/acc394c4f5e1283c19783581790b3dc3105a3697/src/transformers/trainer.py#L2448
Not sure why I put something else, thanks for fixing it :slightly_smiling_face:

PR #570, opened 2023-11-30
sarkar/fix_diffuser_tests

ssarkar2:
# What does this PR do?

Fix diffuser tests by updating expected values. Images have been checked visually, they look correct

Fixes some tests in: `RUN_SLOW=1 GAUDI2_CI=1 python -m pytest tests/test_diffusers.py -v -s -k "test_no_generation_regression"`

Fixes # (issue)

## Before submitting


ssarkar2:
> The problem with this test is that results are not reproducible on Gaudi2 and it fails every now and then

yes, i do see that it produces slightly different images each time. however for this test I notice that the numbers have moved significantly.. from 0.350823, to 0.68306947,... so updating teh "expected"

With this expected I ran the test a few times and it passes.

If this test fails again, maybe we can increase the tolearnce a bit.

The other test in this file test_no_generation_regression_ldm3d has even more variance of output, so not trying to fix that (though i inspected the result visually and it looks fine)

regisss:
Okay sounds good! I don't know exactly why I got so different results before, that's really weird.

Can you rebase your branch on main so that the CI passes please?

ssarkar2:
@regisss , CI passes now

PR #575, opened 2023-12-02
remove redundant requirements

hsubramony:
# What does this PR do?

Fixes # (issue)

## Before submitting


hsubramony:
i removed redundant reqs from pyproject.toml since all these were already part of requirements.txt
any reason we need both ? cant we just keep requirements or toml file ?

regisss (approved):

regisss:
Arf I read this PR too fast, I shouldn't have merged it. We need the pyproject.toml to specify dependencies for Poetry, then the requirements.txt is generated autoamtically.
Anyway, we are moving to a dedicated fork today or tomorrow so this won't have any impact.

PR #576, opened 2023-12-04
In peft, only the trainable parameters need to be saved

sywangyi:
to reduce the storage size and also save the time of checkpoint saving while using deepspeed for training

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -15,6 +15,7 @@
 
 import contextlib
 import copy
+import inspect
 import math
 import os
 import random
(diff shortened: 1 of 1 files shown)


sywangyi:
https://github.com/huggingface/transformers/pull/27825

regisss:
It looks good to me! Let's wait for it to be merged into Transformers before merging here in case some modifications are needed.

libinta (commented):
@sywangyi  how do you validate the saved checkpoint?

sywangyi:
I usually track the wandb log, comparing with the result from A100. DPO is simple, For PPO, I do find some issue and fix it in TRL for habana

regisss (approved):

PR #583, opened 2023-12-05
Enable habana flash attention for llama2 ft

mandy-li:
Add Habana flash attention support to llama2-70b finetuning


mandy-li:
Drop this PR and created a separate one

PR #594, opened 2023-12-12
fix gptj training issue

BaihuiJin:
# What does this PR do?

Fix training runtime error caused by lack of sin and cos value missing when load from checkpoint.

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/gptj/modeling_gptj.py
+++ b/optimum/habana/transformers/models/gptj/modeling_gptj.py
@@ -390,7 +390,7 @@ def gaudi_gptj_model_forward(
             def create_custom_forward(module):
                 def custom_forward(*inputs):
                     # None for past_key_value
-                    return module(*inputs, use_cache, output_attentions, token_idx, sin, cos)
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/gptj/modeling_gptj.py
```diff
@@ -390,7 +390,7 @@ def gaudi_gptj_model_forward(
             def create_custom_forward(module):
                 def custom_forward(*inputs):
                     # None for past_key_value
-                    return module(*inputs, use_cache, output_attentions)
+                    return module(*inputs, use_cache, output_attentions, token_idx, sin, cos)
```
regisss, line 393: `token_idx` should be False when using gradient checkpointing right?

BaihuiJin: It's `None` actually

BaihuiJin: fixed to `None`

regisss (approved):
LGTM

PR #605, opened 2023-12-18
Run Llama2 with torch.compile on Gaudi2

kausikmaiti:
# What does this PR do?

This change allows the user to run Llama2 model with torch.compile on Gaudi2.

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -74,6 +74,11 @@ def setup_parser(parser):
         help="Whether to use the key/value cache for decoding. It should speed up generation.",
     )
     parser.add_argument(
+        "--use_torch_compile",
+        action="store_true",
+        help="Whether to use torch compiled model or not.",
+    )
+    parser.add_argument(
         "--use_hpu_graphs",
         action="store_true",
         help="Whether to use HPU graphs or not. Using HPU graphs should give better latencies.",
@@ -234,6 +239,9 @@ def setup_parser(parser):
 
     args = parser.parse_args()
 
+    if args.use_torch_compile:
+        args.use_hpu_graphs = False
+
     if not args.use_hpu_graphs:
         args.limit_hpu_graphs = False
 
@@ -297,7 +305,8 @@ def main():
             outputs = model.generate(
                 **input_tokens,
                 generation_config=generation_config,
-                lazy_mode=True,
+                lazy_mode=True if not args.use_torch_compile else False,
+                torch_compile = args.use_torch_compile,
                 hpu_graphs=args.use_hpu_graphs,
                 profiling_steps=args.profiling_steps,
                 profiling_warmup_steps=args.profiling_warmup_steps,
@@ -477,7 +486,8 @@ def main():
             outputs = model.generate(
                 **batch,
                 generation_config=generation_config,
-                lazy_mode=True,
+                lazy_mode=True if not args.use_torch_compile else False,
+                torch_compile = args.use_torch_compile,
                 hpu_graphs=args.use_hpu_graphs,
                 profiling_steps=args.profiling_steps,
                 profiling_warmup_steps=args.profiling_warmup_steps,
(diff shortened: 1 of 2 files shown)


optimum/habana/transformers/generation/utils.py
```diff
@@ -406,6 +406,7 @@ def generate(
         negative_prompt_ids: Optional[torch.Tensor] = None,
         negative_prompt_attention_mask: Optional[torch.Tensor] = None,
         lazy_mode: Optional[bool] = False,
+        torch_compile: Optional[bool] = False,
```
vivekgoe, line 409: For normal training, eval, predict models are wrapped within accelerator.prepare_model() call, adding new code for generate() may not be aligned. @regisss any idea how direct model.generate() calls are handled in transformers for compile mode, I tried to search there but did not find anything.

vivekgoe (commented):
@regisss can you please help review this change? this is needed to enable torch.compile for text generation tasks (for Llama and other models).

examples/text-generation/run_generation.py
```diff
@@ -73,6 +73,11 @@ def setup_parser(parser):
         action="store_true",
         help="Whether to use the key/value cache for decoding. It should speed up generation.",
     )
+    parser.add_argument(
+        "--use_torch_compile",
```
regisss, line 77:
```suggestion
        "--torch_compile",
```
to be aligned with Transformers and GaudiTrainingArguments

optimum/habana/transformers/generation/utils.py
regisss: In the trainer, the link with Accelerate is made here: https://github.com/huggingface/optimum-habana/blob/21238af3e2e4eff84d1db3315d977004c49b45b4/optimum/habana/transformers/training_args.py#L474
And then in Accelerate it happens here: https://github.com/huggingface/optimum-habana/blob/21238af3e2e4eff84d1db3315d977004c49b45b4/optimum/habana/accelerate/accelerator.py#L371
It was introduced in https://github.com/huggingface/optimum-habana/pull/465.

regisss: Outside of the trainer, Transformers recommends to simply use:
```
model = torch.compile(model)
```
https://huggingface.co/docs/transformers/v4.36.1/en/perf_torch_compile

optimum/habana/transformers/generation/utils.py
```diff
@@ -1403,14 +1414,26 @@ def greedy_search(
...
 
-            # forward pass to get next token
-            outputs = self(
-                **model_inputs,
-                return_dict=True,
-                output_attentions=output_attentions,
-                output_hidden_states=output_hidden_states,
-                **hpu_graphs_kwargs,
-            )
+            if torch_compile:
```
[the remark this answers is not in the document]
regisss: Not even sure we should do it in `generate` at all. If using the trainer, it should already be taken care of (see discussion above). Otherwise, for example in the text-generation example, I think we should just have a `get_torch_compiled_model` in `text-generation/utils.py`. That seems to be the way recommended by Transformers.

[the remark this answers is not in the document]
vivekgoe: @regisss thanks for your comments, we will check if we can go with adding get_torch_compiled_model in text-generation/utils.py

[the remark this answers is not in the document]
kausikmaiti: Ok. I would create 'get_torch_compiled_model' in text-generation/utils.py.

examples/text-generation/run_generation.py
kausikmaiti: ok. I would change.

optimum/habana/transformers/generation/utils.py
kausikmaiti: As suggested, I would create 'get_torch_compiled_model()' in text-generation/utils.py. And this will be called inside setup_model() in text-generation/utils.py.

kausikmaiti:
I created a separate PR after making necessary changes. Kindly refer to https://github.com/huggingface/optimum-habana/pull/616

PR #609, opened 2023-12-21
Update generation config to enable flash attention for inference

puneeshkhanna:
# What does this PR do?

Fixes # (issue)

## Before submitting


puneeshkhanna:
@regisss  - please review and merge. 

On another note, I suspect that when flash attention is enabled then inference o/p of llama is not correct but that would be a separate issue which I will raise.
But right now if the example script passes flash attention arg then it is not even passed to the model code correctly. So one may think that flash attention is used but it is not actually getting used. So this PR fixes that.

regisss (approved):
LGTM, thanks for catching this!

PR #610, opened 2023-12-21
Adding support for bf16_full_eval

bhargaveede:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -428,6 +428,11 @@ class GaudiTrainer(Trainer):
 
         self.is_in_train = True
 
+        # do_train is not a reliable argument, as it might not be set and .train() still called, so
+        # the following is a workaround:
+        if (args.fp16_full_eval or args.bf16_full_eval) and not args.do_train:
+            self._move_model_to_device(self.model, args.device)
+
         if "model_path" in kwargs:
             resume_from_checkpoint = kwargs.pop("model_path")
             warnings.warn(
@@ -1510,6 +1515,12 @@ class GaudiTrainer(Trainer):
                 )
                 self.already_wrapped_for_hpu_graphs = True
 
+        # if full fp16 or bf16 eval is wanted and this ``evaluation`` or ``predict`` isn't called
+        # while ``train`` is running, cast it to the right dtype first and then put on device
+        if not self.is_in_train:
+            if args.bf16_full_eval:
+                model = model.to(dtype=torch.bfloat16, device=args.device)
+
         batch_size = self.args.eval_batch_size
 
         logger.info(f"***** Running {description} *****")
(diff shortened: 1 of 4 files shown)


bhargaveede:
@regisss I have enabled bf16_full_eval and verified the same with t5 inference.
Can you review it?

optimum/habana/transformers/trainer.py
```diff
@@ -1510,6 +1515,12 @@ def evaluation_loop(
                 )
                 self.already_wrapped_for_hpu_graphs = True
 
+        # if full fp16 or bf16 eval is wanted and this ``evaluation`` or ``predict`` isn't called
+        # while ``train`` is running, cast it to the right dtype first and then put on device
+        if not self.is_in_train:
+            if args.bf16_full_eval:
+                model = model.to(dtype=torch.bfloat16, device=args.device)
+
```
vivekgoe, line 1523: maybe we can add code for args.fp16_full_eval also (for future if we decide to support fp16 also).

optimum/habana/transformers/trainer.py
```diff
@@ -1903,6 +1914,12 @@ def prediction_loop(
...
```
vivekgoe, line 1922: maybe we can add code for args.fp16_full_eval also (for future if we decide to support fp16 also).

vivekgoe:
@bhargaveede please also update (1) summarization README file to include this extra argument for t5-3b prediction example, (2) performance targets in test for both G1 and G2.

bhargaveede:
@regisss  Can your review and merge this?

bhargaveede:
@regisss Can we merge this?

vivekgoe (approved):

vivekgoe:
Looks good to me. We can merge it once @regisss also reviews.

regisss:
I'll review it this week :+1:

tests/test_encoder_decoder_text_summarization.py
```diff
@@ -15,15 +15,15 @@
...
+            ("t5-3b", "Habana/t5", 2.88, 21.56, 2, 1),
         ],
     }
 else:
     # Gaudi1 CI baselines
     MODELS_TO_TEST = {
         "bf16": [
             ("facebook/bart-large-cnn", "Habana/bart", 2.588, 26.0688, 2, 1),
-            ("t5-3b", "Habana/t5", 0.585, 21.72, 2, 1),
+            ("t5-3b", "Habana/t5", 0.98, 21.56, 2, 1),
```
regisss, line 26: For Gaudi1, I get a RougeLsum of 21.3831 and a throughput of 1.005. it doesn't matter much since the test passes.
For Gaudi2 however, runs are not reproducible it seems. I get different RougeLsum from one run to another, is it something you also observed?

bhargaveede: I didn't get different RougeLsum. When I added perf numbers, I ran the test twice to check and I was getting same RogueLsum. Let me check again.

regisss: Interesting, did you run it with Synapse 1.13?

bhargaveede: I could see the variation. However, I'm seeing variation on v1.9-release too for the test "test_run_summarization_t5-small_multi_card". Can you confirm if it's same on your end

regisss: I cannot run multi-card tests on my Gaudi2 instance at the moment but if you observed the same behavior for "test_run_summarization_t5-small_multi_card" it means that this "issue" was already there before.
Anyway, tests still pass so I'm going to merge it and I'll investigate that later.

regisss (approved):

PR #611, opened 2023-12-21
Add inheritance in Diffusers pipelines

regisss:
# What does this PR do?

Make Gaudi pipleines inherit from Diffusers pipelines so that we have to maintain less code. We have to be careful here because of multiple inheritance.

## Before submitting


libinta (commented):
can you add the diffuser link that include the version in the pipeline_stable_diffusionxxx?

libinta (approved):

PR #616, opened 2023-12-28
Run Llama2 with torch.compile on Gaudi2

kausikmaiti:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -254,7 +254,7 @@ def main():
     model, tokenizer, generation_config = initialize_model(args, logger)
 
     use_lazy_mode = True
-    if args.torch_compile:
+    if args.torch_compile and model.config.model_type == "llama":
         use_lazy_mode = False
 
     import habana_frameworks.torch.hpu as torch_hpu
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -231,8 +231,16 @@ def setup_parser(parser):
         action="store_true",
         help="Whether to enable Habana Flash Attention, provided that the model supports it.",
     )
+    parser.add_argument(
+        "--torch_compile",
+        action="store_true",
+        help="Whether to use torch compiled model or not.",
+    )
 
     args = parser.parse_args()
+
+    if args.torch_compile:
+        args.use_hpu_graphs = False
 
     if not args.use_hpu_graphs:
         args.limit_hpu_graphs = False
@@ -244,6 +252,10 @@ def main():
     parser = argparse.ArgumentParser()
     args = setup_parser(parser)
     model, tokenizer, generation_config = initialize_model(args, logger)
+
+    use_lazy_mode = True
+    if args.torch_compile:
+        use_lazy_mode = False
 
     import habana_frameworks.torch.hpu as torch_hpu
 
@@ -297,7 +309,7 @@ def main():
             outputs = model.generate(
                 **input_tokens,
                 generation_config=generation_config,
-                lazy_mode=True,
+                lazy_mode=use_lazy_mode,
                 hpu_graphs=args.use_hpu_graphs,
                 profiling_steps=args.profiling_steps,
                 profiling_warmup_steps=args.profiling_warmup_steps,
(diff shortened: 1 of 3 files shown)


vivekgoe:
@kausikmaiti looks good to me. @regisss please review and help merge this if it looks ok to you. Thanks.

vivekgoe (approved):

libinta (commented):
@kausikmaiti I have following questions
1. what's the default cmd for llama2? torch compile or use_lazy_mode? if the default is torch.compile, please also update the readme
2. what's the performance comparison with use_lazy_mode and use_hpu_graph?
3. any depenency to the to-be-released docker?

examples/text-generation/utils.py
```diff
@@ -151,6 +151,11 @@ def patch_scoped_linear_all_reduce(model):
         patch_scoped_linear_all_reduce(module)
 
 
+def get_torch_compiled_model(model):
+    model.model = torch.compile(model.model, backend="aot_hpu_inference_backend")
```
regisss, line 155: Is "model" a new attribute of `model` or does it already exist?

kausikmaiti:
> @kausikmaiti I have following questions
> 
> 1. what's the default cmd for llama2? torch compile or use_lazy_mode? if the default is torch.compile, please also update the readme
> 2. what's the performance comparison with use_lazy_mode and use_hpu_graph?
> 3. any depenency to the to-be-released docker?

@libinta, Please find my answers below.
1. Here, I am just adding support for torch.compile. But, default mode is still unchanged. So, we don't need to update readme here.
2. This is the very first step of (functional) enablement. Performance data is not yet available.
3. Yes. It has dependency on the content of release 1.14.

examples/text-generation/utils.py
kausikmaiti: @regisss, It already exists.

examples/text-generation/run_generation.py
```diff
@@ -245,6 +253,10 @@ def main():
     args = setup_parser(parser)
     model, tokenizer, generation_config = initialize_model(args, logger)
 
+    use_lazy_mode = True
+    if args.torch_compile:
```
bhargaveede, line 257: We could avoid this by using args.torch_compile directly in 312 and 492 with lazy_mode = not args.torch_compile.

regisss:
> Yes. It has dependency on the content of release 1.14.

So we should not merge it before 1.14 is released @libinta right?

examples/text-generation/run_generation.py
kausikmaiti: Yes, that can be done of course. But, I wanted to have the decision making about use_lazy_mode in one place for all. A secondary reason is that I plan to include additional checks in future PR to control use_lazy_mode.

examples/text-generation/utils.py
```diff
@@ -170,6 +175,10 @@ def setup_model(args, model_dtype, model_kwargs, logger):
             model = wrap_in_hpu_graph(model, hash_with_views=not args.skip_hash_with_views)
         else:
             model = wrap_in_hpu_graph(model)
+
+    if args.torch_compile:
+        model = get_torch_compiled_model(model)
```
bhargaveede, line 180: @kausikmaiti  Can we add model specific check as generation using torch.compile isn't verified on models

kausikmaiti: Added model specific check in separate commit. Please review.

regisss:
@kausikmaiti Let's also add a test to: https://github.com/huggingface/optimum-habana/blob/main/tests/test_text_generation_example.py

You can define a new test at the end of this file:
```py
@pytest.mark.parametrize("model_name, baseline", MODELS_TO_TEST["torch_compile"])
def test_text_generation_torch_compile(model_name: str, baseline: float, token: str):
    _test_text_generation(model_name, baseline, token, torch_compile=True)
```
and adding a new `torch_compile` entry in the dict here: https://github.com/huggingface/optimum-habana/blob/979c13247027fa9c8b9cba5137c4a17e27615589/tests/test_text_generation_example.py#L15

tests/test_text_generation_example.py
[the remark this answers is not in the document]
kausikmaiti: As you mentioned offline, WORLD_SIZE setting does not matter, as I'm not using deepspeed / gaudi_spawn.py script.
Also as per my observation, if I don't set WORLD_SIZE=0, due to the logic like "use_deepspeed = args.world_size > 0", setup_distributed_model() gets called and the test fails at very early stage while importing deepspeed. This is not the expectation.

regisss (approved):

PR #620, opened 2024-01-02
Avoid falcon perf drop from PR#607 when BS=1

schoi-habana:
#607 fixed the graph compilation error for BS>1 but it degrades the perf when BS=1.
So an additional condition was added to apply the change in PR#607 only when BS>1

## Before submitting


mandy-li (approved):
LGTM

regisss (approved):

PR #626, opened 2024-01-08
Flash attention enhancement of repeatKV

puneeshkhanna:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/llama/modeling_llama.py
+++ b/optimum/habana/transformers/models/llama/modeling_llama.py
@@ -228,9 +228,6 @@ class GaudiLlamaAttention(LlamaAttention):
         else:
             past_key_value = None
 
-        key_states = gaudi_llama_repeat_kv(key_states, self.num_key_value_groups)
-        value_states = gaudi_llama_repeat_kv(value_states, self.num_key_value_groups)
-
         if use_flash_attention and FusedSDPA:
             import habana_frameworks.torch.hpu as ht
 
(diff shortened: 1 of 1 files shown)


puneeshkhanna:
@regisss  - Don't merge this yet. Let me add some testing results too in this week. 

Fused SDPA will internally handle repeat KV logic in 1.14 release hence we don't need repeat KV logic when flash attention is enabled.  We should see perf boost for some configs such as TP-4.

Can you please add tag of synapse 1.14 to this PR mean while.

dvarshney-habana (approved):

regisss (approved):
Very nice speedups!

So we have to wait for the release of SynapseAI v1.14 to merge this one right?

puneeshkhanna:
Yes @regisss  - I think we l need to wait for it.

schoi-habana:
@puneeshkhanna we tested the PR with a 4x finetuning, but no performance improvement was seen. is the perf improvement limited to inference, or did I miss something?

libinta (approved):

puneeshkhanna:
@schoi-habana  - Just to double check that you did use "flash attention" for finetuning too right  and also finetuning 70B model right ? I m not sure how flash attention flag is passed for finetuning. Perf improvement surely for inference of 70B model. Change is not applicable for 7b and 13 b models as they don't use GQA.  We should also see some improvement in memory usage too.

schoi-habana:
@puneeshkhanna yes i tested llama2-70b with --use_flash_attention. the sec/iter improvement is about 2-3% and the mem usage didn't improve.

puneeshkhanna:
Thanks for the update @schoi-habana . 2-3% improvement in finetuning is good too I guess. We should be good to merge this with 1.14 synapse release. I will try to fix the non-flash attention path too in next week in separate PR.

puneeshkhanna:
We should first merge this PR and then https://github.com/huggingface/optimum-habana/pull/639 to avoid any merge conflicts hopefully.

PR #627, opened 2024-01-08
fix beamsearch crash and incorrect output in decode-only model and en…

sywangyi:
…code-decode model

found with "do_sample=False, num_beams=4" in generation. test model include
decode only:"gpt2" "tiiuae/falcon-7b-instruct" "distilgpt2"
encode decoder: "facebook/bart-large-cnn" "sshleifer/distilbart-cnn-12-6" "philschmid/bart-large-cnn-samsum"

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -27,9 +27,6 @@ from transformers.generation.beam_search import BeamScorer, BeamSearchScorer, Co
 from transformers.generation.candidate_generator import CandidateGenerator
 from transformers.generation.logits_process import LogitsProcessorList
 from transformers.generation.stopping_criteria import (
-    MaxLengthCriteria,
-    MaxTimeCriteria,
-    StoppingCriteria,
     StoppingCriteriaList,
     validate_stopping_criteria,
 )
@@ -103,16 +100,6 @@ def incrementor(bucket_size, prompt_len):
             "token_idx": token_idx,
             "need_expansion": need_expansion,
         }
-
-
-class StaticMaxLengthCriteria(StoppingCriteria):
-    def __init__(self, max_steps: int):
-        self.max_steps = max_steps
-        self.cur_step = 0
-
-    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs) -> bool:
-        self.cur_step += 1
-        return self.cur_step >= self.max_steps
 
 
 class GaudiGenerationMixin(GenerationMixin):
@@ -307,33 +294,6 @@ class GaudiGenerationMixin(GenerationMixin):
                 )
                 model_kwargs["decoder_attention_mask"] = decoder_attention_mask
         return decoder_input_ids, model_kwargs
-
-    def _get_stopping_criteria(
-        self, generation_config: GaudiGenerationConfig, stopping_criteria: Optional[StoppingCriteriaList]
-    ) -> StoppingCriteriaList:
-        criteria = StoppingCriteriaList()
-        if generation_config.max_length is not None:
-            if (
-                generation_config.static_shapes
-                and self.config.is_encoder_decoder
-                and (
-                    self.generation_config.generation_mode == GenerationMode.GREEDY_SEARCH
-                    or self.generation_config.generation_mode == GenerationMode.BEAM_SEARCH
-                )
-            ):
-                criteria.append(StaticMaxLengthCriteria(generation_config.max_length))
-            else:
-                max_position_embeddings = getattr(self.config, "max_position_embeddings", None)
-                criteria.append(
-                    MaxLengthCriteria(
-                        max_length=generation_config.max_length,
-                        max_position_embeddings=max_position_embeddings,
-                    )
-                )
-        if generation_config.max_time is not None:
-            criteria.append(MaxTimeCriteria(max_time=generation_config.max_time))
-        criteria = self._merge_criteria_processor_list(criteria, stopping_criteria)
-        return criteria
 
     @torch.no_grad()
     def update_model_kwargs_for_bucketing(
@@ -692,7 +652,11 @@ class GaudiGenerationMixin(GenerationMixin):
             else:
                 generation_config.max_length = generation_config.max_new_tokens + input_ids_length
 
-        self._validate_generated_length(generation_config, input_ids_length, has_default_max_length)
+        self._validate_generated_length(
+            generation_config,
+            model_kwargs["token_idx"] if "token_idx" in model_kwargs else input_ids_length,
+            has_default_max_length,
+        )
 
         # determine whether introduce trim_logits feature
         model_kwargs["trim_logits"] = generation_config.trim_logits
@@ -772,23 +736,6 @@ class GaudiGenerationMixin(GenerationMixin):
         prepared_stopping_criteria = self._get_stopping_criteria(
             generation_config=generation_config, stopping_criteria=stopping_criteria
         )
-        if "token_idx" in model_kwargs and not self.config.is_encoder_decoder:
-            if generation_config.max_new_tokens is not None:
-                prepared_stopping_criteria.append(StaticMaxLengthCriteria(generation_config.max_new_tokens))
-            else:
-                raise ValueError(
-                    "You need to set `max_new_tokens` in your generation configuration to use static shapes."
-                )
-
-        if generation_config.static_shapes and generation_config.bucket_size > 0:
-            prepared_stopping_criteria = StoppingCriteriaList(
-                [
-                    StaticMaxLengthCriteria(generation_config.max_new_tokens)
-                    if type(crit) == MaxLengthCriteria
-                    else crit
-                    for crit in prepared_stopping_criteria
-                ]
-            )
 
         # In lazy mode, import Habana torch to be able to add mark_step()
         if lazy_mode:
@@ -1491,7 +1438,7 @@ class GaudiGenerationMixin(GenerationMixin):
                     this_peer_finished = True
 
             # stop if we exceed the maximum length
-            if stopping_criteria(input_ids, scores):
+            if stopping_criteria(input_ids, scores, token_idx=token_idx):
                 this_peer_finished = True
 
             hb_profer.step()
@@ -1811,7 +1758,7 @@ class GaudiGenerationMixin(GenerationMixin):
                     this_peer_finished = True
 
             # stop if we exceed the maximum length
-            if stopping_criteria(input_ids, scores):
+            if stopping_criteria(input_ids, scores, token_idx=token_idx):
                 this_peer_finished = True
 
             hb_profer.step()
@@ -2309,7 +2256,7 @@ class GaudiGenerationMixin(GenerationMixin):
                     and num_eos_tokens >= num_beams_tensor
                 ):
                     break
-                elif stopping_criteria(input_ids[:, :token_idx], scores):
+                elif stopping_criteria(input_ids, scores, token_idx=cur_len):
                     break
             elif stopping_criteria(input_ids, scores) or (beam_scorer.is_done and not lazy_mode):
                 if not synced_gpus:
@@ -2991,7 +2938,8 @@ class GaudiGenerationMixin(GenerationMixin):
             cur_len = cur_len + 1
 
             hb_profer.step()
-            if constrained_beam_scorer.is_done or stopping_criteria(input_ids, scores):
+
+            if constrained_beam_scorer.is_done or stopping_criteria(input_ids, scores, token_idx=token_idx):
                 if not synced_gpus:
                     break
                 else:
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -1332,7 +1332,11 @@ class GaudiGenerationMixin(GenerationMixin):
             inc = iter(incrementor(bucket_size, prompt_len))
         if bucket_size > 0:
             assert "position_ids" not in model_kwargs, "Untested path"
-
+        cur_len = prompt_len
+        token_idx = model_kwargs.get("token_idx", None)
+        if token_idx is not None:
+            # Update cur_len in case of static shapes
+            cur_len = token_idx.item()
         while True:
             if lazy_mode:
                 self.htcore_generation.mark_step()
@@ -1427,6 +1431,7 @@ class GaudiGenerationMixin(GenerationMixin):
             model_kwargs = self._update_model_kwargs_for_generation(
                 outputs, model_kwargs, is_encoder_decoder=self.config.is_encoder_decoder
             )
+            cur_len = cur_len + 1
 
             # if eos_token was found in one sentence, set sentence to finished
             if not ignore_eos and eos_token_id_tensor is not None:
@@ -1438,7 +1443,7 @@ class GaudiGenerationMixin(GenerationMixin):
                     this_peer_finished = True
 
             # stop if we exceed the maximum length
-            if stopping_criteria(input_ids, scores, token_idx=token_idx):
+            if stopping_criteria(input_ids, scores, token_idx=cur_len):
                 this_peer_finished = True
 
             hb_profer.step()
@@ -1663,6 +1668,12 @@ class GaudiGenerationMixin(GenerationMixin):
         hb_profer = HabanaProfile(warmup=profiling_warmup_steps, active=profiling_steps)
         hb_profer.start()
         this_peer_finished = False  # used by synced_gpus only
+        cur_len = input_ids.shape[-1]
+        token_idx = model_kwargs.get("token_idx", None)
+        if token_idx is not None:
+            # Update cur_len in case of static shapes
+            cur_len = token_idx.item()
+
         # auto-regressive generation
         while True:
             if lazy_mode:
@@ -1746,7 +1757,7 @@ class GaudiGenerationMixin(GenerationMixin):
             model_kwargs = self._update_model_kwargs_for_generation(
                 outputs, model_kwargs, is_encoder_decoder=self.config.is_encoder_decoder
             )
-
+            cur_len = cur_len + 1
             # if eos_token was found in one sentence, set sentence to finished
             if not ignore_eos and eos_token_id_tensor is not None:
                 unfinished_sequences = unfinished_sequences.mul(
@@ -1758,7 +1769,7 @@ class GaudiGenerationMixin(GenerationMixin):
                     this_peer_finished = True
 
             # stop if we exceed the maximum length
-            if stopping_criteria(input_ids, scores, token_idx=token_idx):
+            if stopping_criteria(input_ids, scores, token_idx=cur_len):
                 this_peer_finished = True
 
             hb_profer.step()
@@ -2939,7 +2950,7 @@ class GaudiGenerationMixin(GenerationMixin):
 
             hb_profer.step()
 
-            if constrained_beam_scorer.is_done or stopping_criteria(input_ids, scores, token_idx=token_idx):
+            if constrained_beam_scorer.is_done or stopping_criteria(input_ids, scores, token_idx=cur_len):
                 if not synced_gpus:
                     break
                 else:
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -654,7 +654,7 @@ class GaudiGenerationMixin(GenerationMixin):
 
         self._validate_generated_length(
             generation_config,
-            model_kwargs["token_idx"] if "token_idx" in model_kwargs else input_ids_length,
+            model_kwargs["token_idx"].item() if "token_idx" in model_kwargs else input_ids_length,
             has_default_max_length,
         )
 
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -257,6 +257,7 @@ class GaudiGenerationMixin(GenerationMixin):
         decoder_start_token_id: int = None,
         bos_token_id: int = None,
         device: torch.device = None,
+        max_new_tokens: int = None,
     ) -> Tuple[torch.LongTensor, Dict[str, torch.Tensor]]:
         """Prepares `decoder_input_ids` for generation with encoder-decoder models"""
         # 1. Check whether the user has defined `decoder_input_ids` manually. To facilitate in terms of input naming,
@@ -281,9 +282,9 @@ class GaudiGenerationMixin(GenerationMixin):
             )
         else:
             # creating padded decoder_input_ids to achieve static shapes. Later new tokens once generated are copied in to decoder_input_ids based on token_idx
+            max_length = max_new_tokens + 1 if max_new_tokens is not None else self.generation_config.max_length
             decoder_input_ids_start = (
-                torch.ones((batch_size, self.generation_config.max_length), dtype=torch.long, device=device)
-                * decoder_start_token_id
+                torch.ones((batch_size, max_length), dtype=torch.long, device=device) * decoder_start_token_id
             )
 
         # no user input -> use decoder_start_token_id as decoder_input_ids
@@ -536,7 +537,7 @@ class GaudiGenerationMixin(GenerationMixin):
         generation_config = copy.deepcopy(generation_config)
         if generation_config.static_shapes is None:
             generation_config.static_shapes = self.config.model_type in MODELS_OPTIMIZED_WITH_STATIC_SHAPES
-            self.generation_config.static_shapes = generation_config.static_shapes
+        self.generation_config.static_shapes = generation_config.static_shapes
         if generation_config.ignore_eos is None:
             generation_config.ignore_eos = kwargs.get("ignore_eos", lazy_mode)
         generation_config.validate()
@@ -625,8 +626,13 @@ class GaudiGenerationMixin(GenerationMixin):
                 token_idx = 1
                 model_kwargs["token_idx"] = torch.tensor(token_idx, device=inputs_tensor.device)
                 if model_kwargs.get("decoder_attention_mask", None) is None and generation_config.use_cache:
+                    max_length = (
+                        generation_config.max_new_tokens + 1
+                        if generation_config.max_new_tokens is not None
+                        else generation_config.max_length
+                    )
                     model_kwargs["decoder_attention_mask"] = self._prepare_decoder_attention_mask(
-                        generation_config.max_length,
+                        max_length,
                         inputs_tensor.shape[0],
                         generation_config.pad_token_id,
                         inputs_tensor.device,
@@ -654,7 +660,6 @@ class GaudiGenerationMixin(GenerationMixin):
                 inputs_tensor, model_kwargs, model_input_name
             )
         # 5. Prepare `input_ids` which will be used for auto-regressive generation
-
         if self.config.is_encoder_decoder:
             input_ids, model_kwargs = self._prepare_decoder_input_ids_for_generation(
                 batch_size=batch_size,
@@ -663,6 +668,7 @@ class GaudiGenerationMixin(GenerationMixin):
                 decoder_start_token_id=generation_config.decoder_start_token_id,
                 bos_token_id=generation_config.bos_token_id,
                 device=inputs_tensor.device,
+                max_new_tokens=generation_config.max_new_tokens,
             )
         else:
             input_ids = inputs_tensor if model_input_name == "input_ids" else model_kwargs.pop("input_ids")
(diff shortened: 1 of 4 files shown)


sywangyi:
@regisss

sywangyi:
because the static shape in self.generation.config is not set (see https://github.com/huggingface/optimum-habana/pull/627/files#diff-284d6c109e6788a4405e302113203f6b7e98be68afc7dd3d85b89c6329b9275eL540-R541), the issue is hidden in optimum-habana text generation example.

sywangyi:
provide a simple test
```
from transformers import pipeline
import habana_frameworks.torch
import torch
from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
adapt_transformers_to_gaudi()
generation_kwargs = dict(do_sample=False, num_beams=4, use_cache=True, max_new_tokens=32)

generator = pipeline(
        "text-generation",
        model="gpt2",
        torch_dtype=torch.bfloat16,
        device="hpu",
        **generation_kwargs,
)

print(generator("DeepSpeed is a machine learning framework"))
```
before the fix, the output is
[{'generated_text': 'DeepSpeed is a machine learning framework'}]

after the fix, the output is
[{'generated_text': "DeepSpeed is a machine learning framework that can be used to build machine learning models.\n\nIn this tutorial, we'll learn how to build a machine learning model that can be used to build"}

sywangyi:
@libinta @mandy-li

optimum/habana/transformers/generation/utils.py
```diff
@@ -2290,11 +2300,10 @@ def expand_if_needed(tensor, new_size, value, dim=-1):
                 if (
                     self.generation_config.early_stopping
                     and is_min_length_reached
-                    and cur_len % self.generation_config.early_stopping_interval == 0
                     and num_eos_tokens >= num_beams_tensor
                 ):
                     break
-                elif cur_len == self.generation_config.max_length:
+                elif stopping_criteria(input_ids[:, :token_idx], scores):
```
bhargaveede, line 2306: This slicing here will break static_shapes, right?

sywangyi: why? same logic is here https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/generation/utils.py#L1423, the 2297 line makes the output length incorrect. the self.generation.max_length is not equal to generation.max_length outside beam_search.

optimum/habana/transformers/generation/utils.py
```diff
@@ -2128,6 +2137,8 @@ def expand_if_needed(tensor, new_size, value, dim=-1):
             inc = iter(incrementor(bucket_size, prompt_len))
         if bucket_size > 0:
             assert "position_ids" not in model_kwargs, "Untested path"
+        if self.generation_config.static_shapes:
+            initial_ids = input_ids[::num_beams, 0:cur_len]
```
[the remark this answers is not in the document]
sywangyi: input_ids's first dimension is beam_num*batch_size in beam_search. see https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/generation/utils.py#L912

optimum/habana/transformers/generation/utils.py
```diff
@@ -285,9 +286,9 @@ def _prepare_decoder_input_ids_for_generation(
             )
         else:
             # creating padded decoder_input_ids to achieve static shapes. Later new tokens once generated are copied in to decoder_input_ids based on token_idx
+            max_length = max_new_tokens + 1 if max_new_tokens is not None else self.generation_config.max_length
```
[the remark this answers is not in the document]
sywangyi: it's for decoder_input_ids_start in encoder-decoder model. the original size of decoder_input_ids_start is 1 if it's not for static shape. see https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/generation/utils.py#L283

sywangyi:
provide a simple test for summerization using encoder-decoder model.
```bash
from transformers import pipeline
import habana_frameworks.torch
import torch
from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
adapt_transformers_to_gaudi()
generation_kwargs = dict(do_sample=False, num_beams=4, use_cache=True, max_new_tokens=32)

generator = pipeline(
        "summarization",
        model="facebook/bart-large-cnn",
        torch_dtype=torch.bfloat16,
        device="hpu",
        **generation_kwargs,
)

print(generator(["(CNN)The Palestinian Authority officially became", "DeepSpeed is a machine learning"], batch_size=2))

```
without the fix. it will coredump

regisss:
@sywangyi It seems the PR "breaks" regular generation. Running
```
python run_generation.py --model_name_or_path gpt2 --use_hpu_graphs --use_kv_cache --max_new_tokens 100 --prompt "Hello world"
```
on main returns
```
01/15/2024 21:55:18 - INFO - __main__ - Running generate...                                                                                  
                                                                                                                                             
Input/outputs:                                                                                                                               
input 1: ('Hello world',)                                                                                                                    
output 1: ('Hello world, I\'m not sure what to say.\n\n"I\'m sorry, but I\'m not sure what to say.\n\n"I\'m sorry, but I\'m not sure what to 
say.\n\n"I\'m sorry, but I\'m not sure what to say.\n\n"I\'m sorry, but I\'m not sure what to say.\n\n"I\'m sorry, but I\'m not sure what to 
say.\n\n"I\'m sorry, but I\'m not',)                                                                                                         
                                                                                                                                             
                                                                                                                                             
Stats:                                                                                                                                       
--------------------------------------------------------------------------------------------------------------                               
Throughput (including tokenization) = 859.1097929752864 tokens/second                                                                        
Number of HPU graphs                = 12                                                                                                     
Memory allocated                    = 0.51 GB                                                                                                
Max memory allocated                = 0.62 GB                                                                                                
Graph compilation duration          = 1.3922957889735699 seconds
--------------------------------------------------------------------------------------------------------------
```
but on your branch I get
```
01/15/2024 21:55:53 - INFO - __main__ - Running generate...

Input/outputs:
input 1: ('Hello world',)
output 1: ('Hello world,',)


Stats:
-------------------------------------------------------------------------------------------------------------
Throughput (including tokenization) = 34314.56502749868 tokens/second
Number of HPU graphs                = 6
Memory allocated                    = 0.51 GB
Max memory allocated                = 0.62 GB
Graph compilation duration          = 0.5902187939500436 seconds
-------------------------------------------------------------------------------------------------------------
```

sywangyi:
> @sywangyi It seems the PR "breaks" regular generation. Running
> 
> ```
> python run_generation.py --model_name_or_path gpt2 --use_hpu_graphs --use_kv_cache --max_new_tokens 100 --prompt "Hello world"
> ```
> 
> on main returns
> 
> ```
> 01/15/2024 21:55:18 - INFO - __main__ - Running generate...                                                                                  
>                                                                                                                                              
> Input/outputs:                                                                                                                               
> input 1: ('Hello world',)                                                                                                                    
> output 1: ('Hello world, I\'m not sure what to say.\n\n"I\'m sorry, but I\'m not sure what to say.\n\n"I\'m sorry, but I\'m not sure what to 
> say.\n\n"I\'m sorry, but I\'m not sure what to say.\n\n"I\'m sorry, but I\'m not sure what to say.\n\n"I\'m sorry, but I\'m not sure what to 
> say.\n\n"I\'m sorry, but I\'m not',)                                                                                                         
>                                                                                                                                              
>                                                                                                                                              
> Stats:                                                                                                                                       
> --------------------------------------------------------------------------------------------------------------                               
> Throughput (including tokenization) = 859.1097929752864 tokens/second                                                                        
> Number of HPU graphs                = 12                                                                                                     
> Memory allocated                    = 0.51 GB                                                                                                
> Max memory allocated                = 0.62 GB                                                                                                
> Total memory available              = 94.62 GB
> Graph compilation duration          = 1.3922957889735699 seconds
> --------------------------------------------------------------------------------------------------------------
> ```
> 
> but on your branch I get
> 
> ```
> 01/15/2024 21:55:53 - INFO - __main__ - Running generate...
> 
> Input/outputs:
> input 1: ('Hello world',)
> output 1: ('Hello world,',)
> 
> 
> Stats:
> -------------------------------------------------------------------------------------------------------------
> Throughput (including tokenization) = 34314.56502749868 tokens/second
> Number of HPU graphs                = 6
> Memory allocated                    = 0.51 GB
> Max memory allocated                = 0.62 GB
> Total memory available              = 94.62 GB
> Graph compilation duration          = 0.5902187939500436 seconds
> -------------------------------------------------------------------------------------------------------------
> ```

upload a commit to fix it.

optimum/habana/transformers/generation/utils.py
```diff
@@ -2128,6 +2151,8 @@ def expand_if_needed(tensor, new_size, value, dim=-1):
             inc = iter(incrementor(bucket_size, prompt_len))
         if bucket_size > 0:
             assert "position_ids" not in model_kwargs, "Untested path"
+        if self.generation_config.static_shapes:
+            initial_ids = input_ids[::num_beams, 0:cur_len]
```
bhargaveede, line 2155: Just for my understanding, Is this slicing causing any bug?
I'm trying to understand as finalize_beams input shape will be changing now due to slice and will impact perf.

sywangyi: I see the initial_ids.to("cpu") in     
sequence_outputs["sequences"] = finalize_beams(
                initial_ids.cpu(), move(beam_trace, "cpu"), self.config, self.generation_config.length_penalty
            )
and it's in "cpu", will such variable length cause perf too? for decode-only model, the initial_ids is a variable length tensor, you could test with the text-generation pipeline I provide in the review comments.

sywangyi: I see 
    sequence_outputs["sequences"] = finalize_beams(
                initial_ids.cpu(), move(beam_trace, "cpu"), self.config, self.generation_config.length_penalty
            )
in the code. the initial_ids is in cpu when finalize_beams, will it also cause perf issue because of variable length?
for decode only case like text-generation, the initial_ids is a variable length tensor, you could test with the script I provide in the review comment

bhargaveede: @sywangyi I think having initial_ids on HPU with variable length cause perf degradation due to recompilation.
Moving to CPU might be better.

I will verify it once

sywangyi: thanks

bhargaveede:
I see perf drop 1.082(main) vs 0.972 (on your branch)
@sywangyi  Do you see the same?

python3 /root/optimum-habana/examples/summarization/run_summarization.py --model_name_or_path t5-3b --do_predict --predict_with_generate --dataset_name cnn_dailymail --dataset_config 3.0.0 --use_habana --per_device_eval_batch_size 2 --gaudi_config_name Habana/t5 --generation_num_beams 4 --ignore_pad_token_for_loss False --pad_to_max_length --use_hpu_graphs_for_inference --use_lazy_mode --max_predict_samples 200 --bf16 --bf16_full_eval --output_dir /tmp/tmpccx5bpdn

sywangyi:
> I see perf drop 1.082(main) vs 0.972 (on your branch) @sywangyi Do you see the same?
> 
> python3 /root/optimum-habana/examples/summarization/run_summarization.py --model_name_or_path t5-3b --do_predict --predict_with_generate --dataset_name cnn_dailymail --dataset_config 3.0.0 --use_habana --per_device_eval_batch_size 2 --gaudi_config_name Habana/t5 --generation_num_beams 4 --ignore_pad_token_for_loss False --pad_to_max_length --use_hpu_graphs_for_inference --use_lazy_mode --max_predict_samples 200 --bf16 --bf16_full_eval --output_dir /tmp/tmpccx5bpdn

Hi, the difference is caused by the change in 
image
the question is I think the logic of "cur_len == self.generation_config.max_length" as stop criteria is incorrect
1. self.generation_config.max_length is not equal to max_new_length + prompt len in generation_config. lead to output text with incorrected length. (you could use the text generation script I show in the comment to try.
2. there's a lot of stop criteria. even user could define their own stop criteria. and this logic break the rule.

optimum/habana/transformers/generation/utils.py
bhargaveede: Can we use stopping_criteria(input_ids, scores) instead of using input_ids[:, :token_idx] ?
That way we will still maintain static shape across runs and not impacting the perf.

bhargaveede:
> 1. cur_len == self.generation_config.max_length+2

Understood.
cur_len == self.generation_config.max_length+2 is also a static check whereas slicing will cause shape change across runs and that's impacting the perf. We need to use someway of using pad_token_id with input_ids (or) some other workaround to avoid the shape change which is impacting the perf

bhargaveede:
> > 1. cur_len == self.generation_config.max_length+2
> 
> Understood. cur_len == self.generation_config.max_length+2 is also a static check whereas slicing will cause shape change across runs and that's impacting the perf. We need to use someway of using pad_token_id with input_ids (or) some other workaround to avoid the shape change which is impacting the perf

Also, when ever user defines their own stop criteria, perf degradation is expected based on whether they support static_shapes (or) not. However, in default case, we should make it static_shapes for better perf. I will also brainstorm on ways to avoid slicing.

sywangyi:
@bhargaveede and @regisss the is max length criteria that we use to determine if the generation has ended, you could see it use  input_ids.shape[-1] and compare with max_length. if you want to use static shape. we need to pass token idx in. override the max length criteria in transformer. user defined criteria also need the token idx, becuase input_ids.shape contain pad length which is not reliable. what's your point? 
```
class MaxLengthCriteria(StoppingCriteria):
    """
    This class can be used to stop generation whenever the full generated number of tokens exceeds `max_length`. Keep
    in mind for decoder-only type of transformers, this will include the initial prompted tokens.

    Args:
        max_length (`int`):
            The maximum length that the output sequence can have in number of tokens.
        max_position_embeddings (`int`, `optional`):
            The maximum model length, as defined by the model's `config.max_position_embeddings` attribute.
    """

    def __init__(self, max_length: int, max_position_embeddings: Optional[int] = None):
        self.max_length = max_length
        self.max_position_embeddings = max_position_embeddings

    @add_start_docstrings(STOPPING_CRITERIA_INPUTS_DOCSTRING)
    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs) -> bool:
        cur_len = input_ids.shape[-1]
        is_done = cur_len >= self.max_length
        if self.max_position_embeddings is not None and not is_done and cur_len >= self.max_position_embeddings:
            logger.warning_once(
                "This is a friendly reminder - the current text generation call will exceed the model's predefined "
                f"maximum length ({self.max_position_embeddings}). Depending on the model, you may observe "
                "exceptions, performance degradation, or nothing at all."
            )
        return is_done
```

bhargaveede:
> @bhargaveede and @regisss the is max length criteria that we use to determine if the generation has ended, you could see it use input_ids.shape[-1] and compare with max_length. if you want to use static shape. we need to pass token idx in. override the max length criteria in transformer. user defined criteria also need the token idx, because input_ids.shape contain pad length which is not reliable. what's your point?
> 
> ```
> class MaxLengthCriteria(StoppingCriteria):
>     """
>     This class can be used to stop generation whenever the full generated number of tokens exceeds `max_length`. Keep
>     in mind for decoder-only type of transformers, this will include the initial prompted tokens.
> 
>     Args:
>         max_length (`int`):
>             The maximum length that the output sequence can have in number of tokens.
>         max_position_embeddings (`int`, `optional`):
>             The maximum model length, as defined by the model's `config.max_position_embeddings` attribute.
>     """
> 
>     def __init__(self, max_length: int, max_position_embeddings: Optional[int] = None):
>         self.max_length = max_length
>         self.max_position_embeddings = max_position_embeddings
> 
>     @add_start_docstrings(STOPPING_CRITERIA_INPUTS_DOCSTRING)
>     def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs) -> bool:
>         cur_len = input_ids.shape[-1]
>         is_done = cur_len >= self.max_length
>         if self.max_position_embeddings is not None and not is_done and cur_len >= self.max_position_embeddings:
>             logger.warning_once(
>                 "This is a friendly reminder - the current text generation call will exceed the model's predefined "
>                 f"maximum length ({self.max_position_embeddings}). Depending on the model, you may observe "
>                 "exceptions, performance degradation, or nothing at all."
>             )
>         return is_done
> ```

Cant we using StaticMaxLengthCriteria?

sywangyi:
now the the StoppingCriteria contained more than one criterias. for example. for text generation. it contains [<transformers.generation.stopping_criteria.MaxLengthCriteria object at 0x7fc31fc862f0>, <optimum.habana.transformers.generation.utils.StaticMaxLengthCriteria object at 0x7fc31fc86320>]. once one of the criteria meet exit requirment. the generaion will end. and if user defined criter like end in some specific "ids", correct logic is to get the latest generation ids and comparing. see if meet the ending requirement.

bhargaveede:
> now the the StoppingCriteria contained more than one criterias. for example. for text generation. it contains [<transformers.generation.stopping_criteria.MaxLengthCriteria object at 0x7fc31fc862f0>, <optimum.habana.transformers.generation.utils.StaticMaxLengthCriteria object at 0x7fc31fc86320>]. once one of the criteria meet exit requirment. the generaion will end. and if user defined criteria like end in some specific "ids", correct logic is to get the latest generation ids and comparing. see if meet the ending requirement.

Got it. I think both StaticMaxLengthCriteria and MaxLengthCritieria shouldn't be together. If it's there then we should avoid it.
AS per your earlier comment can we modify coden "cur_len == self.generation_config.max_length+2" to something like "cur_len == adjusted_max_length"?
_self.generation_config.max_length is not equal to max_new_length + prompt len in generation_config. lead to output text with incorrected length. (you could use the text generation script I show in the comment to try. also，using current logic will generate 2 less tokenizer than using stop criteria， if you use cur_len == self.generation_config.max_length+2，the perf is apple to apple._
@regisss 
In case of user defined criteria, It's fine if we miss static shape there as it depends on the user.
But for all other default criterias, We should maintain static shape validity in the checks. What do you suggest? @regisss

regisss:
@sywangyi @bhargaveede `MaxLengthCriteria` and `StaticMaxLengthCriteria` should be mutually exclusive if I understand correctly.

Ideally, we could remove the `StaticMaxLengthCriteria` class from the codebase and override the `MaxLengthCriteria` from Transformers:
- it should get `token_idx=None` as an optional argument
- if `token_idx is None`, fall back to Transformers implementation
- if `token_idx` is not None, compare it with max length

Would that solve this issue and work for you?

sywangyi:
agree, we should overide MaxLengthCriteria, pass token_idx in and remove StaticMaxLengthCriteria, which makes things simple.

bhargaveede:
Changes look fine to me. I will check locally if t5 and bart are fine with the patch.

optimum/habana/transformers/generation/utils.py
```diff
@@ -261,6 +248,7 @@ def _prepare_decoder_input_ids_for_generation(
         decoder_start_token_id: int = None,
         bos_token_id: int = None,
         device: torch.device = None,
+        max_new_tokens: int = None,
```
regisss, line 251: Isn't it possible to get it from `self.generation_config.max_new_tokens` instead of passing it as an argument here?

optimum/habana/transformers/generation/utils.py
```diff
@@ -664,6 +629,7 @@ def generate(
                 decoder_start_token_id=generation_config.decoder_start_token_id,
                 bos_token_id=generation_config.bos_token_id,
                 device=inputs_tensor.device,
+                max_new_tokens=generation_config.max_new_tokens,
```
regisss, line 632: Related to my comment above, we can set `self.generation_config.max_new_tokens = generation_config.max_new_tokens` to access it inside the method. WDYT?

regisss (commented):
I left a couple of comments. Also, there is now a merge conflict as I merged #651 yesterday. Let me know if you need any help to solve them.

optimum/habana/transformers/generation/utils.py
sywangyi: Hi, regisss, I think self.generation_config is from model hub like https://huggingface.co/meta-llama/Llama-2-7b-chat-hf/blob/main/generation_config.json. and generation_config is user defined config. if  user does not pass generation_config in . generation_config will be generated from self.generation_config. so do you think it's better to override parameter in self.generation_config using values from generation_config.

regisss: You're right, let's keep the value from `generation_config`

sywangyi:
> Hmm still seeing some throughput regressions, for example with BLOOMZ-7b:
> 
> ```
> python3 /root/workspace/optimum-habana/examples/text-generation/run_generation.py --model_name_or_path bigscience/bloomz-7b1 --batch_size 1 --use_kv_cache --max_new_tokens 100 --use_hpu_graphs --bf16
> ```
> 
> I get 37.98 tokens/s instead of 41.52 on Gaudi1, and 106.42 instead of 130.1 on Gaudi2. Do you get the same @sywangyi ?

yes, I reproduce it and find it's related with token_idx, because it's a hpu tensor? upload a PR to fix it.

sywangyi:
could you try by your side, and see if it's working now @regisss ?

regisss:
> could you try by your side, and see if it's working now @regisss ?

All regression tests passed, except the torch.compile one that seems to be broken:
```
  File "/root/workspace/optimum-habana/examples/text-generation/run_generation.py", line 562, in <module>
    main()
  File "/root/workspace/optimum-habana/examples/text-generation/run_generation.py", line 335, in main
    generate(None, args.reduce_recompile)
  File "/root/workspace/optimum-habana/examples/text-generation/run_generation.py", line 312, in generate
    outputs = model.generate(
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/optimum/habana/transformers/generation/utils.py", line 655, in generate
    self._validate_generated_length(
  File "/usr/local/lib/python3.10/dist-packages/transformers/generation/utils.py", line 1139, in _validate_generated_length
    if input_ids_length >= generation_config.max_length:
RuntimeError: [Rank:0] FATAL ERROR :: MODULE:PT_BRIDGE Exception in Lowering thread...
Graph compile failed. synStatus=synStatus 26 [Generice failure]. 
[Rank:0] Habana exception raised from compile at graph.cpp:503
```
Here is the command to reproduce it (torch.compile is available on Gaudi2 only):
```
PT_ENABLE_INT64_SUPPORT=1 PT_HPU_LAZY_MODE=0 python3 run_generation.py --model_name_or_path meta-llama/Llama-2-7b-hf --batch_size 1 --use_kv_cache --max_new_tokens 100 --attn_softmax_bf16 --reuse_cache --trim_logits --torch_compile --bf16
```
It comes from this line in Transformers: https://github.com/huggingface/transformers/blob/345b9b1a6a308a1fa6559251eb33ead2211240ac/src/transformers/generation/utils.py#L1139
Maybe we should override the `_validate_generated_length` method so that it takes the current length as an input argument for comparison with max length?

sywangyi:
Hi, @regisss , I use you command, and see different error in gaudi 2
error is 
  File "run_generation.py", line 562, in <module>
    main()
  File "run_generation.py", line 335, in main
    generate(None, args.reduce_recompile)
  File "run_generation.py", line 312, in generate
    outputs = model.generate(
  File "/usr/local/lib/python3.8/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/user/wangyi/optimum-habana/optimum/habana/transformers/generation/utils.py", line 789, in generate
    return self.greedy_search(
  File "/home/user/wangyi/optimum-habana/optimum/habana/transformers/generation/utils.py", line 1370, in greedy_search
    outputs = self(
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1521, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1530, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/user/wangyi/optimum-habana/optimum/habana/transformers/models/llama/modeling_llama.py", line 763, in forward
    outputs = self.model(
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1521, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1530, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/_dynamo/eval_frame.py", line 328, in _fn
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1521, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1530, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/user/wangyi/optimum-habana/optimum/habana/transformers/models/llama/modeling_llama.py", line 672, in forward
    layer_outputs = decoder_layer(
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1521, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1530, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/user/wangyi/optimum-habana/optimum/habana/transformers/models/llama/modeling_llama.py", line 465, in forward
    output_pre_attn, self_attn_weights, present_key_value = self.pre_attn(
  File "/home/user/wangyi/optimum-habana/optimum/habana/transformers/models/llama/modeling_llama.py", line 507, in pre_attn
    hidden_states = self.input_layernorm(hidden_states)
  File "/home/user/wangyi/optimum-habana/optimum/habana/transformers/models/llama/modeling_llama.py", line 507, in <resume in pre_attn>
    hidden_states = self.input_layernorm(hidden_states)
  File "/usr/local/lib/python3.8/dist-packages/torch/_dynamo/eval_frame.py", line 328, in _fn
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/_dynamo/external_utils.py", line 17, in inner
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/_functorch/aot_autograd.py", line 3917, in forward
    return compiled_fn(full_args)
  File "/usr/local/lib/python3.8/dist-packages/torch/_functorch/aot_autograd.py", line 1482, in g
    return f(*args)
  File "/usr/local/lib/python3.8/dist-packages/torch/_functorch/aot_autograd.py", line 2533, in runtime_wrapper
    all_outs = call_func_with_args(
  File "/usr/local/lib/python3.8/dist-packages/torch/_functorch/aot_autograd.py", line 1506, in call_func_with_args
    out = normalize_as_list(f(args))
  File "/usr/local/lib/python3.8/dist-packages/torch/_functorch/aot_autograd.py", line 1594, in rng_functionalization_wrapper
    return compiled_fw(args)
  File "/usr/local/lib/python3.8/dist-packages/torch/_functorch/aot_autograd.py", line 1482, in g
    return f(*args)
  File "<eval_with_key>.1220", line 17, in forward
  File "/usr/local/lib/python3.8/dist-packages/torch/_ops.py", line 448, in __call__
    return self._op(*args, **kwargs or {})
RuntimeError: shape mismatch: value tensor of shape [1, 32, 1, 128] cannot be broadcast to indexing result of shape [1, 32, 128]

Also there's code like in this PR
image
so I wander why this error occur

optimum/habana/transformers/generation/utils.py
```diff
@@ -681,8 +647,16 @@ def generate(
...
-            generation_config.max_length = generation_config.max_new_tokens + input_ids_length
-        self._validate_generated_length(generation_config, input_ids_length, has_default_max_length)
+            if "token_idx" in model_kwargs:
+                generation_config.max_length = input_ids_length
+            else:
+                generation_config.max_length = generation_config.max_new_tokens + input_ids_length
+
+        self._validate_generated_length(
+            generation_config,
+            model_kwargs["token_idx"] if "token_idx" in model_kwargs else input_ids_length,
```
regisss, line 657:
```suggestion
            model_kwargs["token_idx"].item() if "token_idx" in model_kwargs else input_ids_length,
```
This fixes it on my side. Can you try it and see if that solves your error?

sywangyi: my gaudi2 image and firmware are 1.13 since my system is ubuntu 20.04, and I have some issue to upgrade to 22.04. 1.14 does only support ubuntu 20.04.I think this is why my error is different with you

regisss: Ah yes that must be related then!

regisss (approved):

PR #661, opened 2024-01-24
Fix error in PR#654

schoi-habana:
ensure backward compatibility for models other than Falcon

## Before submitting


mandy-li (commented):
@schoi-habana , can you combine the two conditions to one so that only one `else` will be needed?

mandy-li (approved):

PR #669, opened 2024-01-30
Fix tests

ssarkar2:
# What does this PR do?

Certain tests like these [here](https://github.com/huggingface/optimum-habana/blob/937a411fc59b3a926f4894876929b3b60f4c634b/tests/transformers/tests/generation/test_utils.py#L880) calls `greedy_search` and `beam_search` directly without [generate](https://github.com/huggingface/optimum-habana/blob/937a411fc59b3a926f4894876929b3b60f4c634b/optimum/habana/transformers/generation/utils.py#L396)

flags like `reduce_recompile` and `bucket_size` are set to default values [here](https://github.com/huggingface/optimum-habana/blob/937a411fc59b3a926f4894876929b3b60f4c634b/optimum/habana/transformers/generation/utils.py#L596), hence they were accessed without defaults in `greedy_search` and `beam_search`. However of the *_search functions are called directly (like in the unit tests), it can fail.

Therefore setting defaults in this PR

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -1368,8 +1368,8 @@ class GaudiGenerationMixin(GenerationMixin):
         hb_profer = HabanaProfile(warmup=profiling_warmup_steps, active=profiling_steps)
         hb_profer.start()
         this_peer_finished = False  # used by synced_gpus only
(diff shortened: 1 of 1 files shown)


regisss (approved):
LGTM

PR #675, opened 2024-01-31
Added check if reuse_cache supported by selected model_type

iermolae:
# Add a check if reuse_cache flag supported by selected model_type in GaudiGenerationMixin.generate()

Previously, if run_generation.py ran with other than "llama" model and --reuse_check there was an error:

```AttributeError: 'GaudiMistralForCausalLM' object has no attribute 'allocate_kv_cache'```

Now there is a more user-freindly message describing the problem:

```AssertionError: reuse_cache only supported by llama at the moment```


ssarkar2 (approved):

regisss (approved):
LGTM!

PR #651, opened 2024-01-21
Upgrade to Transformers 4.37

regisss:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #687, opened 2024-02-05
Add cmdline arguments to run phi-2

nprotasov:
To run Phi-2 model we need to do some workarounds, because Phi-2 is require  (4.37.0.dev) of transformers, we can't call adapt_transformers_to_gaudi and we need to pass trust_remote_code to AutoModelForCausalLM
1) trust_remote_code - to load phi-2 model
2) default_transformers - this argument will disable adapt_transformers_to_gaudi() method, and Habana inference args

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -232,6 +232,20 @@ def setup_parser(parser):
         action="store_true",
         help="Whether to use torch compiled model or not.",
     )
+    parser.add_argument(
+        "--trust_remote_code",
+        action="store_true",
+        help="Whether or not to allow for custom models defined on the Hub in their own modeling files. This option"
+             "should only be set to `True` for repositories you trust and in which you have read the code, as it will "
+             "execute code present on the Hub on your local machine.",
+    )
+    parser.add_argument(
+        "--default_transformers",
+        action="store_true",
+        help="Use Hugging Face transformers implementation"
+    )
+
+
     parser.add_argument("--temperature", default=1.0, type=float, help="Temperature value for text generation")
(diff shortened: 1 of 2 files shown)


regisss:
I think we should just wait for https://github.com/huggingface/optimum-habana/pull/651 to be merged, which should happen very soon

nprotasov:
> I think we should just wait for #651 to be merged, which should happen very soon

This workarounds are not specific for Phi-2, it's useful for all models which require trust_remote_code or need another transformers version

regisss:
> > I think we should just wait for #651 to be merged, which should happen very soon
> 
> This workarounds are not specific for Phi-2, it's useful for all models which require trust_remote_code or need another transformers version

I get that, but the problem is that we have absolutely no guarantee on the API provided by this kind of models. The name of the inputs or methods could be different from what Transformers expects. So it's basically impossible to ensure that these models will work.
I think an issue should be opened by users if they cannot make it work for a specific model. Or we can add a workaround for trendy models as we did for Falcon.
Now, if customers and users are asking for it, we can move forward and do something with this PR. Otherwise, I would rather keep it as it is.

regisss:
@nprotasov The release of Optimum Habana v1.10.2 is fully compatible with Transformers v4.37 so phi-2 can be used out of the box. I quickly tried the text-generation example with
```
python run_generation.py \
  --model_name_or_path microsoft/phi-2 \
  --use_hpu_graphs \
  --use_kv_cache \
  --max_new_tokens 100 \
  --do_sample \
  --prompt "Here is my prompt"
```
and it ran successfully. Note that it is not compatible with a static KV cache, which leads to a high number of generated HPU graphs (and thus high compilation time) due to the increasing size of this cache throughout the generation process. We could implement it and override parts of its modeling as we do for Llama and other optimized models if needed.

PR #690, opened 2024-02-06
Patch for Gaudi Text-Generation Pipeline

sjagtap1803:
# What does this PR do?
This PR includes some minor changes to the text-generation pipeline code for langchain==0.0.191 compatibility.

## Before submitting

### DIFF
--- a/examples/text-generation/text-generation-pipeline/README.md
+++ b/examples/text-generation/text-generation-pipeline/README.md
@@ -31,6 +31,11 @@ If you plan to use [DeepSpeed-inference](https://docs.habana.ai/en/latest/PyTorc
 pip install git+https://github.com/HabanaAI/DeepSpeed.git@1.14.0
 ```
 
+If you would like to use the pipeline with LangChain classes, you can install LangChain as follows:
+```bash
+pip install langchain==0.0.191
+```
+
 ## Usage
 
 To run generation with DeepSpeed-inference, you must launch the script as follows:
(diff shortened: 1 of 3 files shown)


sjagtap1803:
Hi @regisss. As per our conversation this morning, I have changed the output format of the pipeline class in order to make it work with langchain==0.0.191. 

I would appreciate it if you could try running the `run_pipeline.py` and let me know if you face any issues. Thanks!

regisss (commented):
Maybe we can test the type of the output and keep the former way if it is a string (I guess it is, maybe the type is different) else  return the new format. WDYT?

sjagtap1803:
> Maybe we can test the type of the output and keep the former way if it is a string (I guess it is, maybe the type is different) else return the new format. WDYT?

I tried the former output type (string) but faced some issues with langchain==0.0.191. It looks like langchain expects the output to be a dictionary wrapped in a list as shown here: https://github.com/langchain-ai/langchain/blob/b3ae6bcd3f42ec85ee65eb29c922ab22a17a0210/langchain/llms/huggingface_pipeline.py#L169

regisss:
Yep I understand. What I propose is to have something like:
```py
if isinstance(output, str):
    # return the same as before
else:
    # return LangChain format
```
Would that work for you?

sjagtap1803:
> Yep I understand. What I propose is to have something like:
> 
> ```python
> if isinstance(output, str):
>     # return the same as before
> else:
>     # return LangChain format
> ```
> 
> Would that work for you?

Where do you suggest adding this check?

regisss:
@sjagtap1803 That looks good to me :+1:

regisss (commented):
Could you also run the following from the root of the repo please?
```
pip install -U ruff
make style
```

Besides, do you already know which version of LangChain should be used? We should specify it in the README.

sjagtap1803:
I applied code formatting by running `make style`. 

Regarding the LangChain version, I will test a few examples with 0.0.191 later today and update the blog accordingly. If the examples run as expected, I will specify the version in the README.

regisss (approved):

PR #694, opened 2024-02-07
Update transformer model test script to match with 4.37.1

yeonsily:
Fixed two issues from tests.models.bert.test_modeling_bert.BertModelTest

1) test_model_weights_reload_no_missing_tied_weights
The script is updated same as https://github.com/huggingface/transformers/blob/v4.37.2/tests/test_modeling_common.py#L1912

2) test_save_load_fast_init_to_base
Change 'model.config.save_pretrained(tmpdirname)' to 'model.save_pretrained(tmpdirname)'
Referred from https://github.com/huggingface/transformers/issues/9517

### DIFF
--- a/tests/transformers/tests/test_modeling_common.py
+++ b/tests/transformers/tests/test_modeling_common.py
@@ -83,6 +83,7 @@ if is_accelerate_available():
 
 if is_torch_available():
     import torch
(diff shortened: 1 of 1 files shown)


regisss (approved):
LGTM!

PR #695, opened 2024-02-07
Sarkar/fix test

ssarkar2:
# What does this PR do?
Fixes this test: `python -m pytest -vs -k test_assisted_decoding_sample bert/test_modeling_bert.py`

First it makes some changes in `optimum/habana/transformers/generation/utils.py` (like removing `assistant_model=assistant_model` to bring the code in line with 4.37). Similarly `test_assisted_decoding_sample` is brought up to date with 4.37

Then we observe a bug/error in `_speculative_sampling` in this line: `q_i = q[:, torch.arange(candidate_length), new_candidate_input_ids].squeeze(0, 1)`

So we monkey patch the test's model to supply it with an `assisted_decoding` which will call a local `speculative_sampling` which has the workaround line: `q_i = q[:, torch.arange(candidate_length), new_candidate_input_ids.squeeze()].squeeze(0, 1)`

Once the SW indexing issue is fixed, the monkey patch can be removed.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -812,7 +812,6 @@ class GaudiGenerationMixin(GenerationMixin):
             return self.assisted_decoding(
                 input_ids,
                 candidate_generator=candidate_generator,
-                assistant_model=assistant_model,
                 do_sample=generation_config.do_sample,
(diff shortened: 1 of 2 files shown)


regisss (approved):
LGTM!

PR #693, opened 2024-02-07
add support for TinyLlama model

tjs-intel:
:warning: **Do not merge this PR before Habana DeepSpeed 1.15 is released** :warning: 

# What does this PR do?

The [TinyLlama model](https://huggingface.co/TinyLlama/TinyLlama-1.1B-Chat-v1.0) only has checkpoints in the form of `model.safetensors`. This checkpoint needs to be included in the list of checkpoints that is passed to DeepSpeed in order for the model to function properly when initialized with DeepSpeed.

This PR adds the safetensor checkpoints to the list of checkpoints passed to DeepSpeed.

**Note**: This change requires an [upstream commit in microsoft/DeepSpeed](https://github.com/microsoft/DeepSpeed/commit/c8c57b8c24ad761267ab43b0252fb62f0c30d289) to be merged downstream to HabanaAI/DeepSpeed in order for DeepSpeed to support the provided safetensor format.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #697, opened 2024-02-08
enable falcon-180b inference

schoi-habana:
1 The current code can load only .bin or .pt whereas Falcon-180B ckpt files are in .safetensors format. This change enabled loading ckpt files in .safetensor format
2 The ValueError "The following model_kwargs are not used by the model: ['token_type_ids'] )" is addressed by introducing a workaround that removes ‘token_type_ids’ from the model_kwargs.

## Before submitting


regisss:
@schoi-habana From what I understood in [this PR](https://github.com/huggingface/optimum-habana/pull/693), loading safetensors checkpoint with DeepSpeed is not supported yet with Habana's DeepSpeed fork v1.14. So my understanding is that it will work when v1.15 is released right?

schoi-habana:
@regisss that's true. Please merge this after 1.15.0 release

schoi-habana:
no longer needed as https://github.com/huggingface/optimum-habana/pull/773 is a permanent solution

PR #698, opened 2024-02-09
Remove falcon model cache unit test 

jiminha:
Test failed with below error. 
E       AttributeError: 'GaudiFalconForCausalLM' object has no attribute '_convert_to_rw_cache'

_convert_to_rw_cache is no longer supported in transformer v4.37.2. Test case is removed.


regisss (approved):
LGTM!

PR #700, opened 2024-02-09
437 upgrade fix for falcon

libinta:
As torch.all inside _prepare_4d_causal_attention_mask_for_sdpa causes… d2h copy and impact falcon performance,

 we disable the falcon model to use _prepare_4d_causal_attention_mask_for_sdpa. Instead passing normal constructed 4D
to sdpa.

# What does this PR do?

Fixes # (issue)

## Before submitting


regisss (approved):
LGTM

PR #709, opened 2024-02-13
To fix LLAMA-V2-70B-FT-HF (8x) for eager mode

mkumargarg:
'mark_step()' should not be called for eager mode so conditionally skipped for eager mode.


regisss (approved):
LGTM, good catch!

PR #711, opened 2024-02-13
Adding a flag whether to save checkpoint or not.

yeonsily:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/language-modeling/run_clm.py
+++ b/examples/language-modeling/run_clm.py
@@ -646,7 +646,7 @@ def main():
         elif last_checkpoint is not None:
             checkpoint = last_checkpoint
         train_result = trainer.train(resume_from_checkpoint=checkpoint)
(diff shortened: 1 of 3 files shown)


regisss:
Adding the training argument `save_strategy="no"` should lead to no checkpoint saved no?

yeonsily:
> Adding the training argument `save_strategy="no"` should lead to no checkpoint saved no?

I think they used for different cases.
1. save_strategy is for saving checkpoint per epoch or per checkpoint step.
If we set this to 'no', it will still save the last checkpoint at the end of the training.
 
2. trainer.save_model() in the script is the one for saving last checkpoint at the end of the training.
But if save_strategy is not 'no', it may still save checkpoint during the training per epoch or steps.

regisss (approved):
LGTM

PR #713, opened 2024-02-14
assert added for BART only works with num_beams=1

srajabos:
# What does this PR do?

This PR asserts that num_beams =1 is the only supported for BART based models


regisss (commented):
I think this should be managed at the generation level here: https://github.com/huggingface/optimum-habana/blob/00c8ec4088a47d10dfe9eacf15adacf49c19cf0f/optimum/habana/transformers/generation/utils.py#L394
Otherwise this check will only be performed in the summarization example.

@bhargaveede I don't remember if this num beam restriction was done on design or not

PR #715, opened 2024-02-15
Fix throughput calculation for diffusion models

skavulya:
# What does this PR do?
Fixes the throughput calculation for stable diffusion models when using configurable warm-up steps.

## Before submitting

### DIFF
--- a/optimum/habana/diffusers/pipelines/controlnet/pipeline_controlnet.py
+++ b/optimum/habana/diffusers/pipelines/controlnet/pipeline_controlnet.py
@@ -616,7 +616,9 @@ class GaudiStableDiffusionControlNetPipeline(GaudiDiffusionPipeline, StableDiffu
             speed_measures = speed_metrics(
                 split=speed_metrics_prefix,
                 start_time=t0,
-                num_samples=num_batches * batch_size if t1 == t0 else (num_batches - throughput_warmup_steps) * batch_size,
+                num_samples=num_batches * batch_size
+                if t1 == t0
+                else (num_batches - throughput_warmup_steps) * batch_size,
                 num_steps=num_batches,
                 start_time_after_warmup=t1,
             )
--- a/optimum/habana/diffusers/pipelines/controlnet/pipeline_controlnet.py
+++ b/optimum/habana/diffusers/pipelines/controlnet/pipeline_controlnet.py
@@ -496,10 +496,11 @@ class GaudiStableDiffusionControlNetPipeline(GaudiDiffusionPipeline, StableDiffu
             hb_profiler.start()
 
             # 8. Denoising loop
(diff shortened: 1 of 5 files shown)


optimum/habana/diffusers/pipelines/controlnet/pipeline_controlnet.py
```diff
@@ -496,10 +496,11 @@ def __call__(
             hb_profiler.start()
 
             # 8. Denoising loop
+            throughput_warmup_steps = kwargs.get("throughput_warmup_steps", 3)
```
skavulya, line 499: @regisss Should we change the default from 3 to 2 since the compilation overhead occurs in the first 2 iterations?

regisss: Do you get the same throughput values with either `throughput_warmup_steps=2` or `throughput_warmup_steps=3`? I changed that last week because I saw different numbers, or maybe not all pipelines are equal regarding that.

skavulya: Yes, we get the same throughput values with `throughput_warmup_steps=2` and `throughput_warmup_steps=3`. The earlier code had an issue where we were skipping 3 warmup iterations but only subtracting 2 in `speed_metrics` so the reported throughput was higher than it should be. 

I think using `throughput_warmup_steps=2` is preferable because in situations where `num_batches` <  `throughput_warmup_steps` we do not subtract any warmup steps so the reported throughput is low.

@dsocek  and I also observed that the inference throughput is significantly higher if we don't use hpu_graphs and we would like to update the recommended options in the readme to reflect this. We can do this in a separate PR.

regisss: My bad, good catch! Let's switch to 2 then :slightly_smiling_face: 

Regarding the use of HPU graphs, yep let's do that in another PR!

skavulya: @regisss Thanks Regis. I re-run on Synapse 1.14.0 and I am seeing warmup-overhead for 3 steps so I think we should keep the warmup steps as 3 so we can merge the PR if it looks ok to you.

regisss: Sounds good!
Let's just run `make style` and then I'll merge :slightly_smiling_face:

skavulya: Will do. Thanks

skavulya:
@regisss I ran the style check. Thanks

regisss (approved):
Thanks for fixing this!

PR #717, opened 2024-02-16
Add new arg `eval_split` to LoRA example

regisss:
# What does this PR do?

Used for [timdettmers/openassistant-guanaco](https://huggingface.co/datasets/timdettmers/openassistant-guanaco) since it doesn't contain a validation set and this now triggers an error using the Datasets library.

## Before submitting

### DIFF
--- a/examples/language-modeling/run_lora_clm.py
+++ b/examples/language-modeling/run_lora_clm.py
@@ -248,6 +248,10 @@ class DataArguments:
         default=False,
         metadata={"help": "Whether to have a SQL style prompt"},
     )
+    eval_split: str = field(
+        default="validation",
+        metadata={"help": "The split to use to set up the evaluation set with `--do_eval`."},
+    )
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


examples/language-modeling/run_lora_clm.py
```diff
@@ -612,13 +616,13 @@ def concatenate_data(dataset, max_seq_length):
...
             if training_args.do_eval:
-                tokenized_datasets_eval_ = tokenized_datasets["validation"].remove_columns(
+                tokenized_datasets_eval_ = tokenized_datasets[data_args.eval_split].remove_columns(
                     ["prompt_sources", "prompt_targets"]
                 )
         elif data_args.dataset_name == "timdettmers/openassistant-guanaco":
             tokenized_datasets_ = tokenized_datasets["train"].remove_columns(["input", "output"])
             if training_args.do_eval:
-                tokenized_datasets_eval_ = tokenized_datasets["test"].remove_columns(["input", "output"])
+                tokenized_datasets_eval_ = tokenized_datasets[data_args.eval_split].remove_columns(["input", "output"])
```
libinta, line 625: should we add --eval_split to "test" to falcon to get the same dataset as before?

regisss: Since you managed to reach the same perf as before using a different validation split percentage, I think we can just close this PR to not overcomplicate things. WDYT?

libinta: sure. let's close it.

PR #720, opened 2024-02-18
enable internal kv bucket in llama

xt574chen:
# What does this PR do?

To enhance throughput in scenarios with long new tokens, break down the KV cache into multiples of the bucket width. Use this to compute attention rather than using the entire KV cache.

LLaMA v2 70B (8x, max_input_tokens 128, max_new_tokens 2048, batch_size 240):
**5528 tps (original performance) -> 6378 tps (w/ internal bucket size 128)**

Add `--bucket_size=128 --bucket_internal` to the commands to enable the feature.

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -568,7 +568,9 @@ class GaudiGenerationMixin(GenerationMixin):
         if generation_config.reuse_cache:
             assert self.config.model_type in ["llama"], "reuse_cache only supported by llama at the moment"
             if not generation_config.bucket_internal:
-                assert generation_config.bucket_size <= 0, "please set bucket_internal along with reuse_cache and bucket_size"
+                assert (
+                    generation_config.bucket_size <= 0
+                ), "please set bucket_internal along with reuse_cache and bucket_size"
             else:
                 assert generation_config.bucket_size >= 0, "please set valid bucket_size to use bucket_internal"
 
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -568,9 +568,9 @@ class GaudiGenerationMixin(GenerationMixin):
         if generation_config.reuse_cache:
             assert self.config.model_type in ["llama"], "reuse_cache only supported by llama at the moment"
             if not generation_config.bucket_internal:
-                assert generation_config.bucket_size <= 0, "reuse_cache and bucketing flags set together"
+                assert generation_config.bucket_size <= 0, "please set bucket_internal along with reuse_cache and bucket_size"
             else:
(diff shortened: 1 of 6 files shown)


optimum/habana/transformers/generation/utils.py
```diff
@@ -550,19 +550,27 @@ def generate(
...
         )
         if model_kwargs["reduce_recompile"]:
             assert generation_config.bucket_size
         if generation_config.reuse_cache:
             assert self.config.model_type in ["llama"], "reuse_cache only supported by llama at the moment"
-            assert generation_config.bucket_size <= 0, "reuse_cache and bucketing flags set together"
+            if not generation_config.bucket_internal:
+                assert generation_config.bucket_size <= 0, "reuse_cache and bucketing flags set together"
+            else:
+                assert generation_config.bucket_size >= 0, "bucket_internal and bucket_size flags set together"
```
regisss, line 573: Here, we are in the case where `generation_config.bucket_internal` is True, so if this assert fails (i.e. `generation_config.bucket_size` < 0), it means that bucket_size is not set right? But the error message says otherwise

xt574chen: @puneeshkhanna I see you've corrected some error messages. I hope this update won't cause conflict.

optimum/habana/transformers/generation/utils.py
```diff
@@ -550,19 +550,27 @@ def generate(
...
         model_kwargs["reduce_recompile"] = (
             generation_config.reduce_recompile if generation_config.reduce_recompile is not None else False
         )
         if model_kwargs["reduce_recompile"]:
             assert generation_config.bucket_size
         if generation_config.reuse_cache:
             assert self.config.model_type in ["llama"], "reuse_cache only supported by llama at the moment"
-            assert generation_config.bucket_size <= 0, "reuse_cache and bucketing flags set together"
+            if not generation_config.bucket_internal:
+                assert generation_config.bucket_size <= 0, "please set bucket_internal along with reuse_cache and bucket_size"
```
regisss, line 571: Is it not possible to use bucket_size and reuse_cache without bucket_internal?

xt574chen: It's the logical in the origin code 
https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/generation/utils.py#L565

optimum/habana/transformers/generation/utils.py
puneeshkhanna: @xt574chen - I will update my PR once this gets merged first

regisss:
@xt574chen Can you run the following from the root of the repo to make the code style check pass please?
```
pip install -U ruff
make style
```

regisss (approved):

PR #722, opened 2024-02-19
speecht5 optimization

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/speecht5/modeling_speecht5.py
+++ b/optimum/habana/transformers/models/speecht5/modeling_speecht5.py
@@ -1,0 +1,552 @@
+from typing import List, Optional, Tuple, Union
+
+import torch
+import torch.utils.checkpoint
+from torch import nn
+from transformers.integrations.deepspeed import is_deepspeed_zero3_enabled
+from transformers.modeling_attn_mask_utils import _prepare_4d_attention_mask
+from transformers.modeling_outputs import BaseModelOutputWithPastAndCrossAttentions
+from transformers.models.speecht5.modeling_speecht5 import SpeechT5EncoderWithSpeechPrenet, SpeechT5PreTrainedModel
+from transformers.utils import logging
+
+from ...modeling_attn_mask_utils import (
+    _gaudi_prepare_4d_causal_attention_mask,
+)
+
+
+logger = logging.get_logger(__name__)
+
+
+def gaudi_SpeechT5SpeechDecoderPrenet_forward(
+    self,
+    input_values: torch.Tensor,
+    speaker_embeddings: Optional[torch.Tensor] = None,
+):
+    """
+    Copied from SpeechT5SpeechDecoderPrenet.forward: https://github.com/huggingface/transformers/blob/v4.37.2/src/transformers/models/speecht5/modeling_speecht5.py
+    The only differences are:
+    - disable dropout in inference, or else hpu graph could not be used
+    """
+
+    inputs_embeds = input_values
+    for layer in self.layers:
+        inputs_embeds = nn.functional.relu(layer(inputs_embeds))
+        if self.training:
+            inputs_embeds = self._consistent_dropout(inputs_embeds, self.config.speech_decoder_prenet_dropout)
+
+    inputs_embeds = self.final_layer(inputs_embeds)
+    inputs_embeds = self.encode_positions(inputs_embeds)
+
+    if speaker_embeddings is not None:
+        speaker_embeddings = nn.functional.normalize(speaker_embeddings)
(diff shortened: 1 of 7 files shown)


sywangyi:
from transformers import pipeline
from datasets import load_dataset
import soundfile as sf
import torch
import time
import habana_frameworks.torch

from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
adapt_transformers_to_gaudi()

synthesiser = pipeline("text-to-speech", "microsoft/speecht5_tts", device="hpu", torch_dtype=torch.bfloat16)

embeddings_dataset = load_dataset("Matthijs/cmu-arctic-xvectors", split="validation")
speaker_embedding = torch.tensor(embeddings_dataset[7306]["xvector"]).unsqueeze(0).to("hpu")

with torch.autocast("hpu", torch.bfloat16, True), torch.no_grad(), torch.inference_mode():
    for i in range(20):
        s0 = time.time()
        speech = synthesiser("Hello, my dog is cooler than you!", forward_params={"speaker_embeddings": speaker_embedding})
        print((time.time()-s0)*1000)
sf.write("speech.wav", speech["audio"], samplerate=speech["sampling_rate"])
~

sywangyi:
before optimization:
89197.1423625946
893.7644958496094
5745.201826095581
919.0852642059326
880.3527355194092
5667.682886123657
862.3275756835938
879.6939849853516
875.6730556488037
5622.7867603302
886.8467807769775
852.0243167877197
849.4389057159424
871.5994358062744
859.8270416259766
815.2995109558105
687.913179397583
662.3406410217285
663.1982326507568
717.8483009338379

after optimization:
6845.948219299316
255.29980659484863
218.94311904907227
221.1463451385498
221.50874137878418
218.1394100189209
214.89882469177246
216.66884422302246
220.9482192993164
221.04668617248535
226.33123397827148
223.01626205444336
226.28045082092285
220.9162712097168
221.33421897888184
223.2530117034912
222.49364852905273
223.09398651123047
223.03080558776855
223.14834594726562

sywangyi:
@regisss @libinta could someone review it？ I also add example for text to speech.

optimum/habana/transformers/models/speecht5/modeling_speecht5.py
```diff
@@ -0,0 +1,552 @@
...
+
+    # Start the output sequence with a mel spectrum that is all zeros.
+    output_sequence = encoder_last_hidden_state.new_zeros(bsz, 1, model.config.num_mel_bins)
+    output_sequence = torch.nn.functional.pad(output_sequence, (0, 0, 0, maxlen - 1), value=model.config.pad_token_id)
+    spectrogram = []
+    cross_attentions = []
+    past_key_values = None
+    idx = 0
+    result_spectrogram = {}
+    token_idx = torch.tensor(1, device=output_sequence.device)
```
regisss, line 462: This means that static KV cache will always be used right?

regisss (commented):
It looks good to me. I just have one question regarding always setting
```py
token_idx = torch.tensor(1, device=output_sequence.device)
```
Maybe this should be done only when `self.generation_config.static_shapes` is True, WDYT?

optimum/habana/transformers/models/speecht5/modeling_speecht5.py
sywangyi: yes, if you see _generate_speech in https://github.com/huggingface/transformers/blob/main/src/transformers/models/speecht5/modeling_speecht5.py#L2459, no kwargs is passed here. so the static kv cache will always be used here.

sywangyi:
also speecht5 has its own generate func. see https://github.com/huggingface/transformers/blob/main/src/transformers/models/speecht5/modeling_speecht5.py#L2756.
if we use pipeline, the calling sequence like pipeline forward->model.generate->_generate_speech
also we could use application like model.generate_speech calling sequence like model.generate_speech->_generate_speech.
to cover both case. I just assupt static shape is always true. WDYT?

regisss:
> also speecht5 has its own generate func. see https://github.com/huggingface/transformers/blob/main/src/transformers/models/speecht5/modeling_speecht5.py#L2756. if we use pipeline, the calling sequence like pipeline forward->model.generate->_generate_speech also we could use application like model.generate_speech calling sequence like model.generate_speech->_generate_speech. to cover both case. I just make a assumption that static shape is always true since the overiden _generate_speech is just enabled in Gaudi. WDYT?

Yes I agree with you, let's keep it like this and if anyone needs dynamic shapes we'll change that later.

regisss (approved):

PR #729, opened 2024-02-21
Update text generation example to support qwen model

mgonchar:
# What does this PR do?

Small fixes to support text generation with Qwen.

As stated in [doc](https://github.com/QwenLM/Qwen/blob/main/tokenization_note.md):
```
The concepts of bos, eos, unk, pad, mask, sep and such are not applicable to our pretrained models (Qwen-7B and Qwen-7B-Chat).
```
So special handling for ```pad``` token is required.

## Before submitting

### DIFF
--- a/examples/text-generation/requirements.txt
+++ b/examples/text-generation/requirements.txt
@@ -1,2 +1,5 @@
 datasets
 peft
+transformers_stream_generator
+einops
+tiktoken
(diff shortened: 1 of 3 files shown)


examples/text-generation/requirements.txt
```diff
@@ -1,2 +1,5 @@
 datasets
 peft
+transformers_stream_generator
```
mgonchar, line 3: Not sure if it is right solution to expand this requirement file. As alternative I see adding note in doc, indicating that additional packages required. Please suggest which one is better option

mgonchar:
@regisss thanks for your review. I got your point, closing PR then

regisss:
We can revisit this if Qwen makes it into Transformers :slightly_smiling_face:

PR #736, opened 2024-02-23
Adding a flag whether to save checkpoint or not in run_lora_clm.py

yeonsily:
This is the same change as PR #711 but for different script.

# What does this PR do?

Fixes # (issue)

## Before submitting


regisss:
@yeonsily Can you run `make style` please?

yeonsily:
> @yeonsily Can you run `make style` please?

Strange.. I didn't see any style change from make style.. Let me check further.

regisss (approved):
LGTM

@yeonsily I've seen a few occurrences where the same version of `ruff` produces different results in two different environments. It's rare enough so that it's not really an issue but it may be the reason why you got a different output :slightly_frowning_face:

PR #752, opened 2024-03-03
Fix Llama-70B-FSDP model loading issue

hlahkar:
When running large models with FSDP, if we try to load full model params to each card we will run out of CPU RAM (refer to this these PRs https://github.com/huggingface/transformers/pull/25107  https://github.com/huggingface/accelerate/pull/1777. This is avoided by loading model params only to card 0 and broadcasting params to other cards from card 0. Enabled CPU_RAM_EFFICIENT_LOADING configuration in accelerate to support above described functionality.


vivekgoe (changes requested):
Please add a brief description of changes done in this PR in 1st section of PR.

vivekgoe (approved):

PR #757, opened 2024-03-04
Fix BART inference failure due to attn_implementation initialization

vivekgoe:
Fix BART inference failure due to attn_implementation initialization to "sdpa".  Initialize attn_implementation to use "eager" to avoid failure seen with "sdpa". 
Note that this failure is seen only with torch > 2.1.1 because for higher torch versions transformer chooses "sdpa" as default for config._attn_implementation.  

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/summarization/run_summarization.py
+++ b/examples/summarization/run_summarization.py
@@ -477,6 +477,7 @@ def main():
         token=model_args.token,
         trust_remote_code=model_args.trust_remote_code,
         use_cache=False if training_args.gradient_checkpointing else model_args.use_cache,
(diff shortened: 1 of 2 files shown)


vivekgoe:
@regisss Fix for an issue we see with BART inference & Synapse 1.15.0 (which has torch == 2.2.0), can be merged without waiting for 1.15.0 as the fix is backward compatible.

libinta (approved):

vivekgoe:
@jiminha @libinta I am closing this pull request, please add BART as part of generic fix you are making for SDPA issue.

PR #753, opened 2024-03-03
Set _use_sdpa to False to disable torch sdpa

hlahkar:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #761, opened 2024-03-05
Add Llama test for fsdp

hlahkar:
Add Llama test for FSDP

### DIFF
--- a/examples/language-modeling/README.md
+++ b/examples/language-modeling/README.md
@@ -559,6 +559,46 @@ python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
   --use_flash_attention True
 ```
 
+- Multi-card finetuning of Llama2-70B with FSDP and LoRA:
+
+```bash
+PT_HPU_MAX_COMPOUND_OP_SIZE=10 DEEPSPEED_HPU_ZERO3_SYNC_MARK_STEP_REQUIRED=1 \
+python3 ../gaudi_spawn.py --use_mpi  --world_size 8  run_lora_clm.py \
+  --model_name_or_path meta-llama/Llama-2-70b-hf \
+  --dataset_name tatsu-lab/alpaca \
+  --bf16 True \
+  --output_dir ./lora_out \
+  --num_train_epochs 2 \
+  --max_seq_len 2048 \
+  --per_device_train_batch_size 10 \
+  --per_device_eval_batch_size 10 \
+  --gradient_checkpointing \
+  --evaluation_strategy epoch \
+  --eval_delay 2 \
+  --save_strategy no \
+  --learning_rate 0.0004 \
+  --warmup_ratio 0.03 \
+  --lr_scheduler_type "constant" \
+  --logging_steps 1 \
+  --dataset_concatenation \
+  --attn_softmax_bf16 True \
+  --do_train \
+  --do_eval \
+  --use_habana \
+  --use_lazy_mode False \
+  --pipelining_fwd_bwd False \
+  --throughput_warmup_steps 3 \
+  --lora_rank 4 \
+  --lora_target_modules "q_proj" "v_proj" "k_proj" "o_proj" \
+  --validation_split_percentage 4 \
+  --use_flash_attention True \
+  --fsdp_config fsdp_config.json \
+  --fsdp "auto_wrap" \
+  --torch_compile_backend hpu_backend \.
+  --torch_compile \
+  --use_fused_rope False
+```
+
 - Multi-card finetuning of Falcon-180B:
   - Falcon-180B example command saves only the LoRA parameters at end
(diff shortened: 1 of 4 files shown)


hlahkar:
@regisss please review this test case. Our end goal is to test llama2; but we need to pass a auth token for that. We would like to know whether we can pass a token to the test case; in that case we can change the test from Llama to Llama2.

examples/language-modeling/README.md
```diff
@@ -559,6 +559,47 @@ python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
...
 ```
 
+- Multi-card finetuning of Llama2-70B with FSDP and LoRA:
+
+
+```bash
+PT_HPU_LAZY_MODE=0 LOWER_LIST=ops_bf16.txt \
+python3 ../gaudi_spawn.py --use_mpi  --world_size 8  run_lora_clm.py \
+  --model_name_or_path meta-llama/Llama-2-70b-hf \
+  --deepspeed llama2_ds_zero3_config.json \
```
greg-serochi, line 569: why is the deepspeed json file being called out here?

examples/language-modeling/README.md
```diff
@@ -559,6 +559,47 @@ python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
...
+  --pipelining_fwd_bwd \
+  --throughput_warmup_steps 3 \
+  --lora_rank 4 \
+  --lora_target_modules "q_proj" "v_proj" "k_proj" "o_proj" \
+  --validation_split_percentage 4 \
+  --fsdp_config fsdp_config.json \
+  --fsdp "auto_wrap" \
+  --torch_compile_backend hpu_backend \
+  --torch_compile
+```
```
vivekgoe, line 601: FusedRoPE = false is missing.

vivekgoe (changes requested):

examples/language-modeling/README.md
hlahkar: Added

regisss:
> @regisss please review this test case. Our end goal is to test llama2; but we need to pass a auth token for that. We would like to know whether we can pass a token to the test case; in that case we can change the test from Llama to Llama2.

Yes that should be feasible. Take a look at how it is done in https://github.com/huggingface/optimum-habana/blob/main/tests/test_text_generation_example.py.
Basically your tests functions needs a `token` argument and you pass the token to pytest running:
```
pytest tests/test_fsdp_examples.py --token my_token
```
Let me know if that works.

regisss (approved):
LGTM!

I pushed a new commit to add the token arg in the CI command.

regisss:
@hlahkar When I run
```
pytest tests/test_fsdp_examples.py -v -s -k "llama" --token my_token
```
the test fails because the train loss is nan. Can you run it and see if you get the same please?

hlahkar:
@regisss on latest optimum-habana (after upgrade to transformer 4.38) this test is failing; prior to that this test was passing. Also this test need to be tested on Gaudi2 also is advisable to test this on 1.15 synapse release

regisss:
> @regisss on latest optimum-habana (after upgrade to transformer 4.38) this test is failing; prior to that this test was passing. Also this test need to be tested on Gaudi2 also is advisable to test this on 1.15 synapse release

On Synapse 1.15 does it pass?

hlahkar:
> > @regisss on latest optimum-habana (after upgrade to transformer 4.38) this test is failing; prior to that this test was passing. Also this test need to be tested on Gaudi2 also is advisable to test this on 1.15 synapse release
> 
> On Synapse 1.15 does it pass?

@regisss On Synapse 1.15 it passes with optimum-habana 1.10.4 + https://github.com/huggingface/optimum-habana/pull/753/files

But it fails with the latest optimum habana because of upgrade to transformers 4.38

regisss:
> > > @regisss on latest optimum-habana (after upgrade to transformer 4.38) this test is failing; prior to that this test was passing. Also this test need to be tested on Gaudi2 also is advisable to test this on 1.15 synapse release
> > 
> > 
> > On Synapse 1.15 does it pass?
> 
> @regisss On Synapse 1.15 it passes with optimum-habana 1.10.4 + https://github.com/huggingface/optimum-habana/pull/753/files
> 
> But it fails with the latest optimum habana because of upgrade to transformers 4.38

Okay. With Synapse 1.14 it fails even before the upgrade to Transformers 4.38.
Can you update your main branch and rebase this PR on it so that we can find a fix with the latest codebase?

regisss:
It works in fp32. There must be an operation between a bf16 and a fp32 somewhere that leads to nan.

hlahkar:
@regisss I tested the test case on top of latest optimum-habana + https://github.com/huggingface/optimum-habana/pull/812; the test is passing

PR #50, opened 2022-05-16
Update the repo with changes brought by Transformers 4.19.1

regisss:
# What does this PR do?

Make the repo up to date with Transformers 4.19.1.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #762, opened 2024-03-05
Enable Llama2 70B to run with hqt on single card (#50)

HolyFalafel:
Add disk_offload flag that controls device_map=auto. Setting this flag enbales weights offload to disk when cpu memory runs OOM.
Add const serialization path flag that gets a path for where to serialize const sections, so if there is no space on device to save all const sections they will be offloaded to disk.

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -98,12 +98,9 @@ def setup_distributed(args):
 
 def setup_quantization(args, model):
     import habana_frameworks.torch.core as htcore
-    from habana_frameworks.torch.core.quantization import _check_params_as_const, _mark_params_as_const
     from habana_frameworks.torch.hpu import hpu
 
     print("Initializing inference with quantization")
-    _mark_params_as_const(model)
-    _check_params_as_const(model)
     if not args.quant_config:
         hpu.enable_quantization()
     htcore.hpu_initialize(model)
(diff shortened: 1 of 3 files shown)


examples/text-generation/utils.py
```diff
@@ -98,12 +98,9 @@ def setup_distributed(args):
 
 def setup_quantization(args, model):
     import habana_frameworks.torch.core as htcore
-    from habana_frameworks.torch.core.quantization import _check_params_as_const, _mark_params_as_const
     from habana_frameworks.torch.hpu import hpu
 
     print("Initializing inference with quantization")
-    _mark_params_as_const(model)
-    _check_params_as_const(model)
```
libinta, line 106: why these 2 are not needed anymore?

libinta (commented):
@HolyFalafel  can you provide a cmd in readme for  run with hqt on single card?

examples/text-generation/utils.py
HolyFalafel: We're using hpu_initialize(model) instead, which calls those functions

libinta (commented):
Please add how to run hqt on single card in README

regisss (commented):
Agree with Libin, please provide an example.

Can you also rebase your branch on main and run the following please?
```
pip install -U ruff
make style
```

PR #771, opened 2024-03-06
Override sdpa option in Gaudi

jiminha:
From torch 2.1 and above, SDPA attention is enabled by default. Due to this some models that we enabled are failing due to SDPA attention layer is not comparable with our code. Affected models are bart,gpt_bigcode,mistral and mixtral

 TypeError: GPTBigCodeSdpaAttention.forward() got an unexpected keyword argument 'token_idx'
TypeError: BartSdpaAttention.forward() got an unexpected keyword argument 'token_idx'

We are overriding check sdpa method for Gaudi for these model to disable sdpa until we have fusedSDPA support for them. All other models will follow original implementation. 
https://github.com/huggingface/transformers/blob/v4.37.2/src/transformers/modeling_utils.py#L1542

### DIFF
--- a/optimum/habana/transformers/modeling_utils.py
+++ b/optimum/habana/transformers/modeling_utils.py
@@ -72,6 +72,7 @@ from .models import (
     gaudi_bloom_convert_to_bloom_cache,
     gaudi_bloom_convert_to_standard_cache,
     gaudi_bloom_model_forward,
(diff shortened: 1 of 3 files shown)


regisss (approved):
LGTM!

This can be merged now I think no?

PR #773, opened 2024-03-07
Update ckpt loading

schoi-habana:
PR#15 reads a set of ckpt file names from the index json file.
When OH downloads files from the hub instead of loading from a cache dir, get_repo_root() skips downloading the index json file. Thus the PR#15 fails to load file names.
This PR scans the path and returns a list of names that matches the pattern

Tested the models in cache dir:
Falcon-7b, Falcon-40b, Falcon-180b, Llama-70b

Tested the models downloaded from HF hub:
Falcon-7b, Falcon-40b, TinyLlama

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #777, opened 2024-03-08
Enable Falcon FP8 inference

schoi-habana:
falcon-7b, falcon-40b, falcon-180b quantization enabled with --reuse_cache

dependent to #773, #831 and deepspeed v1.15.0

## Before submitting


optimum/habana/transformers/models/falcon/modeling_falcon.py
```diff
@@ -373,8 +649,17 @@ class GaudiFalconModel(FalconModel):
     - set past_key_values_length=0 when token_idx is used (with static input shape)
     - add new arg tgt_len to _expand_mask because past_key_values_length is no longer valid with token_idx
     - use old version of _make_causal_mask to workaround toch.triu that is not supported in Synapse
```
ssarkar2, line 651: wondering, triu isnt supported but tril is? because tril is used in ScaledDotProductAttention
Or maybe this comment is old

schoi-habana: you are right, the usage of '_make_causal_mask' is removed when transformers was upgraded to 4.27.0. i will remove old comments

tests/test_text_generation_example.py
```diff
@@ -117,7 +128,16 @@ def _test_text_generation(
         pattern = re.compile(r"([\"\'].+?[\"\'])|\s")
         command = [x for y in command for x in re.split(pattern, y) if x]
 
-        proc = subprocess.run(command, env=env_variables)
+        if fp8:
+            os.environ["QUANT_CONFIG"] = os.path.join(
+                path_to_example_dir, "text-generation/quantization_config/maxabs_measure_include_outputs.json"
```
libinta, line 133: we can't have env set this way as it breaks all other test. why didn't use env_variable?

regisss: I agree, we should update `env_variables` here

optimum/habana/transformers/models/falcon/modeling_falcon.py
```diff
@@ -111,257 +130,526 @@ def gaudi_falcon_attention_split_heads(
...
+
+        if alibi is None:
+            cos, sin = self.rotary_emb(value_layer, seq_len=kv_seq_len)
+            query_layer, key_layer = apply_customized_rope(query_layer, key_layer, cos, sin, position_ids)
+
+        if use_cache:
+            if self.training:
+                if layer_past is not None:
+                    key_layer = update(layer_past[0], key_layer, -2, token_idx, self.inp_seq_len)
+                    value_layer = update(layer_past[1], value_layer, -2, token_idx, self.inp_seq_len)
```
regisss, line 320: `token_idx` is supposed to be None at training no?

optimum/habana/transformers/models/falcon/modeling_falcon.py
```diff
@@ -111,257 +130,526 @@ def gaudi_falcon_attention_split_heads(
...
+                attention_mask=attention_mask,
+                position_ids=position_ids,
+                alibi=alibi,
+                head_mask=head_mask,
+                use_cache=use_cache,
+                output_attentions=output_attentions,
+                token_idx=token_idx,
+                reuse_cache=reuse_cache,
+                cache_idx=cache_idx,
+            )
```
regisss, line 650: What's the difference between these two cases? It's the same call right?

schoi-habana: you are right, removed redundant

optimum/habana/transformers/models/falcon/modeling_falcon.py
schoi-habana: actually token_idx and layer_past are None during training, so i directly set present=None for training.

regisss (approved):

PR #778, opened 2024-03-08
Fixing tests by making static_shapes False

bhargaveede:
# What does this PR do?

Fixes # (issue)

## Before submitting


vivekgoe (approved):

regisss:
@bhargaveede @vivekgoe Can you just explain quickly why a piece of code was removed in `generation/utils.py` please? That way it will be documented somewhere if someone wants to understand this change in the future.

bhargaveede:
@regisss 
This code is added for enabling static_shapes in T5.
get_extended_attention_mask  (https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/models/t5/modeling_t5.py#L368) is fulfilling the need. This is redundant and we don't need this now

PR #780, opened 2024-03-11
Enable Llama2 70B to run with hqt on single card (#50)

HolyFalafel:
Add disk_offload flag that controls device_map=auto. Setting this flag enbales weights offload to disk when cpu memory runs OOM.
Add const serialization path flag that gets a path for where to serialize const sections, so if there is no space on device to save all const sections they will be offloaded to disk.

This branch replaces the branch from PR #762

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -384,9 +384,11 @@ def initialize_model(args, logger):
 
     if args.const_serialization_path:
         import uuid
-        args.const_serialization_path = os.path.join(args.const_serialization_path  + uuid.uuid4().hex)
+
+        args.const_serialization_path = os.path.join(args.const_serialization_path + uuid.uuid4().hex)
         os.makedirs(args.const_serialization_path)
         from habana_frameworks.torch.hpu import enable_const_section_serialization
+
         print("Serializing const params to {}".format(args.const_serialization_path))
         enable_const_section_serialization(args.const_serialization_path, False, True)
     if args.fp8:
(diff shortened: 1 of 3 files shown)


HolyFalafel:
> Agree with Libin, please provide an example.
> 
> Can you also rebase your branch on main and run the following please?
> 
> ```
> pip install -U ruff
> make style
> ```

We'll update this

regisss:
@HolyFalafel I guess this PR wouldn't work with Synapse 1.14 right?

HolyFalafel:
> @HolyFalafel I guess this PR wouldn't work with Synapse 1.14 right?

You are right, v1.14 doesn't support some of the changes done here, and llama 70B won't be able to run on a single card

examples/text-generation/utils.py
```diff
@@ -373,13 +370,27 @@ def initialize_model(args, logger):
...
+
+    if args.const_serialization_path:
+        import uuid
+
+        args.const_serialization_path = os.path.join(args.const_serialization_path + uuid.uuid4().hex)
+        os.makedirs(args.const_serialization_path)
+        from habana_frameworks.torch.hpu import enable_const_section_serialization
+
+        print("Serializing const params to {}".format(args.const_serialization_path))
+        enable_const_section_serialization(args.const_serialization_path, False, True)
```
hsubramony, line 393: this api looks to be changed in the latest builds,
was seeing this error 
TypeError: enable_const_section_serialization() takes 2 positional arguments but 3 were given

HolyFalafel: You are correct, fixed

libinta (approved):

regisss (approved):

PR #782, opened 2024-03-11
text generation pipeline: Extended functionality to align with run_generation script

mgonchar:
- added batched input support
- added support of iterations
- improved output and throughput calculation

# What does this PR do?

This aligns options utilized by run_pipeline.py with those in run_generation.py

Since these two scripts share same argument parser it is quite expected that all command line options should work the same for both scripts, which is not true now. Options like batch_size and n_iterations were not supported by run_pipeline.py, but these are quite valuable options for performance measurements

## Before submitting


mgonchar:
@regisss can you give your review here?

mgonchar:
hi @regisss sorry to ping you again, but I have dependent langchain example commit, which depends on this PR. Could you please give your review?

PR #772, opened 2024-03-06
Stable diffusion XL change

yeonsily:
Also changed pipeline stable diffusion as DDPMScheduler has no reset_timestep_dependent_params.
Removes LoRA from the SDXL example.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #787, opened 2024-03-12
Sarkar/mediapipe sdxl

ssarkar2:
# What does this PR do?

1. Add stable diffusion XL model (like  https://github.com/huggingface/optimum-habana/pull/772)
2. Adds media pipe for SDXL (for 8x perf)
3. Adds flag to discount checkpoint saving time in throughput calculation `--adjust_throughput`

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/stable-diffusion/requirements.txt
+++ b/examples/stable-diffusion/requirements.txt
@@ -1,2 +1,1 @@
 opencv-python
-imagesize
--- a/examples/stable-diffusion/requirements.txt
+++ b/examples/stable-diffusion/requirements.txt
@@ -1,1 +1,2 @@
 opencv-python
+imagesize
--- a/examples/stable-diffusion/training/README.md
+++ b/examples/stable-diffusion/training/README.md
@@ -253,5 +253,5 @@ PT_HPU_MAX_COMPOUND_OP_SIZE=5 python train_text_to_image_sdxl.py \
   --bf16
 ```
 
-> [!NOTE] There is a known issue that in the first 2 steps, graph compilation takes longer than 10 seconds. This will be fixed in a future release.
-> [!NOTE] `--mediapipe` only works on Gaudi2.
+> **_NOTE:_** There is a known issue that in the first 2 steps, graph compilation takes longer than 10 seconds. This will be fixed in a future release.
+> **_NOTE:_** `--mediapipe` only works on Gaudi2.
--- a/examples/stable-diffusion/training/README.md
+++ b/examples/stable-diffusion/training/README.md
@@ -92,7 +92,7 @@ python train_controlnet.py \
  --train_batch_size=4 \
  --throughput_warmup_steps=3 \
  --use_hpu_graphs \
- --bf16 \
+ --bf16
 ```
 
 ### Multi-card Run
--- a/examples/stable-diffusion/training/README.md
+++ b/examples/stable-diffusion/training/README.md
@@ -210,3 +210,4 @@ PT_HPU_MAX_COMPOUND_OP_SIZE=5 python train_text_to_image_sdxl.py \
   --bf16
 ```
 
+**Note:** There is a known issue that first 2 steps graph compilation takes long.
--- a/examples/stable-diffusion/training/README.md
+++ b/examples/stable-diffusion/training/README.md
@@ -192,7 +192,7 @@ python train_text_to_image_sdxl.py \
 ### Multi-card Training
 ```bash
 PT_HPU_RECIPE_CACHE_CONFIG=/tmp/stdxl_recipe_cache,True,1024  \
-python ../gaudi_spawn.py --world_size 8 --use_mpi train_text_to_image_sdxl.py \
+python ../../gaudi_spawn.py --world_size 8 --use_mpi train_text_to_image_sdxl.py \
   --pretrained_model_name_or_path stabilityai/stable-diffusion-xl-base-1.0 \
   --pretrained_vae_model_name_or_path stabilityai/sdxl-vae \
   --dataset_name lambdalabs/pokemon-blip-captions \
--- a/examples/stable-diffusion/training/README.md
+++ b/examples/stable-diffusion/training/README.md
@@ -92,7 +92,7 @@ python train_controlnet.py \
  --train_batch_size=4 \
  --throughput_warmup_steps=3 \
  --use_hpu_graphs \
- --bf16
+ --bf16 \
 ```
 
 ### Multi-card Run
@@ -185,7 +185,8 @@ python train_text_to_image_sdxl.py \
   --validation_prompt="a robotic cat with wings" \
   --validation_epochs 48 \
   --checkpointing_steps 2500 \
-  --logging_step 10
+  --logging_step 10 \
+  --adjust_throughput
 ```
 
 
@@ -215,7 +216,10 @@ python ../../gaudi_spawn.py --world_size 8 --use_mpi train_text_to_image_sdxl.py
   --use_hpu_graphs_for_training \
   --use_hpu_graphs_for_inference \
   --validation_prompt="a robotic cat with wings" \
-  --validation_epochs 48
+  --validation_epochs 48 \
+  --checkpointing_steps 336 \
+  --mediapipe dataset_sdxl_pokemon \
+  --adjust_throughput
 ```
 
 ### Single-card Training on Gaudi1
--- a/examples/stable-diffusion/training/README.md
+++ b/examples/stable-diffusion/training/README.md
@@ -115,7 +115,7 @@ image.save("cat-backpack.png")
 ```
 
 
-## Fine-Tuning
+## Fine-Tuning for SDXL
 
 The `train_text_to_image_sdxl.py` script shows how to implement the fine-tuning of Stable Diffusion models on Habana Gaudi.
 
@@ -126,15 +126,74 @@ Install the requirements:
 pip install -r requirements.txt
 ```
 
-### Example for SDXL
-We can launch the fine-tuning of SDXL model using:
+### Single-card Training
 
 ```bash
 python train_text_to_image_sdxl.py \
   --pretrained_model_name_or_path stabilityai/stable-diffusion-xl-base-1.0 \
   --pretrained_vae_model_name_or_path stabilityai/sdxl-vae \
   --dataset_name lambdalabs/pokemon-blip-captions \
-  --resolution 1024 \
+  --resolution 512 \
+  --crop_resolution 512 \
+  --center_crop \
+  --random_flip \
+  --proportion_empty_prompts=0.2 \
+  --train_batch_size 16 \
+  --max_train_steps 2500 \
+  --learning_rate 1e-05 \
+  --max_grad_norm 1 \
+  --lr_scheduler constant \
+  --lr_warmup_steps 0 \
+  --output_dir sdxl-pokemon-model \
+  --gaudi_config_name Habana/stable-diffusion \
+  --throughput_warmup_steps 3 \
+  --dataloader_num_workers 8 \
+  --bf16 \
+  --use_hpu_graphs_for_training \
+  --use_hpu_graphs_for_inference \
+  --validation_prompt="a robotic cat with wings" \
+  --validation_epochs 48 \
+  --checkpointing_steps 2500 \
+  --logging_step 10
+```
+
+
+### Multi-card Training
+```bash
+PT_HPU_RECIPE_CACHE_CONFIG=/tmp/stdxl_recipe_cache,True,1024  \
+python ../gaudi_spawn.py --world_size 8 --use_mpi train_text_to_image_sdxl.py \
+  --pretrained_model_name_or_path stabilityai/stable-diffusion-xl-base-1.0 \
+  --pretrained_vae_model_name_or_path stabilityai/sdxl-vae \
+  --dataset_name lambdalabs/pokemon-blip-captions \
+  --resolution 512 \
+  --crop_resolution 512 \
+  --center_crop \
+  --random_flip \
+  --proportion_empty_prompts=0.2 \
+  --train_batch_size 16 \
+  --max_train_steps 336 \
+  --learning_rate 1e-05 \
+  --max_grad_norm 1 \
+  --lr_scheduler constant \
+  --lr_warmup_steps 0 \
+  --output_dir sdxl-pokemon-model \
+  --gaudi_config_name Habana/stable-diffusion \
+  --throughput_warmup_steps 3 \
+  --dataloader_num_workers 8 \
+  --bf16 \
+  --use_hpu_graphs_for_training \
+  --use_hpu_graphs_for_inference \
+  --validation_prompt="a robotic cat with wings" \
+  --validation_epochs 48
+```
+
+### Single-card Training on Gaudi1
+```bash
+PT_HPU_MAX_COMPOUND_OP_SIZE=5 python train_text_to_image_sdxl.py \
+  --pretrained_model_name_or_path stabilityai/stable-diffusion-xl-base-1.0 \
+  --pretrained_vae_model_name_or_path stabilityai/sdxl-vae \
+  --dataset_name lambdalabs/pokemon-blip-captions \
+  --resolution 512 \
   --center_crop \
   --random_flip \
   --proportion_empty_prompts=0.2 \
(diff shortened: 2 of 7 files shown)


examples/stable-diffusion/training/README.md
```diff
@@ -146,7 +146,7 @@ image.save("cat-backpack.png")
 ```
 
 
-## Fine-Tuning
+## Fine-Tuning for SDXL
```
libinta, line 149: change to full name

examples/stable-diffusion/training/README.md
```diff
@@ -157,86 +157,92 @@ Install the requirements:
...
-  use_hpu_graphs=True,
-  gaudi_config="Habana/stable-diffusion",
-)
-pipe.load_lora_weights(lora_model_id)
-
-prompt = "cute dragon creature"
-image = pipe(prompt).images[0]
-image.save("green-pokemon.png")
-```
+**Note:** There is a known issue that first 2 steps graph compilation takes long.
```
libinta, line 248: > 10 seconds

ssarkar2 (commented):
run style
~expand SDXL in readme~
~do dataset dump only from main process~
~known issue: takes longer than 10 seconds. And it will be fixed in future release.~
check is PT_HPU_RECIPE_CACHE_CONFIG can be removed

examples/stable-diffusion/training/train_text_to_image_sdxl.py
```diff
@@ -859,12 +789,26 @@ def main(args):
     # In distributed training, the load_dataset function guarantees that only one local process can concurrently
     # download the dataset.
     if args.dataset_name is not None:
-        # Downloading and loading a dataset from the hub.
-        dataset = load_dataset(
-            args.dataset_name,
-            args.dataset_config_name,
-            cache_dir=args.cache_dir,
-        )
+        if len(args.mediapipe) > 0:
```
libinta, line 792: the whole section use rank 0 to process

examples/stable-diffusion/training/train_text_to_image_sdxl.py
```diff
@@ -1547,102 +1440,79 @@ def compute_time_ids(original_size, crops_coords_top_left):
...
-    throughput = args.max_train_steps * total_batch_size / duration
+    duration = time.perf_counter() - t0 - (checkpoint_time if args.adjust_throughput else 0)
+    ttt = time.perf_counter() - t_start
+    throughput = (args.max_train_steps - args.throughput_warmup_steps) * total_batch_size / duration
 
     accelerator.wait_for_everyone()
     if accelerator.is_main_process:
         logger.info(f"Throughput = {throughput} samples/s")
         logger.info(f"Train runtime = {duration} seconds")
+        logger.info(f"Total Train runtime = {ttt} seconds")
```
regisss, line 1451: Not very clear, I would just keep the TTT

examples/stable-diffusion/training/train_text_to_image_sdxl.py
```diff
@@ -364,11 +330,6 @@ def parse_args(input_args=None):
         default=1,
         help="Number of updates steps to accumulate before performing a backward/update pass.",
     )
-    parser.add_argument(
-        "--gradient_checkpointing",
-        action="store_true",
-        help="Whether or not to use gradient checkpointing to save memory at the expense of slower backward pass.",
-    )
```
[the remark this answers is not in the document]
libinta: we had issue with gradient checkpointing now. So disable it for now

examples/stable-diffusion/training/train_text_to_image_sdxl.py
```diff
@@ -1280,58 +1179,48 @@ def unwrap_model(model):
...
-            if args.train_text_encoder:
-                text_encoder_one.train()
-                text_encoder_two.train()
-        train_loss = 0.0
+        train_loss.zero_()
+        if hb_profiler:
+            hb_profiler.start()
         for step, batch in enumerate(train_dataloader):
-            if t0 is None:  # and global_step == args.throughput_warmup_steps:
+            if t0 is None or global_step == args.throughput_warmup_steps:
```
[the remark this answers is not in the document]
libinta: if it's and, then if some one only test for 1 step and througput_warmup_steps still as default 3, then there is issue for metrics calculation as it's minus None

examples/stable-diffusion/training/train_text_to_image_sdxl.py
libinta: As we had long graph compilation for R1.15 as known issue, if we keep just TTT, we don't show the train time excluding graph compilation time. We will remove the train runtime after the graph compilation issue is fixed in R1.16. WDYT?

examples/stable-diffusion/training/train_text_to_image_sdxl.py
```diff
@@ -859,12 +848,31 @@ def main(args):
...
+            if args.local_rank == 0:
+                if not os.path.exists(args.mediapipe):
+                    os.mkdir(args.mediapipe)
+                if len(os.listdir(args.mediapipe)) == 0:
+                    dataset = load_dataset(args.dataset_name, None)
+                    with open(f"{args.mediapipe}/label.txt", "w") as f:
+                        for idx, dt in enumerate(dataset["train"]):
+                            dt["image"].save(f"{args.mediapipe}/{idx}.jpg")
+                            f.write(dt["text"] + "\n")
+            torch.distributed.barrier()
```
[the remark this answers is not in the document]
libinta: for 1x, we are not using mediapipeline as it's device bounded. But we can check for multi-card here

examples/stable-diffusion/training/train_text_to_image_sdxl.py
```diff
@@ -933,95 +954,58 @@ def preprocess_train(examples):
...
-                "pooled_prompt_embeds": pooled_prompt_embeds,
-                "original_sizes": original_sizes,
-                "crop_top_lefts": crop_top_lefts,
-            }
+    with accelerator.main_process_first():
+        from datasets.fingerprint import Hasher
+
+        # fingerprint used by the cache for the other processes to load the result
+        # details: https://github.com/huggingface/diffusers/pull/4038#discussion_r1266078401
+        new_fingerprint = Hasher.hash(args)
```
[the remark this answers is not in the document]
libinta: right, the reason to move VAE to device is because it's slow on cpu. Also, the original code was using cache that  is not right as our images are transformed.

examples/stable-diffusion/training/train_text_to_image_sdxl.py
```diff
@@ -301,7 +260,16 @@ def parse_args(input_args=None):
...
+            " resolution."
+        ),
+    )
+    parser.add_argument(
+        "--crop_resolution",
+        type=int,
+        default=1024,
+        help=(
+            "The resolution for cropping input images, all the images in the train/validation dataset will be resized to this"
+            " resolution."
```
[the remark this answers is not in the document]
libinta: we only do crop only if the crop_resolution is less than resolution. See preprocess_trainining

examples/stable-diffusion/requirements.txt
```diff
@@ -1 +1,2 @@
 opencv-python
+imagesize
```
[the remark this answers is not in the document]
ssarkar2: If we are using mediapipe, the network still needs "original input sizes". So using this library to get that, without reading the image in python:
https://github.com/huggingface/optimum-habana/blob/01bb2bddc04d2be524acbf38a6619add9d0b5b63/examples/stable-diffusion/training/train_text_to_image_sdxl.py#L984

examples/stable-diffusion/training/train_text_to_image_sdxl.py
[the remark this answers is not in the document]
regisss: Okay. Shouldn't it be `args.crop_resolution` instead of `args.resolution` here
```py
train_crop = transforms.CenterCrop(args.resolution) if args.center_crop else transforms.RandomCrop(args.resolution)
```
?

examples/stable-diffusion/training/train_text_to_image_sdxl.py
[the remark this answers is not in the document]
regisss: If t0 is None we shouldn't compute the throughput as it doesn't make much sense to do it. With "or", `throughput_warmup_steps` is useless and we will always compute a lower throughput than the real one.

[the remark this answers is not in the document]
libinta: this is found when I tried to take profiler for couple of steps. and t0 is None that causes crash at the end. It's not real use case for performance calculations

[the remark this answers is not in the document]
regisss: Yes I understand, that's why I propose to:
- keep and, otherwise `throughpuit_warmup_steps` is useless
- if t0 is None, we don't compute and log the throughput

examples/stable-diffusion/training/train_text_to_image_sdxl.py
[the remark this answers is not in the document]
ssarkar2: @regisss 
The original code is from [here](https://github.com/huggingface/diffusers/blob/01ac37b331d87368741d5f9554eaf47e0dd05a59/examples/text_to_image/train_text_to_image_sdxl.py#L842).

I think it should have been `args.crop_resolution` in the original as well, but that might be a bug. Should we fix it to `args.crop_resolution` or keep it aligned to original?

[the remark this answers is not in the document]
regisss: Yes, I also think it's a mistake in the original example. Let's keep it that way then, I'll ask the Diffusers team about it.

examples/stable-diffusion/training/train_text_to_image_sdxl.py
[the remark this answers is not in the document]
ssarkar2: uploaded change

examples/stable-diffusion/training/train_text_to_image_sdxl.py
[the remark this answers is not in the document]
regisss: So actually it's not a bug. The crops are squared images whereas images are resized with the smaller side equal to `args.resolution` and the other proportionally resized.

tests/test_diffusers.py
```diff
@@ -1776,19 +1776,20 @@ def test_train_text_to_image_sdxl(self):
...
+                 --train_batch_size 16
                  --learning_rate 1e-05
                  --max_grad_norm 1
                  --lr_scheduler constant
                  --lr_warmup_steps 0
                  --gaudi_config_name Habana/stable-diffusion
                  --throughput_warmup_steps 3
-                 --use_hpu_graphs
+                 --dataloader_num_workers 8
+                 --use_hpu_graphs_for_training
```
regisss, line 1792: We could also add `--use_hpu_graphs_for_inference` no?

examples/stable-diffusion/training/README.md
```diff
@@ -163,86 +163,92 @@ Install the requirements:
...
+### Single-card Training
 
 ```bash
 python train_text_to_image_sdxl.py \
   --pretrained_model_name_or_path stabilityai/stable-diffusion-xl-base-1.0 \
   --pretrained_vae_model_name_or_path stabilityai/sdxl-vae \
   --dataset_name lambdalabs/pokemon-blip-captions \
-  --resolution 1024 \
+  --resolution 512 \
+  --crop_resolution 512 \
```
regisss, line 174: If we use the same value, maybe `crop_resolution` is not needed. Or do we want to give users the possibility to specify different crop sizes?
I replied in the other comment thread and actually cropping with `args.resolution` return a square image while resizing will keep the aspect ratio. Maybe the mediapipe has a different behavior?

examples/stable-diffusion/training/README.md
```diff
@@ -163,86 +163,92 @@ Install the requirements:
...
+  --pretrained_model_name_or_path stabilityai/stable-diffusion-xl-base-1.0 \
+  --pretrained_vae_model_name_or_path stabilityai/sdxl-vae \
+  --dataset_name lambdalabs/pokemon-blip-captions \
+  --resolution 512 \
+  --crop_resolution 512 \
+  --center_crop \
+  --random_flip \
+  --proportion_empty_prompts=0.2 \
+  --train_batch_size 16 \
+  --max_train_steps 336 \
```
regisss, line 212: Does 336 steps correspond to one or several epochs?

examples/stable-diffusion/training/README.md
```diff
@@ -163,86 +163,92 @@ Install the requirements:
...
-  use_hpu_graphs=True,
-  gaudi_config="Habana/stable-diffusion",
-)
-pipe.load_lora_weights(lora_model_id)
-
-prompt = "cute dragon creature"
-image = pipe(prompt).images[0]
-image.save("green-pokemon.png")
-```
+**Note:** There is a known issue that in the first 2 steps, graph compilation takes longer than 10 seconds. This will be fixed in a future release.
```
regisss, line 254:
```suggestion
> [!NOTE]
> There is a known issue that in the first 2 steps, graph compilation takes longer than 10 seconds. This will be fixed in a future release.
```

examples/stable-diffusion/training/README.md
```diff
@@ -163,86 +163,92 @@ Install the requirements:
...
+  --train_batch_size 1 \
+  --gradient_accumulation_steps 4 \
+  --max_train_steps 3000 \
+  --learning_rate 1e-05 \
+  --max_grad_norm 1 \
+  --lr_scheduler constant \
+  --lr_warmup_steps 0 \
+  --output_dir sdxl-pokemon-model \
+  --gaudi_config_name Habana/stable-diffusion \
+  --throughput_warmup_steps 3 \
```
regisss, line 250: Maybe we can add `--use_hpu_graphs_for_inference` for Gaudi1?

examples/stable-diffusion/training/README.md
ssarkar2: here is the original code:
https://github.com/huggingface/diffusers/blob/main/examples/text_to_image/train_text_to_image_sdxl.py#L846

this resizes to `resolution`, and then crops, also to `resolution`. Thus effectively, the crop doesnt do anything. The crop resolution should be < resolution for something to happen. This we think is a bug in the original code

To get around that we also add "crop_resolution". The idea being, first resize to "resolution" and then center crop or random crop to "crop_resolution". In the command line we keep both same, so that we have same behaviour as original diffuser code, where resolution and crop_resolution are same.

So how do you want this code to be like?
1. Go back to diffuser code, with one argument "resolution", in that case I think we should just remove the crop, because it does nothing.
2. Keep both, such that there is a possibility of random crop or center crop.

examples/stable-diffusion/training/README.md
ssarkar2: several epochs

tests/test_diffusers.py
```diff
@@ -1776,19 +1776,21 @@ def test_train_text_to_image_sdxl(self):
                  --pretrained_model_name_or_path stabilityai/stable-diffusion-xl-base-1.0
                  --pretrained_vae_model_name_or_path stabilityai/sdxl-vae
                  --dataset_name lambdalabs/pokemon-blip-captions
```
ssarkar2, line 1778: @regisss : 
All tests that use `--dataset_name lambdalabs/pokemon-blip-captions` probably should bedisabled. The dataset has been taken down

We need to figure out a different dataset for the next release

What's the best way to disable these tests?

examples/stable-diffusion/training/README.md
regisss: The behavior is actually a bit different. [The doc of resize](https://pytorch.org/vision/main/generated/torchvision.transforms.Resize.html) says that "If size is an int, smaller edge of the image will be matched to this number. i.e, if height > width, then image will be rescaled to (size * height / width, size)." So if images are not square, only the smaller edge will be equal to `resolution`.

I would tend to stick to Diffusers code, but if having another argument for the crop resolution is better for you, then let's keep it.

examples/stable-diffusion/training/README.md
regisss: Okay, but a specific number of epochs? If yes, it would be better to use `--num_train_epochs`

examples/stable-diffusion/training/README.md
```diff
@@ -163,86 +163,95 @@ Install the requirements:
...
-  gaudi_config="Habana/stable-diffusion",
-)
-pipe.load_lora_weights(lora_model_id)
-
-prompt = "cute dragon creature"
-image = pipe(prompt).images[0]
-image.save("green-pokemon.png")
-```
+> **_NOTE:_** There is a known issue that in the first 2 steps, graph compilation takes longer than 10 seconds. This will be fixed in a future release.
+> **_NOTE:_** `--mediapipe` only works on Gaudi2.
```
regisss, line 257:
```suggestion
> [!NOTE]
> There is a known issue that in the first 2 steps, graph compilation takes longer than 10 seconds. This will be fixed in a future release.

> [!NOTE]
> `--mediapipe` only works on Gaudi2.
```
This renders better

tests/test_diffusers.py
regisss: Yes, I saw that too. Not sure exactly what the best way of disabling them is, I'll do it in another :+1:

examples/stable-diffusion/training/README.md
ssarkar2: Hmm i see what you mean. However we have difficulties retesting it now that the dataset is no longer available. I think once we figure out a new dataset in the next release, I can come back to this and clean up reso/crop_reso

regisss: Sounds good, let's do that

regisss (approved):

PR #795, opened 2024-03-13
move img_mask@get_attn_mask() to hpu

hsubramony:
In certain 1x systems , we notice a drop in performance for swin models as get_attn_mask gets executed in cpu instead of hpu. This was seen after pytorch upgrade to 2.2
This fix allows img_mask in get_attn_mask() to be moved back hpu.


libinta (approved):

regisss (approved):
LGTM!

Issue #111, opened 2022-10-05
Add a utility method to get the memory consumptions for various batch sizes


regisss:
### Feature request

The `GaudiTrainer` class should provide a method that takes a list of batch sizes as argument and returns the memory consumptions on HPU for each batch size.
For each batch size in the list, a training of 5 steps should be run without logging anything and the maximum memory consumption of this run should be returned.

### Motivation

This feature will avoid users launching a full training just to check the memory consumption in the logs.

### Your contribution

I will submit a PR.

PR #802, opened 2024-03-14
Add --bucket_size support for gpt_bigcode #111

jiminha:
# What does this PR do?
Currently, bucket_size is only supported for a few models such as llama and falcon. This PR is to add bucket_size support for gpt_bigcode model , this can be useful if the dataset has varying prompt lengths.


ssarkar2 (approved):
failing style. lgtm otherwise

regisss (approved):

PR #807, opened 2024-03-15
Add warmup step feature support for inference

ANSHUMAN87:
This PR add support for warm up steps consideration during inference.
Issue was seen in BERT text classification, the inference perf was unstable between runs in same device.
This PR handles the issue.

JH...!

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -1621,6 +1621,110 @@ class GaudiTrainer(Trainer):
         # Good practice: save your training arguments together with the trained model
         torch.save(self.args, os.path.join(output_dir, TRAINING_ARGS_NAME))
 
+    def evaluate(
+        self,
+        eval_dataset: Optional[Union[Dataset, Dict[str, Dataset]]] = None,
+        ignore_keys: Optional[List[str]] = None,
+        metric_key_prefix: str = "eval",
+    ) -> Dict[str, float]:
+        """
+        Run evaluation and returns metrics.
+
+        The calling script will be responsible for providing a method to compute metrics, as they are task-dependent
+        (pass it to the init `compute_metrics` argument).
+
+        You can also subclass and override this method to inject custom behavior.
+
+        Args:
+            eval_dataset (Union[`Dataset`, Dict[str, `Dataset`]), *optional*):
+                Pass a dataset if you wish to override `self.eval_dataset`. If it is a [`~datasets.Dataset`], columns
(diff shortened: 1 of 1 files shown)


ANSHUMAN87:
@regisss could you please help review, Thanks!


JH...!

optimum/habana/transformers/trainer.py
```diff
@@ -1621,6 +1621,110 @@ def _save(self, output_dir: Optional[str] = None, state_dict=None):
...
+        total_batch_size = self.args.eval_batch_size * self.args.world_size
+        if f"{metric_key_prefix}_jit_compilation_time" in output.metrics:
+            start_time += output.metrics[f"{metric_key_prefix}_jit_compilation_time"]
+        num_samples = output.num_samples
+        num_steps = math.ceil(output.num_samples / total_batch_size)
+        if self.start_time_after_warmup is not None:
+            num_samples = output.num_samples - self.args.throughput_warmup_steps * total_batch_size
+            num_steps = num_steps - self.args.throughput_warmup_steps
+            output.metrics[f"{metric_key_prefix}_samples_excluding_warmup"] = num_samples
+            output.metrics[f"{metric_key_prefix}_runtime_excluding_warmup"] = (
```
regisss, line 1707: Not a big fan of having two kinds of metrics: with and without warmup. I think this will create confusion.
IMO we should just compute the adjusted throughput if `throughput_warmup_steps` is larger than 0. And we should not adjust the runtime. That would also be aligned with what we do at training.

optimum/habana/transformers/trainer.py
```diff
@@ -1621,6 +1621,110 @@ def _save(self, output_dir: Optional[str] = None, state_dict=None):
...
+
+            ignore_keys (`List[str]`, *optional*):
+                A list of keys in the output of your model (if it is a dictionary) that should be ignored when
+                gathering predictions.
+            metric_key_prefix (`str`, *optional*, defaults to `"eval"`):
+                An optional prefix to be used as the metrics key prefix. For example the metrics "bleu" will be named
+                "eval_bleu" if the prefix is "eval" (default)
+        Return:
+            Dict[str, float]: A dictionary containing the evaluation loss and the potential metrics computed from the predictions. The
+            dictionary also contains the epoch number which comes from the training state.
```
regisss, line 1665: We can remove this docstring since it is exactly the same as in Transformers, and replace it with one or two lines explaining the difference with Transformers.

regisss:
Done in #855

PR #806, opened 2024-03-15
Fix graph breaks in torch compile mode

hlahkar:
# What does this PR do?
1. Fixes Graph Breaks in torch compile mode
2. Fixes comments in https://github.com/huggingface/optimum-habana/pull/751

PR #811, opened 2024-03-17
Fix torch.compile for text generation

regisss:
# What does this PR do?

Text-generation tests with torch.compile currently fail, this PR fixes this issue.

## Before submitting


regisss:
@vivekgoe @hlahkar The two torch.compile tests for text generation (1x and 8x) now fail after #806. It seems reverting from `has_fused_rms_norm` to `FusedRMSNorm` is enough to fix it, let me know if that's okay for you.

libinta (approved):

vivekgoe:
@regisss has_fused_rms_norm will work with Synapse 1.15.0 release content. We need to revert this change back once 1.15.0 is available, without this there will be a performance drop for Llama-FSDP fine-tuning.  FYI @libinta  @hlahkar

PR #812, opened 2024-03-17
Fix FSDP in transformer4.38

libinta:
Fix FSDP ci test issue with "TypeError: Object of type Tensor is not JSON serializable"

Transformer 4.38 logs the grad_norm in log_history. But FSDP doesn't have global grad norm function. When logging non-scalar tensor, the .item fails. The solution now is not to log grad_norm in logging_history for FSDP.

# What does this PR do?

Fixes # (issue)

## Before submitting


regisss (approved):
LGTM!

PR #813, opened 2024-03-18
Fix llama rotary pos emb issue for transformers 4.38

libinta:
Transformer 4.38 changes the rotary embedding implementation, we changed rotaryembedding class to be the same as transformers4.37, but not apply_rotary_pos_emb. We should use apply_rotary_pos_emb as 4.37 too.
# What does this PR do?

Fixes # (issue)

## Before submitting


regisss (approved):
LGTM

PR #815, opened 2024-03-19
Sarkar/split model markstep

ssarkar2:
# What does this PR do?
Add an option to split the model using mark_step in the middle for inference. Might be useful for supporting higher batches.

Fixes # (issue)

## Before submitting


optimum/habana/transformers/models/mixtral/modeling_mixtral.py
```diff
@@ -494,7 +495,9 @@ def gaudi_mixtral_model_forward(
     all_router_logits = () if output_router_logits else None
     next_decoder_cache = None
 
-    for decoder_layer in self.layers:
+    for layer_idx, decoder_layer in enumerate(self.layers):
+        if split_model_markstep and (layer_idx == len(self.layers)//2) and not self.training:
+            htcore.mark_step()
```
ssarkar2, line 500: import mark_step in this file

regisss:
@ssarkar2 I think this was added in another PR right? Can we close this one?

PR #818, opened 2024-03-19
Add Llama7b FSDP test for torch.compile mode

pankd:
Adds a test for FSDP

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -164,7 +164,7 @@ class ExampleTestMeta(type):
     """
 
     @staticmethod
-    def to_test(model_name: str, multi_card: bool, deepspeed: bool, example_name: str):
+    def to_test(model_name: str, multi_card: bool, deepspeed: bool, example_name: str, fsdp: bool):
         models_with_specific_rules = [
             "albert-xxlarge-v1",
             "gpt2-xl",
@@ -198,14 +198,14 @@ class ExampleTestMeta(type):
             return True
         elif "bridgetower" in model_name and os.environ.get("GAUDI2_CI", "0") == "1":
             return True
-        elif "falcon" in model_name and os.environ.get("GAUDI2_CI", "0") == "1":
+        elif "falcon" in model_name and os.environ.get("GAUDI2_CI", "0") == "1" and not fsdp:
             return True
         elif "bloom" in model_name and deepspeed and os.environ.get("GAUDI2_CI", "0") == "0":
             return True
 
         return False
 
-    def __new__(cls, name, bases, attrs, example_name=None, multi_card=False, deepspeed=False):
+    def __new__(cls, name, bases, attrs, example_name=None, multi_card=False, deepspeed=False, fsdp=False):
         distribution = "single_card"
         if multi_card:
             distribution = "multi_card"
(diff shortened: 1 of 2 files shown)


regisss:
Is it a different test from the one that was added in https://github.com/huggingface/optimum-habana/pull/761 ?

vivekgoe:
@regisss This attempts to add generic support for FSDP (similar to what you already have for DeepSpeed) such that going forward any model that runs with DeepSpeed can also be run with FSDP.  Also I assume that tests in this file are "slow" tests and we can add longer tests here, we wanted to add models such as Llama-70B + FSDP which take longer time here.

Test in https://github.com/huggingface/optimum-habana/pull/761 is a short test which is intended to run as part of Gaudi2 CI. Going forward if you feel there is repetition we can consider removing it.

tests/test_examples.py
```diff
@@ -310,11 +315,15 @@ def test(self):
                 env_variables["DEEPSPEED_HPU_ZERO3_SYNC_MARK_STEP_REQUIRED"] = "1"
                 env_variables["PT_HPU_MAX_COMPOUND_OP_SYNC"] = "1"
                 env_variables["PT_HPU_MAX_COMPOUND_OP_SIZE"] = "1"
+            elif ("llama" in model_name) and fsdp:
```
vivekgoe, line 318: Can this be made generic so that we can use fsdp with other models in future? You can test with BERT for now, in future we may have more.

tests/test_examples.py
```diff
@@ -329,6 +338,7 @@ def test(self):
                     .get("extra_arguments", []),
                 )
 
+                print(env_variables)
```
vivekgoe, line 341: Remove print.

vivekgoe (changes requested):

vivekgoe:
@pankd please add a test for Llama-70B also.

tests/test_examples.py
```diff
@@ -310,11 +315,15 @@ def test(self):
                 env_variables["DEEPSPEED_HPU_ZERO3_SYNC_MARK_STEP_REQUIRED"] = "1"
                 env_variables["PT_HPU_MAX_COMPOUND_OP_SYNC"] = "1"
                 env_variables["PT_HPU_MAX_COMPOUND_OP_SIZE"] = "1"
+            elif fsdp:
+                env_variables["LOWER_LIST"] = str(example_script.parent / "ops_bf16.txt")
```
vivekgoe, line 319: Not every FSDP model will specify this environment variable. Can we change this check to populate this env variable only if "ops_bf16.txt" file is present in example_script.parent directory?

vivekgoe (changes requested):

tests/test_examples.py
regisss: We can also use the variable `model_name` that is available in this scope to check the model, see the `elif` right above

pankd: Change pushed to update the env on model basis.

regisss: @vivekgoe Does the new change look good to you? If yes, I'll merge this PR

vivekgoe:
@regisss we are checking last few things, will let you know when it's done.

vivekgoe:
@regisss this is ready for merge now.

regisss (approved):

PR #823, opened 2024-03-20
Update pytest version to avoid error

jiminha:
Latest habana docker has pytest 8.1.1 installed by default. 
For transformer, we need the pytest lower than 8.0.0


jiminha:
Seems the change is already in. Let me closet this PR.

PR #824, opened 2024-03-21
SpeechT5 static consistent dropout

Spycsh:
# What does this PR do?

Fixes # [722](https://github.com/huggingface/optimum-habana/pull/722)

## Before submitting


Spycsh:
Hi @regisss @sywangyi , SpeechT5 has issue with static shape and has been mostly fixed by the previous [PR](https://github.com/huggingface/optimum-habana/pull/722). However, that PR forced to disable the dropout [link](https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/models/speecht5/modeling_speecht5.py#L413) to make sure the HPU graph is correctly generated and make the latency to be OK. This is not the right way and the audio quality it generates is worse with noises.

In this PR I fixed this issue. By setting the seed for each generation we can ensure every time the `consistent_dropout` keeps the same bernouli sequence and outputs the correct audio with the original quality. The HPU graphs in the warmup/inference are the same, so the latency of the original example is perfect. This PR is also a bit faster (~10%) than the main branch because an early threshold of generation is met.

sywangyi (approved):

regisss:
> Hi @regisss @sywangyi , SpeechT5 has issue with static shape and has been mostly fixed by the previous [PR](https://github.com/huggingface/optimum-habana/pull/722). However, that PR forced to disable the dropout [link](https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/models/speecht5/modeling_speecht5.py#L413) to make sure the HPU graph is correctly generated and make the latency to be OK. This is not the right way and the audio quality it generates is worse with noises.
> 
> In this PR I fixed this issue. By setting the seed for each generation we can ensure every time the `consistent_dropout` keeps the same bernouli sequence and outputs the correct audio with the original quality. The HPU graphs in the warmup/inference are the same, so the latency of the original example is perfect. This PR is also a bit faster (~10%) than the main branch because an early threshold of generation is met.

Thanks for the clear explanation @Spycsh!
So, if I understand correctly, this is only needed when we use HPU graphs right? If yes, we can call `set_seed` under
```py
if not hasattr(model.speecht5.decoder.prenet, "clear_cache"):
```

Spycsh:
Hi @regisss , I doubt whether this is relevant to enable/not enable HPU graph. set_seed basically make sure that if an input has been already warmuped, the same input afterwards will be processed quickly. This is SpeechT5 specific  setting because it forces to use a dropout with training=True. I guess you mean there is a parameter that controls to enable HPU graph explicitly or not, right? I think both cases should use this set_seed, so the if statement is meaningless here. Maybe I'm wrong here, and could you explain a bit more about this if statement? If we do not use HPU graph explicitly, we still can use warmup to make sure the same input is handled quickly the second time, right? In that case we still need the same seed.

regisss:
@Spycsh My question basically comes down to whether calling `set_seed` is only needed in the HPU graphs case or it is needed in all configurations?
Asking this because `set_seed` is not part of the Transformers implementation and I guess it behaves similarly. What I do see in Transformers is this: https://github.com/huggingface/transformers/blob/517a3e670d8fc11374895e870dd0dd041467c7fe/src/transformers/models/speecht5/modeling_speecht5.py#L2687
So `set_seed` is called in the example, outside of the modeling code. I think we should do the same here and call it there: https://github.com/huggingface/optimum-habana/blob/eaac913c56617a4deefa99548d89366722b7397e/examples/text-to-speech/run_pipeline.py#L88
If we call it in the modeling file, it will basically override a previous call that a user may have done like in the Transformers example.

Spycsh:
Hi @regisss I agree with you that we should move this set_seed in the example file rather than in the modeling file, just like how it is used in HF transformers. Earlier @sywangyi said that this example (run_pipeline.py) may be used for other models so adding a set_seed may seem weird here. What do you think @sywangyi , maybe here we just move the set_seed in the example file?

regisss:
We could call `set_seed` only if the model that is being used is speechT5:
```py
if generator.model.config.model_type == "speecht5":
     set_seed(555)
```

sywangyi:
Hi, @regisss if we need to move the set_seed to example. it's add here https://github.com/huggingface/optimum-habana/blob/main/examples/text-to-speech/run_pipeline.py#L92. before each generator is called. and we should make sure user know this limitation for speechT5 when they write their own app.

Spycsh:
Hi @regisss @sywangyi  , I've changed move the `set_seed` into the example script and add necessary explanation. Welcome to further advices.

examples/text-to-speech/run_pipeline.py
```diff
@@ -89,10 +90,20 @@ def main():
     with torch.autocast("hpu", torch.bfloat16, enabled=args.bf16), torch.no_grad(), torch.inference_mode():
         # warm up
         for i in range(args.warmup):
+            if generator.model.config.model_type == "speecht5":
+                # SpeechT5 forces a dropout with training=True, which may zero out some elements randomly.
+                # A random dropout may need different lengths of spectrograms to fit probability thresholds,
+                # which violates the HPU static shape, so we have to fix the seed here.
+                set_seed(555)
```
libinta, line 97: only this seed works or it's just the one working after you tested

Spycsh: @libinta Any seed should work. Basically you just need to set the same seed again for each generation.

regisss (approved):

PR #833, opened 2024-03-25
Add mark step and inplace residual add in llama model code

puneeshkhanna:
Mark step helping in reducing workspace memory by approx twice of (BS,seq len, hidden dim).

Inplace add helping in reducing persistent tensors by approx twice of (BS, seq len, hidden dim).

Add lazy mode parameter.

---------

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -1430,6 +1430,7 @@ class GaudiGenerationMixin(GenerationMixin):
                 )
 
             # prepare model inputs
+            model_kwargs["lazy_mode"] = lazy_mode
             model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
 
             hpu_graphs_kwargs = self._get_hpu_graphs_kwargs(model_kwargs)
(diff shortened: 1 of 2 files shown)


puneeshkhanna:
As an example for the config of BS-172, seq len-2048, hidden dim-8192 (size is ~5.3 GB) for llama-70B on 8x.
Max memory usage (without flash attention) reduced from ~86 Gb to ~66Gb.
Max memory usage (with flash attention) reduced from ~70 GB to ~59GB.

We can further go higher batch sizes because of reduced memory consumptions with the changes in this PR hence very important changes.

@regisss  - please review and merge.

regisss (commented):
LGTM! Just need to run `make style`.

Quick question, the new `mark_step` in the forward of `GaudiLlamaModel` will also benefit training right?

puneeshkhanna:
@regisss  - mark step has no side effect on training. It should help in reducing workspace memory mainly.

puneeshkhanna:
@regisss  - make style fixed. Just required an empty line after the import statement. It should pass now.

regisss (approved):

PR #835, opened 2024-03-26
Save CI test output image

MohitIntel:
Enables saving of output images in Stable Diffusion CI test for offline analysis.

### DIFF
--- a/tests/test_diffusers.py
+++ b/tests/test_diffusers.py
@@ -28,7 +28,7 @@ import requests
 import torch
 from diffusers import AutoencoderKL, ControlNetModel, UNet2DConditionModel, UniPCMultistepScheduler
 from diffusers.pipelines.controlnet.pipeline_controlnet import MultiControlNetModel
-from diffusers.utils import load_image
+from diffusers.utils import load_image, numpy_to_pil
 from diffusers.utils.torch_utils import randn_tensor
 from huggingface_hub import snapshot_download
 from parameterized import parameterized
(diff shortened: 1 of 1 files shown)


tests/test_diffusers.py
```diff
@@ -643,6 +643,9 @@ def test_no_generation_regression(self):
                 [0.70760196, 0.7136303, 0.7000798, 0.714934, 0.6776865, 0.6800843, 0.6923707, 0.6653969, 0.6408076]
             )
         image = outputs.images[0]
+        import matplotlib.pyplot as plt
+        plt.imsave('test_no_generation_regression_output.png', image)
```
sayakpaul, line 647: If `image` is a PIL image it could be `image.save("test_no_generation_regression_output.png")`.

sayakpaul: Also, if the outputs are pure latents (and are not decoded by the subsequent VAE in the process), then the output won't make to much of a sense.

regisss: I agree with @sayakpaul. We can specify `output_type="pil"` when instantiating the pipeline and save the image as suggested.

MohitIntel: I had already tried the the conversion to PIL image like you mentioned @sayakpaul , but got a blank image instead. (Probably due to incorrect clamping)
@regisss , I am aware of the output_type="pil" but did not use that since the test is checking for raw pixel values later.

Will be trying out this diffusers numpy_to_pil function and update: https://github.com/huggingface/diffusers/blob/v0.23.1/src/diffusers/image_processor.py#L82

regisss (approved):
LGTM

PR #843, opened 2024-03-27
Add instructions for Mixtral-8x7B inference

tjs-intel:
# What does this PR do?

The Mixtral-8x7B model is too large to be loaded onto a single Gaudi2 device in 32 bits. This PR demonstrates how to run the Mixtral-8x7B model on a single card, by leveraging the `--bf16` command line option to `run_generation.py`.

## Before submitting
- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

### DIFF
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -159,6 +159,18 @@ python ../gaudi_spawn.py --use_deepspeed --world_size 8 run_generation.py \
 --do_sample
 ```
 
+To run Mixtral-8x7B inference on a single Gaudi2 card, use the following command:
+```bash
+python run_generation.py \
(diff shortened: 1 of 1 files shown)


mounikamandava (approved):

libinta:
@tjs-intel  as long as it's working on gaudi, it's ok. no need to put into readme unless there is any special arguments.

tjs-intel:
@libinta the `--bf16` argument is required for single-card inference because the model is so large.

libinta:
Taulor,
Thanks. As there is no new argument added, I don't think we need special instructions for it.
Libin

PR #849, opened 2024-03-29
Remove contingous call

libinta:
# What does this PR do?
The contingous call causes performance degradation for not reuse_cache case such as TGI.
Remove it for performance improvement.

Fixes # (issue)

## Before submitting


szutenberg (changes requested):
This change does not solve the issue with performance regression in the TGI MLPerf scenario.

My proposal is to revert https://github.com/huggingface/optimum-habana/commit/844908dc85875362270f6e58fb952b37f84cb53a which introduced the regression. I'll push PR shortly.

PR #850, opened 2024-03-29
Revert "Synapse 1.15 addition2 (#844)"

szutenberg:
This reverts commit 844908dc85875362270f6e58fb952b37f84cb53a.

# What does this PR do?

This PR fixes huge perf regression in LLaMA2-70B MLPerf scenario (using tgi-gaudi).

PR https://github.com/huggingface/optimum-habana/pull/849 does not solve the issue so I recommend to fix synapse_1.15 first and then see how the issue with TGI can be resolved.

CC @libinta @kdamaszk @regisss

### DIFF
--- a/tests/transformers/tests/models/falcon/test_modeling_falcon.py
+++ b/tests/transformers/tests/models/falcon/test_modeling_falcon.py
@@ -353,7 +353,7 @@ class FalconModelTest(ModelTesterMixin, GenerationTesterMixin, unittest.TestCase
             outputs = model(**inputs)
 
             # If "past_key_values" is not returned, pass the test (e.g. RWKV uses a different cache name and format)
-            if "past_key_values" not in outputs:
+            if "past_key_values" not in outputs or all(ele is None for ele in outputs["past_key_values"]):
                 return
 
             num_hidden_layers = (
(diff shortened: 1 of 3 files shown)


tests/transformers/tests/models/falcon/test_modeling_falcon.py
```diff
@@ -353,7 +353,7 @@ def test_past_key_values_format(self):
             outputs = model(**inputs)
 
             # If "past_key_values" is not returned, pass the test (e.g. RWKV uses a different cache name and format)
-            if "past_key_values" not in outputs or all(ele is None for ele in outputs["past_key_values"]):
```
regisss, line 356: We should keep this change. It's not part of the source code so 100% sure it's not responsible for the regression.

regisss (commented):
Looks good, there's just one change in the commit you're reverting that we should keep

tests/transformers/tests/models/falcon/test_modeling_falcon.py
szutenberg: How about reverting only change in modeling_llama? I heard from @libinta that some tests were failing without this change but unfortunately I don't have more context.

regisss: Sounds good to me! We'll patch Falcon later if there is an issue there too

regisss (approved):

PR #870, opened 2024-04-07
add `--trust_remote_code` for text generation examples

yangulei:
Add `--trust_remote_code` argument for text generation examples to allow the custom models defined on the Hub in their own modeling files.

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -260,6 +260,11 @@ def setup_parser(parser):
         action="store_true",
         help="Whether to enable device map auto. In case no space left on cpu, weights will be offloaded to disk.",
     )
+    parser.add_argument(
+        "--trust_remote_code",
+        action="store_true",
+        help="Whether or not to allow for custom models defined on the Hub in their own modeling files.",
+    )
     args = parser.parse_args()
 
     if args.torch_compile:
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -375,7 +375,7 @@ def initialize_model(args, logger):
         "trust_remote_code": args.trust_remote_code,
     }
     if args.trust_remote_code:
-        print("trust_remote_code is set. Model may not be complaint and may fail")
+        logger.warning("`trust_remote_code` is set, there is no guarantee this model works properly and it may fail")
     if args.disk_offload:
         model_kwargs["device_map"] = "auto"
         model_kwargs["offload_folder"] = "/tmp/offload_folder/"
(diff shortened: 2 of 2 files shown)


examples/text-generation/run_generation.py
```diff
@@ -260,6 +260,11 @@ def setup_parser(parser):
         action="store_true",
         help="Whether to enable device map auto. In case no space left on cpu, weights will be offloaded to disk.",
     )
+    parser.add_argument(
+        "--trust_remote_code",
+        action="store_true",
+        help="Whether or not to allow for custom models defined on the Hub in their own modeling files.",
```
ssarkar2, line 266: Maybe could be reworded a bit.

Like: `Allow custom models defined on the Hub in their own modeling files`

Also the files could be local path, so doesn't have to be from the hub. So maybe: `Allow models with their own modeling files outside transformers repo`

ssarkar2 (commented):
LGTM otherwise, minor comment regarding the help message

examples/text-generation/run_generation.py
yangulei: Sorry for the latency, the help message here come from huggingface doc: https://huggingface.co/docs/transformers/model_doc/auto#transformers.AutoConfig.from_pretrained.trust_remote_code

ssarkar2 (approved):

regisss (commented):
This one is tricky as remote code offers absolutely no guarantee of compliance with the Transformers API. I've pushed back against it in the past because this example will for sure fail with at least a few models with remote code out there.
I'm okay with allowing it, but let's log a warning that clearly states that `trust_remote_code` is true so this example may fail as this model doesn't necessarily follow the Transformers API.

examples/text-generation/utils.py
```diff
@@ -370,7 +372,10 @@ def initialize_model(args, logger):
     model_kwargs = {
         "revision": args.model_revision,
         "token": args.token,
+        "trust_remote_code": args.trust_remote_code,
     }
+    if args.trust_remote_code:
+        print("trust_remote_code is set. Model may not be complaint and may fail")
```
regisss, line 378:
```suggestion
        logger.warning("`trust_remote_code` is set, there is no guarantee this model works properly and it may fail")
```

regisss (approved):

ImtiazSajwani:
Thanks for the quick workaround but I am still seeing the error

$git status
On branch remote_code

  File "/usr/local/lib/python3.10/dist-packages/transformers/dynamic_module_utils.py", line 621, in resolve_trust_remote_code
    raise ValueError(
ValueError: Loading requires you to execute the configuration file in that repo on your local machine. Make sure you have read the code there to avoid malicious use, then set the option `trust_remote_code=True` to remove this error.
[

yangulei:
@ImtiazSajwani Could you share the command to reproduce this error?

PR #871, opened 2024-04-08
Llama Guard for text classification example

dsmertin:
# What does this PR do?

Enabled Llama Guard for text classification example:
* Added `add_pad_token` property (Will add `pad_token` to tokenizer and model's config as `eos_token` if it's not defined).
There's general issue that **meta/llama** models don't have *pad_token* in their configuration. Switching off padding for fine-tuning and inference leads to low performance and maybe to other issues for sure. Please, let me know if there's possible another way to do it.
* Changed README, Validated Models section. The script with Llama Guard was validated on Gaudi 2. Inference works with both 1 card and Multicard (DeepSpeed). Training is not possible on 1 card due to out of memory exception.
Script was also validated with original `bert-large-uncased-whole-word-masking` used in example - no regressions was found.

Dear maintainers, I want to mention in advance that I have doubts about making the property `add_pad_token=True` by default. If this flag should be off by default, then I'll add additional paragraph to the readme, with mentioning this property and its usage for Llama Guard.

## Before submitting


examples/text-classification/run_glue.py
```diff
@@ -230,6 +230,10 @@ class ModelArguments:
         default=False,
         metadata={"help": "Will enable to load a pretrained model whose head dimensions are different."},
     )
+    add_pad_token: bool = field(
+        default=True,
```
libinta, line 234: please change to default as False, and add readme change for llama_gaudi

libinta (commented):
add a short ci test for this

libinta (commented):
@dsmertin how is it comparing with gpu? also please add a small ci test.

libinta:
@dsmertin  can you add a section in readme [of ](https://github.com/huggingface/optimum-habana/blob/main/examples/question-answering/README.md) to specify the cmd?
Also for ci test case, you need to do 
export GAUDI2_CI=1

examples/text-classification/README.md
```diff
@@ -155,3 +155,78 @@ python run_glue.py \
...
+    },
+    "gradient_clipping": 1.0,
+    "zero_optimization": {
+        "stage": 2,
+        "overlap_comm": false,
+        "reduce_scatter": false,
+        "contiguous_gradients": false
+    }
+}
+```
```
libinta, line 208: can we use the one from https://github.com/huggingface/optimum-habana/blob/main/tests/configs/deepspeed_zero_2.json ? and remove this?
check how other example README with deepspeed

examples/text-classification/README.md
```diff
@@ -155,3 +155,59 @@ python run_glue.py \
...
+Llama Guard can be used for text classification task. The *transformers* library will change the head of the model for you during fine-tuning or inference. You can use same general command as for BERT, except you need to add `--add_pad_token=True`, because Llama based models don't have `pad_token` in their model and tokenizer configuration files. So, `--add_pad_token=True` will add `pad_token` to tokenizer and model's config as `eos_token` if it's not defined.
+
+### Fine-tuning with DeepSpeed
+
+Llama Guard can be fine-tuned with DeepSpeed, here is how you would do it on the text classification MRPC task using DeepSpeed with 8 HPUs:
+
+```bash
+python ../gaudi_spawn.py /
+    --world_size 8 --use_deepspeed run_glue.py \
+    --model_name_or_path meta-llama/LlamaGuard-7b \
```
libinta, line 170: For 7b model, can you fit to 1 HPU? have you checked if the deepspeed has better perf than mpi?

examples/text-classification/README.md
```diff
@@ -155,3 +155,59 @@ python run_glue.py \
...
+    --gaudi_config gaudi_config.json  \
```
libinta, line 171: I didn't see any gaudi_config.json. where is it?

examples/text-classification/README.md
```diff
@@ -155,3 +155,59 @@ python run_glue.py \
...
+
+> If your model classification head dimensions do not fit the number of labels in the dataset, you can specify `--ignore_mismatched_sizes` to adapt it.
+
+### Inference
+
+You can run inference with Llama Guard on GLUE on 1 Gaudi card with the following command:
+
+```bash
+python run_glue.py \
+  --gaudi_config gaudi_config.json \
```
libinta, line 199: where is the gaudi_config.json?

examples/text-classification/README.md
libinta: can you send me your ci test cmd and log from run?

examples/text-classification/README.md
libinta: @dsmertin  we just create https://huggingface.co/Habana/llama, can you try to use this?

libinta (approved):

regisss (approved):
I just pushed a new commit to modify a bit the CI because Llama Guard was tested for all Llama tasks.

PR #877, opened 2024-04-09
Remove deprecated AOT_HPU_TRAINING_BACKEND (#138)

astachowiczhabana:
The new backend has been introduced to pytorch-integraton - HPU_BACKEND . The deprecated backend ( AOT_HPU_TRAINING_BACKEND ) shall no longer be available in optimum habana as it's going to be removed from pytorch-integration.

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -151,7 +151,7 @@ def patch_scoped_linear_all_reduce(model):
 
 
 def get_torch_compiled_model(model):
-    model.model = torch.compile(model.model, backend="aot_hpu_inference_backend")
+    model.model = torch.compile(model.model, backend="hpu_backend")
     return model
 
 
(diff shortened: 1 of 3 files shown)


optimum/habana/version.py
libinta, line 1: why do we need this change?

setup.py
```diff
@@ -49,7 +49,7 @@
 
 QUALITY_REQUIRES = [
     "ruff",
-    "hf_doc_builder @ git+https://github.com/huggingface/doc-builder.git",
+    "hf_doc_builder",
```
libinta, line 52: why do we need this change?

libinta (commented):
@astachowiczhabana  can you rebase to main

optimum/habana/version.py
regisss: This change appears because I change the target of the PR from `v1.11-release` to `main`. Merging main into the PR branch will make it disappear.

setup.py
regisss: Same as above

astachowiczhabana:
@libinta @regisss, it's rebased on main

regisss (approved):

PR #886, opened 2024-04-12
Enable HQT matmul fp8 quantization for Mistral

jiminha:
Enable HQT matmul fp8 quantization for Mistral Attention layer.
Improves performance (from 11313 samples/s to 12935 for 128x128x bs 896 configuration). 

Softmax fp8 quantization doesn't seem to add any performance improvement, but actually slows it down since fused kernel is not being used.


jiminha:
Supposed to go to fork branch. I'm closing this

PR #888, opened 2024-04-14
Fixed "reuse_cache" Bug

Danielohayon:
Fixed a bug which adds the "reuse_cache" flag regardless of the model type


regisss (approved):
Good catch!

PR #892, opened 2024-04-16
Add linear and dynamic RoPE to Mistral and Mixtral

regisss:
# What does this PR do?

As per title.

## Before submitting


libinta (commented):
@regisss for the configuration, do we need to change the model card or let user to pass from cmd line? so we need to add arguments from run_generation?

regisss:
> @regisss for the configuration, do we need to change the model card or let user to pass from cmd line? so we need to add arguments from run_generation?

We don't need to change the `config.json`. It will use regular RoPE by default. If users want linear or dynamic RoPE, they can modify the config of the model. That's exactly the same as with Llama so I don't think we should add any new argument.

libinta (approved):

PR #906, opened 2024-04-19
Set ds config "reduce scatter" to false

vineethanandh:
# What does this PR do?
Set ds config "reduce scatter" to false

Fixes # (issue)
FlanT5 is giving a perf drop with "reduce_scatter" value True in DS config
## Before submitting


vineethanandh:
This was a wrong pull request

PR #908, opened 2024-04-19
fix perf drop in flan-t5 summarization

MohitIntel:
Upto 7% performance drop observed in flan-t5 8x summarization task with deepspeed-fork (synapse 1.16 version).
This PR fixes that.


regisss (approved):
LGTM! It also works with 1.15 so merging this now.

astachowiczhabana:
https://github.com/HabanaAI/optimum-habana-fork/pull/173

PR #913, opened 2024-04-22
Add support for symbolic tracing of GPT2 models

tjs-intel:
# What does this PR do?

Addresses test failures from `GAUDI2_CI=1  RUN_SLOW=1 python -m pytest tests/transformers/tests/models/gpt2/test_modeling_gpt2.py -s -v -k test_torch_fx`

Removes an obstacle for symbolic tracing of GPT2 models. Problem line was `[1 for _ in x.shape]`. It is impossible to iterate over `x.shape` during symbolic execution. AFAICT `self.bias` is broadcast-able with `x` without re-shaping, and according to PyTorch's broadcasting rules the behavior is identical to the re-shaping.

Adds the `GaudiGPT2*` models to the supported fx model list in transformers.

@ZhaiFeiyue seems to be the original author.

## Before submitting


regisss (commented):
LGTM, there is one thing to fix though

regisss (approved):

PR #916, opened 2024-04-23
Pytest accuracy updates for Falcon, T5, GPT2

Luca-Calabria:
# Update expected output and check if it is deterministic, to fix failures on Pytest.

Reference task: [GS-114](https://habana.atlassian.net/browse/GS-114)

“Falcon” and “T5” produce a deterministic result. Expected output has been updated.
”GPT2” produces each time a different result.

All three models produce deterministic token length.

## Before submitting


regisss (commented):
LGTM!

Can you also run
```
pip install -U ruff
make style
```
from the root of the repo please to have the code style check pass?

regisss (approved):

PR #934, opened 2024-04-30
Added Llama-2 fp8 text-generation test cases

yeonsily:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/tests/test_text_generation_example.py
+++ b/tests/test_text_generation_example.py
@@ -229,8 +229,8 @@ def test_text_generation_bf16(model_name: str, baseline: float, batch_size: int,
 
 @pytest.mark.parametrize("model_name, baseline", MODELS_TO_TEST["fp8"])
 def test_text_generation_fp8(model_name: str, baseline: float, token: str):
-    deepspeed = True if "falcon-180B" or "Llama-2-70b" in model_name else False
-    world_size = 8 if "falcon-180B" or "Llama-2-70b" in model_name else None
+    deepspeed = True if "falcon-180B" in model_name or "Llama-2-70b" in model_name else False
+    world_size = 8 if "falcon-180B" in model_name or "Llama-2-70b" in model_name else None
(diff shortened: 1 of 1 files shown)


tests/test_text_generation_example.py
```diff
@@ -37,6 +37,8 @@
         ],
         "fp8": [
             ("tiiuae/falcon-180B", 52.85086442722326),
+            ("meta-llama/Llama-2-7b-hf", 0.0),
+            ("meta-llama/Llama-2-70b-hf", 0.0),
```
regisss, line 41: Baselines are 0.0 here because we use the baselines defined in `LLAMA2_FP8_CONFIG` right?

tests/test_text_generation_example.py
```diff
@@ -134,6 +150,11 @@ def _test_text_generation(
             "--reuse_cache",
             "--trim_logits",
         ]
+        if "Llama-2" in model_name:
+            command += [
+                "--attn_softmax_bf16",
```
regisss, line 155: This could be added for all Llama models and configs (i.e. not only fp8) no?

regisss (commented):
Are these tests supposed to pass with 1.15 or is it for 1.16?

yeonsily:
> Are these tests supposed to pass with 1.15 or is it for 1.16?

The numbers are collected on 1.15 release.

tests/test_text_generation_example.py
yeonsily: Yes.

tests/test_text_generation_example.py
yeonsily: Actually I see that it already added on the latest main branch.
https://github.com/huggingface/optimum-habana/blob/main/tests/test_text_generation_example.py#L114
Let me remove this line.

regisss (approved):

PR #936, opened 2024-04-30
Initial enablement with FP8 Training (port from OHF #91)

libinta:
# What does this PR do?
Initial enablement with FP8 Training with Intel Gaudi Transformer Engine ([(porting from OHF #91)
Only linear layer is replaced with FP8.

Fixes # (issue)

## Before submitting


optimum/habana/accelerate/accelerator.py
```diff
@@ -75,6 +75,10 @@
     GaudiDynamoBackend,
     GaudiFullyShardedDataParallelPlugin,
     GaudiTorchDynamoPlugin,
+    te_forward_convert,
```
SanjuCSudhakaran, line 78: Function names are a little confusing

optimum/habana/accelerate/accelerator.py
```diff
@@ -176,6 +181,7 @@ def __init__(
         self.scaler_handler = None
         self.init_handler = None
         self.fp8_recipe_handler = None
+        self.fp8_recipe_format = None
```
SanjuCSudhakaran, line 184: Why are we resetting the fp8_recipe_format here?

optimum/habana/accelerate/accelerator.py
```diff
@@ -458,7 +467,11 @@ def _prepare_deepspeed(self, *args):
 
         is_dataloader_present = any(isinstance(obj, torch.utils.data.DataLoader) for obj in args)
         result = [
```
SanjuCSudhakaran, line 469: The changes are specific to deepspeed. If you really don't need to use deepspeed, you could avoid making these changes for now.

libinta: We also tried with mixral8-7b, and it uses deepspeed

optimum/habana/accelerate/accelerator.py
libinta: we want to user to have option to select the fp8_recipe_format

regisss (approved):
LGTM

optimum/habana/accelerate/accelerator.py
regisss: I agree, but on the other hand I feel like this has more to do with the Transformer Engine library which pushed the use of `te`. So I think it's okay since it will always be used in the context of fp8 mixed precision.

PR #937, opened 2024-04-30
Fix wav2vec test load adapter

malkomes:
# What does this PR do?

It fixes the ```test_modeling_wav2vec2.py::test_load_attn_adapter``` bypassing a problem of how we save the `adapter_weights` and load them in the test.

Fixes # (issue)

When we try to do `model.load_adapter` in the following code

```python
import os
import tempfile
import torch
from transformers import Wav2Vec2ForCTC
from transformers.models.wav2vec2.modeling_wav2vec2 import WAV2VEC2_ADAPTER_PT_FILE

model = Wav2Vec2ForCTC.from_pretrained("hf-internal-testing/tiny-random-wav2vec2", adapter_attn_dim=16)

with tempfile.TemporaryDirectory() as tempdir:
    model.save_pretrained(tempdir)
    model = Wav2Vec2ForCTC.from_pretrained(tempdir)
    adapter_weights = model._get_adapters()

    pt_filepath = os.path.join(tempdir, WAV2VEC2_ADAPTER_PT_FILE.format("eng"))
    torch.save(adapter_weights, pt_filepath)

    model.load_adapter("eng")
```

we get the following error

```text
  File "/usr/local/lib/python3.10/dist-packages/transformers/models/wav2vec2/modeling_wav2vec2.py", line 1339, in load_adapter
    state_dict = torch.load(
  File "/usr/local/lib/python3.10/dist-packages/torch/serialization.py", line 1025, in load
    raise pickle.UnpicklingError(UNSAFE_MESSAGE + str(e)) from None
_pickle.UnpicklingError: Weights only load failed. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution.Do it only if you get the file from a trusted source. WeightsUnpickler error: Unsupported operand 149
```
which suggests that we should change how we load and/or save the `adapter_weights` in the `Wav2Vec2PreTrainedModel` class. Since this error is unrelated to `Gaudi` we just bypass this problem
by directly loading the `adapter_weights` in this PR. 

Once this issue is fixed on the `transformers` package, this code will crash and we only need to revert the changes to use `load_adapter` directly.

## Before submitting

### DIFF
--- a/tests/transformers/tests/models/wav2vec2/test_modeling_wav2vec2.py
+++ b/tests/transformers/tests/models/wav2vec2/test_modeling_wav2vec2.py
@@ -1257,8 +1257,16 @@ class Wav2Vec2RobustModelTest(ModelTesterMixin, unittest.TestCase):
             pt_filepath = os.path.join(tempdir, WAV2VEC2_ADAPTER_PT_FILE.format("eng"))
             torch.save(adapter_weights, pt_filepath)
 
-            model.load_adapter("eng")
-            model.load_adapter("eng", use_safetensors=False)
+            # model.load_adapter is broken in transformers
+            # since adapter_weights fails to load with weights_only=True
+            with self.assertRaises(OSError):
+                model.load_adapter("eng")
+            with self.assertRaises(OSError):
+                model.load_adapter("eng", use_safetensors=False)
+            # we will load adapter_weights directly while model.load_adapter fails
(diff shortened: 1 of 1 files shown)


jiminha (approved):
R&D working on the load_adapter fix, but in the meanwhile, this would fix the CI failure.

regisss (approved):
LGTM!

Issue #933, opened 2024-04-30
RuntimeError: a Tensor with 32768 elements cannot be converted to Scalar


astachowiczhabana:
### System Info

```shell
docker image:
vault.habana.ai/gaudi-docker/1.15.1/ubuntu22.04/habanalabs/pytorch-installer-2.2.0:latest

oh:
https://github.com/huggingface/optimum-habana/tree/bb6be54d112bcaed2d0ff8ab10ea66b8578ba0ba
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

1. Pull latest HabanaAI docker
```
docker pull vault.habana.ai/gaudi-docker/1.15.1/ubuntu22.04/habanalabs/pytorch-installer-2.2.0:latest
```
2. Clone latest optimum-habana
```
apt update; apt install gh; gh repo clone huggingface/optimum-habana
```
3. Reproduce the issue
```
cd optimum-habana-fork/examples/language-modeling/; pip install -r requirements.txt; pip install -e ../..; PT_HPU_LAZY_MODE=0 python ../gaudi_spawn.py \
    --world_size 4 --use_mpi run_lora_clm.py \
    --model_name_or_path Llama-2-7b-chat-hf \
    --dataset_name tatsu-lab/alpaca \
    --bf16 True \
    --output_dir /tmp/model_lora_llama_ddp \
    --num_train_epochs 2 \
    --per_device_train_batch_size 8 \
    --evaluation_strategy "no" \
    --save_strategy "no" \
    --learning_rate 1.0e-05 \
    --warmup_ratio  0.03 \
    --lr_scheduler_type "linear" \
    --max_grad_norm  0.3 \
    --logging_steps 1 \
    --do_train \
    --use_habana \
    --use_lazy_mode False \
    --throughput_warmup_steps 3 \
    --lora_rank=8 \
    --lora_alpha=32 \
    --lora_dropout=0.1 \
    --lora_target_modules "q_proj" "v_proj" \
    --dataset_concatenation \
    --max_seq_length 512 \
    --ddp_bucket_cap_mb 50 \
    --adam_epsilon 1e-08 \
    --low_cpu_mem_usage True 
```

4. Result
```
    main()
  File "/root/optimum-habana/examples/language-modeling/run_lora_clm.py", line 792, in main
[...]

PR #938, opened 2024-04-30
Grad norm tensor fix

yeonsily:
# What does this PR do?

Fixes https://github.com/huggingface/optimum-habana/issues/933

## Before submitting


libinta (approved):

regisss:
@yeonsily I pushed a new commit because `_grad_norm.size() == 1` will always return False as we have `_grad_norm.size() == torch.Size([1])` in lazy mode.
I also prefer not logging the grad norm if it is a list to not pollute the logs with a very big tensor.

regisss (approved):

PR #946, opened 2024-05-03
Workarounds for failures in using HF Transformers auto classes in Classification sample 

sandeep-maddipatla:
# What does this PR do?

This PR provides options to workaround failures seen in the image-classification example when invoking the HF Transformer autoclass pretrained functions, while running with certain classification models on HF.

Specifically, two cases are handled:
1) Cases where the model's safetensor file format is not compatible with the HF Transformer lib expectation. (a `metadata` field is expected).
2) Print warning and directions to workaround if `AutoImageProcessor.from_pretrained` fails on account of a missing `preprocessor_config.json` file.

Fixes GS-123

## Before submitting

### DIFF
--- a/examples/image-classification/run_image_classification.py
+++ b/examples/image-classification/run_image_classification.py
@@ -357,13 +357,19 @@ def main():
         ignore_mismatched_sizes=model_args.ignore_mismatched_sizes,
         use_safetensors=model_args.use_safetensors
     )
-    image_processor = AutoImageProcessor.from_pretrained(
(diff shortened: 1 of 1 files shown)


regisss:
@sandeep-maddipatla I think this change should first be merged into Transformers and then we can add it here. Would you like to open a PR in Transformers?

yafshar:
@sandeep-maddipatla would you please respond to regiss' comment? We would like to move forward

sandeep-maddipatla:
Closing this for now. Will look into merging this into transformers. Will reopen if necessary at that point.

PR #947, opened 2024-05-03
Add leaderboard link

echarlaix:
cc @mfuntowicz after discussing this during the bi-weekly, not sure if this should be added to `optimum-intel` instead (not sure that make sense as resulting models won't be compatible with the leaderboard but we could mention `optimum-habana` and add the link for example https://github.com/huggingface/optimum-intel/pull/699)

Feel free to close @regisss


regisss (approved):
LGTM!

PR #960, opened 2024-05-07
Fix warmup for diffusers when batch size < throughput_warmup_steps

dsocek:
# What does this PR do?

Fixes throughput calculation for diffusers when `batch size < throughput_warmup_steps`

### DIFF
--- a/examples/stable-diffusion/text_to_image_generation.py
+++ b/examples/stable-diffusion/text_to_image_generation.py
@@ -203,24 +203,30 @@ def main():
         "--ldm3d", action="store_true", help="Use LDM3D to generate an image and a depth map from a given text prompt."
     )
     parser.add_argument(
-        "--profiling_warmup_steps",
-        default=0,
+        "--throughput_warmup_steps",
         type=int,
+        default=None,
+        help="Number of steps to ignore for throughput calculation.",
+    )
+    parser.add_argument(
+        "--profiling_warmup_steps",
+        type=int,
+        default=0,
         help="Number of steps to ignore for profiling.",
     )
     parser.add_argument(
         "--profiling_steps",
-        default=0,
         type=int,
+        default=0,
         help="Number of steps to capture for profiling.",
     )
     args = parser.parse_args()
 
     # Set image resolution
-    res = {}
+    kwargs_call = {}
     if args.width > 0 and args.height > 0:
-        res["width"] = args.width
-        res["height"] = args.height
+        kwargs_call["width"] = args.width
+        kwargs_call["height"] = args.height
 
     # ControlNet
     if args.control_image is not None:
@@ -293,6 +299,9 @@ def main():
     if args.bf16:
         kwargs["torch_dtype"] = torch.bfloat16
 
+    if args.throughput_warmup_steps is not None:
+        kwargs_call["throughput_warmup_steps"] = args.throughput_warmup_steps
+
     # Generate images
     if args.control_image is not None:
         model_dtype = torch.bfloat16 if args.bf16 else None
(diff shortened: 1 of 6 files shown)


examples/stable-diffusion/text_to_image_generation.py
```diff
@@ -202,25 +202,31 @@ def main():
     parser.add_argument(
         "--ldm3d", action="store_true", help="Use LDM3D to generate an image and a depth map from a given text prompt."
     )
+    parser.add_argument(
+        "--throughput_warmup_steps",
+        type=int,
+        default=None,
```
yeonsily, line 208: If this is not given, I see that it uses 3 by default in denoising step. Then it's better to give default value 3 here?

dsocek: It may be better to keep default to be given in the pipeline itself, so different pipelines may have specific warmup defaults

yeonsily:
LGTM

optimum/habana/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py
```diff
@@ -495,18 +495,27 @@ def __call__(
 
             # 8. Denoising loop
             throughput_warmup_steps = kwargs.get("throughput_warmup_steps", 3)
+            logger.info(f"{throughput_warmup_steps} warmup steps will be used in measuring throughput")
```
regisss, line 498: I don't think we should log that, I think it would be confusing for users

examples/stable-diffusion/text_to_image_generation.py
```diff
@@ -293,6 +299,9 @@ def main():
     if args.bf16:
         kwargs["torch_dtype"] = torch.bfloat16
 
+    if args.throughput_warmup_steps != None:
```
regisss, line 302:
```suggestion
    if args.throughput_warmup_steps is not None:
```

dsocek: Fixed

optimum/habana/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py
dsocek: Sounds good, removed the logger info

optimum/habana/utils.py
```diff
@@ -117,6 +117,29 @@ def speed_metrics(
     return result
 
 
+def warmup_inference_steps_time_adjustment(start_time_after_warmup, start_time_after_inference_steps_warmup, num_inference_steps, warmup_steps):
+    """
+    Adjust start time after warmup to account for warmup inference steps.
+
+    When warmup is applied to multiple inference steps within a single sample generation we need to account for
+    skipped inference steps time to estmate "per sample generation time".  This functioin computes the average
```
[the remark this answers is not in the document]
dsocek: Fixed

dsocek:
@regisss thanks for review, updated PR with all of your suggested fixes

regisss:
@dsocek Will just need to run `make style` and then I merge this one :slightly_smiling_face:

dsocek:
> @dsocek Will just need to run `make style` and then I merge this one 🙂

@regisss Sounds good, done!

regisss:
@dsocek The CI failed with this error:
```
UnboundLocalError: local variable 't1_inf' referenced before assignment
```
There must be a case in CI where `num_inference_steps` is smaller than `throughput_warmup_steps`.

dsocek:
> @dsocek The CI failed with this error:
> 
> ```
> UnboundLocalError: local variable 't1_inf' referenced before assignment
> ```
> 
> There must be a case in CI where `num_inference_steps` is smaller than `throughput_warmup_steps`.

hmm possibly for turbo case.  I will check an fix this

regisss (approved):

PR #963, opened 2024-05-08
llama70b one card to infer device map with max memory limitation 

Yantom1:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -160,17 +160,27 @@ def setup_model(args, model_dtype, model_kwargs, logger):
 
     if args.disk_offload:
         from accelerate import infer_auto_device_map, init_empty_weights
+
         config = AutoConfig.from_pretrained(args.model_name_or_path)
         with init_empty_weights():
             model = AutoModelForCausalLM.from_config(config)
         max_memory = {"cpu": "10GiB"}
         device_map = infer_auto_device_map(model, max_memory=max_memory, dtype=model_dtype)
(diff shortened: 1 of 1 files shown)


examples/text-generation/utils.py
```diff
@@ -158,10 +158,19 @@ def get_torch_compiled_model(model):
...
+        with init_empty_weights():
+            model = AutoModelForCausalLM.from_config(config)
+        max_memory = {"cpu": "10GiB"}
+        device_map = infer_auto_device_map(model, max_memory=max_memory, dtype=model_dtype)
+        model = AutoModelForCausalLM.from_pretrained(args.model_name_or_path, device_map=device_map, offload_folder="/tmp/offload_folder/", offload_state_dict=True, torch_dtype=model_dtype, **model_kwargs)  
+    else: 
+        if args.peft_model is not None:
+            model = peft_model(args, model_dtype, logger, **model_kwargs)
+        else:
+            model = AutoModelForCausalLM.from_pretrained(args.model_name_or_path, torch_dtype=model_dtype, **model_kwargs)
```
regisss, line 173: This means we cannot use disk offloading with PEFT models, is that intended?

regisss (commented):
LGTM!
Can you also share a command that I can execute to test it please?
You will also need to run
```
pip install -U ruff
make style
```
from the root of the repo to make the code style check pass.

Yantom1:
> LGTM! Can you also share a command that I can execute to test it please? You will also need to run
> 
> ```
> pip install -U ruff
> make style
> ```
> 
> from the root of the repo to make the code style check pass.

@regisss fixed the formation. The disk_offload is tested nightly

regisss (approved):
LGTM!

It would probably work the same for PEFT models where we call `merge_and_unload`.

PR #965, opened 2024-05-08
fix wrongly all_gather for mixtral finetune

ccrhx4:
# What does this PR do?
When enabling finetune for Mixtral-7x8b, the train_loss is totally wrong when running with deepspeed on 8 cards. We found these all_gather code were hard coded in the model code, it lead to training totally wrong.


regisss:
@ccrhx4 Good catch! Can you share the command you ran for Mixtral fine-tuning please?

regisss (approved):

PR #975, opened 2024-05-13
fix prefix tuning finetune issue and update test

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/habana/transformers/models/llama/modeling_llama.py
```diff
@@ -169,11 +169,10 @@ def update(self, prev, cur, dim, idx, inp_seq_len):
         if prev.shape == cur.shape:
             prev.copy_(cur)
             return orig_cur
-        if cur.shape[2] > 1 and cur.shape[2] <= prev.shape[2]:
+        if idx is not None and cur.shape[2] > 1 and cur.shape[2] <= prev.shape[2]:
```
regisss, line 172: Why do we need to check if idx is not None? It's not used in the block right below no?

sywangyi: in prefix-tuning finetuning case, past-key-value(trainable) is generated by peft and  should be concat with current-key-value even if cur.shape[2] > 1 and cur.shape[2]<=prev.shape[2]. which is different with inference case where idx is not None.

regisss (approved):

Issue #865, opened 2024-04-04
Update to transformers 4.39.1+


vidyasiv:
### Feature request

If an update is already planned, could you please let us know the expected timeline. We had begun working on this and found we needed a more changes (that could also affect other models) than we were comfortable making at once.

### Motivation

In order to enable [CohereForAI/c4ai-command-r-v01](https://huggingface.co/CohereForAI/c4ai-command-r-v01) an update is needed to v4.39.1

### Your contribution

Part of the upgrade work can be found (very cohereforai specific) : https://github.com/vidyasiv/optimum-habana/tree/cohereforai (only works for batch size=1 for run_generation.py) and higher batch sizes fail as it needs further changes.
Contributors: @skavulya and @vidyasiv

regisss:
Transformers v4.40 should be released soon so I'll work on directly jumping from 4.38 to 4.40

vidyasiv:
Thanks @regisss , do you have a rough estimate of when that could be i.e end of April or earlier etc?

regisss:
I'll push a branch compatible with Transformers v4.40 next week. If there is no major regression, it should be merged a few days after.

regisss:
Upgraded to Transformers v4.40 in #1027.

PR #982, opened 2024-05-14
Tests for LogitsProcessor updated with transformers 4.38.2

malkomes:
# What does this PR do?

- Updates `LogitsProcessorTest` to match tests from `transformer == 4.38.2`. 
- Assert that outputs are in the expected device,  HPU.

Fixes # (issue)

Aside from new tests, the only issue was related to using different `torch_device`s. The solution here was to import the `torch_device` from `test_modeling_common` which uses the HPU.

## Before submitting

### DIFF
--- a/tests/transformers/tests/generation/test_logits_process.py
+++ b/tests/transformers/tests/generation/test_logits_process.py
@@ -17,15 +17,17 @@ import unittest
 from typing import List, Union
 
 from parameterized import parameterized
-from transformers import is_torch_available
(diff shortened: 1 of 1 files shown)


malkomes:
@regisss is this PR still useful? or do you want me to update to 4.40+? https://github.com/huggingface/optimum-habana/issues/865#issuecomment-2079006079

regisss:
@malkomes That would be great if you can update it to v4.40!

PR #983, opened 2024-05-15
Llama adapter

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting


emascarenhas:
Yi, I have 2 initial questions.
1. The default of 2 adapter_layers is that expected to give comparable results in terms of accuracy like LoRA? What do the authors of llama_adapter say about the default?
2. why the change from unittest to pytest in test_peft_inference.py?

sywangyi:
> Yi, I have 2 initial questions.
> 
> 1. The default of 2 adapter_layers is that expected to give comparable results in terms of accuracy like LoRA? What do the authors of llama_adapter say about the default?
> 2. why the change from unittest to pytest in test_peft_inference.py?

1. please see https://github.com/OpenGVLab/LLaMA-Adapter, the simple fine-tuning code of LLaMA-Adapter on LLaMA-7B model, I just change  default value of adapter_layer  and adapter_len to adapt to the  command.
2. for use of pytest.mark.parametrize. We can have only one method called  test_text_generation_llama(with the @pytest.mark.parametrize(...) decorator to avoid defining one peft test per model

regisss (approved):

PR #984, opened 2024-05-15
Fix for the simulate_dyn_prompt flag assertion

alekseyfa:
# What does this PR do?

This PR fixes an error that occurred when the `--dataset_name` flag was set. 

```bash
AssertionError: Both dataset_name and simulate_dyn_prompt are set
```

The `assert args.simulate_dyn_prompt == ""` never passed because `None` was compared to an empty string.

## Before submitting


ssarkar2 (approved):

MohitIntel (approved):
I tested this change with the command:
python run_generation.py --model_name_or_path gpt2 --batch_size 2 --max_new_tokens 100 --use_hpu_graphs --use_kv_cache --dataset_name JulesBelveze/tldr_news --column_name content --bf16 --simulate_dyn_prompt 25 35 45

The fix works correctly and throws the assertion. Without the fix, the assertion was being thrown wrongly (even when not using --simulate_dyn_prompt 25 35 45).

ssarkar2 (changes requested):
@alekseyfa , can you also make this readme fix in this section:
https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation#using-growing-bucket-optimization

`For example --simulate_dyn_prompt 25,35,45`
this should be: `For example --simulate_dyn_prompt 25 35 45`

Its used correctly at the end of the para:
`Thus the suggested configuration to simulate dynamicity after warmup is to use all three arguments: --simulate_dyn_prompt 25 35 45 --reduce_recompile --bucket_size 30`

alekseyfa:
> @alekseyfa , can you also make this readme fix in this section: https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation#using-growing-bucket-optimization
> 
> `For example --simulate_dyn_prompt 25,35,45` this should be: `For example --simulate_dyn_prompt 25 35 45`
> 
> Its used correctly at the end of the para: `Thus the suggested configuration to simulate dynamicity after warmup is to use all three arguments: --simulate_dyn_prompt 25 35 45 --reduce_recompile --bucket_size 30`

 @ssarkar2 Done

ssarkar2 (approved):

regisss (approved):

PR #986, opened 2024-05-16
Use the lenght of timesteps as the inference step num

yuanwu2017:
According to original stable-diffusion, it uses the timesteps lenght as the inference step number.

# What does this PR do?
In [pipeline_stable_diffusion.py](https://github.com/huggingface/diffusers/blob/cbea5d1725a09e7f45a145bbeb4fab7539b0faa0/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py#L996), it uses the timesteps length as the inference steps. Usually the length of timesteps is not exactly equal to num_inference_steps, and is usually 1 greater than num_inference_steps. If we use num_inference_steps as the number of loops, then our results may not be the same as the original stable-diffusion results.

Fixes # (issue)

## Before submitting


regisss (commented):
LGTM!
Can you also run
```
pip install -U ruff
make style
```
please?

yuanwu2017:
> LGTM! Can you also run
> 
> ```
> pip install -U ruff
> make style
> ```
> 
> please?

Done

regisss (approved):

PR #997, opened 2024-05-21
Fix for ValueError in Whisper model CI test

MohitIntel:
Fixes the ValueError: `hpu_graphs` is True but `lazy_mode` is False. HPU graphs require `lazy_mode` to be set to True.
seen in the eval mode of speech_recognition whisper model CI test on this branch.

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -471,7 +471,7 @@ class GaudiGenerationMixin(GenerationMixin):
         return generation_config
 
     def _prepare_generation_config(
-        self, generation_config: GaudiGenerationConfig, lazy_mode: bool, **kwargs: Dict
+        self, generation_config: GaudiGenerationConfig, **kwargs: Dict
     ) -> Tuple[GaudiGenerationConfig, Dict]:
         """
         Copied from https://github.com/huggingface/transformers/blob/v4.40.2/src/transformers/generation/utils.py#L1230
@@ -530,7 +530,7 @@ class GaudiGenerationMixin(GenerationMixin):
                     )
             self.generation_config.static_shapes = generation_config.static_shapes
             if generation_config.ignore_eos is None:
-                generation_config.ignore_eos = kwargs.get("ignore_eos", lazy_mode)
+                generation_config.ignore_eos = kwargs.get("ignore_eos", kwargs.get("lazy_mode", None))
                 self.generation_config.ignore_eos = generation_config.ignore_eos
             model_kwargs = generation_config.update(**kwargs)  # All unused kwargs must be model kwargs
             if self.config.model_type == "falcon" and "token_type_ids" in kwargs.keys():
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/generation/utils.py
```diff
@@ -471,7 +471,7 @@ def _prepare_generated_length(
         return generation_config
 
     def _prepare_generation_config(
-        self, generation_config: GaudiGenerationConfig, lazy_mode: bool, **kwargs: Dict
+        self, generation_config: GaudiGenerationConfig, **kwargs: Dict
```
MohitIntel, line 474: The lazy_mode variable is not propogating downstream after this function call.
Replacing the function signature as in the original Transformer (https://github.com/huggingface/transformers/blob/4fdf58afb72b0754da30037fc800b6044e7d9c99/src/transformers/generation/utils.py#L1231) fixes this problem.

jiminha: @MohitIntel  can you check how it was working on main branch(transformer 4.38) without any issue?

MohitIntel: This function is originally called in https://github.com/huggingface/transformers/blob/main/src/transformers/models/whisper/generation_whisper.py#L484 with only two parameters : generation_config and **kwargs.
The lazy_mode variable is part of the kwargs dict in this call.
But, in the transformer_4.40 branch, this lazy_mode variable is getting **unpacked as a separate variable and gets popped from the kwargs dict.** 
Thus, it does not propagate downstream as part of kwargs in https://github.com/huggingface/transformers/blob/60bb571e993b7d73257fb64044726b569fef9403/src/transformers/models/whisper/generation_whisper.py#L591 as seen here from the debug logs:

transformers_4.40 **kwargs:
```python
(Pdb) {'labels': tensor([[50276, 50359, 50363,  ...,  -100,  -100,  -100],
        [50276, 50359, 50363,  ...,  -100,  -100,  -100],
        [50276, 50359, 50363,  ...,  -100,  -100,  -100],
        ...,
        [50276, 50359, 50363,  ...,  -100,  -100,  -100],
        [50276, 50359, 50363,  ...,  -100,  -100,  -100],
        [50276, 50359, 50363,  ...,  -100,  -100,  -100]], device='hpu:0'), 'hpu_graphs': True}
``` 
 v/s Fixed version **kwargs which contains the **'lazy_mode': True key-value pair** :
```python
 (Pdb) {'labels': tensor([[50276, 50359, 50363,  ...,  -100,  -100,  -100],
        [50276, 50359, 50363,  ...,  -100,  -100,  -100],
        [50276, 50359, 50363,  ...,  -100,  -100,  -100],
        ...,
        [50276, 50359, 50363,  ...,  -100,  -100,  -100],
        [50276, 50359, 50363,  ...,  -100,  -100,  -100],
        [50276, 50359, 50363,  ...,  -100,  -100,  -100]], device='hpu:0'), 'hpu_graphs': True, 'lazy_mode': True}
```

PR #1001, opened 2024-05-23
Sasarkar/transformers 4.40 patches for stopping criteria

ssarkar2:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/stopping_criteria.py
+++ b/optimum/habana/transformers/generation/stopping_criteria.py
@@ -33,7 +33,7 @@ def gaudi_MaxLengthCriteria_call(
 ) -> Union[torch.BoolTensor, bool]:
     token_idx = kwargs.get("token_idx", None)
     if token_idx is not None:
-        assert not kwargs["out_type_tensor"]
+        assert not kwargs["needs_tensor_output"]
         return token_idx >= self.max_length
     else:
         cur_len = input_ids.shape[-1]
@@ -52,7 +52,7 @@ def gaudi_MaxNewTokensCriteria_call(
 ) -> Union[torch.BoolTensor, bool]:
     token_idx = kwargs.get("token_idx", None)
     if token_idx is not None:
-        assert not kwargs["out_type_tensor"]
+        assert not kwargs["needs_tensor_output"]
         return token_idx >= self.max_length
     else:
         is_done = input_ids.shape[-1] >= self.max_length
@@ -61,7 +61,7 @@ def gaudi_MaxNewTokensCriteria_call(
 
 def gaudi_MaxTimeCriteria_call(self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs) -> Union[torch.BoolTensor, bool]:
     is_done = time.time() - self.initial_timestamp > self.max_time
-    if kwargs["out_type_tensor"]:
+    if kwargs["needs_tensor_output"]:
         return create_return_const_tensor(input_ids, is_done)
     else:
         return is_done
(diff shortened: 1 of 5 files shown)


optimum/habana/transformers/generation/stopping_criteria.py
```diff
@@ -37,26 +44,50 @@ def gaudi_MaxLengthCriteria_call(
...
+    else:
+        # token_idx is present, so we have static shapes, so using single boolean
+        False
 
 def gaudi_StoppingCriteriaList_call(
     self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs
-) -> torch.BoolTensor:
-    is_done = torch.full((input_ids.shape[0],), 0, device=input_ids.device, dtype=torch.int8)
+) -> Union[torch.BoolTensor, bool]:
+    kwargs["out_type_tensor"] = needs_tensor_output(kwargs.get("token_idx",None), kwargs.get("ignore_eos",True), kwargs.get("eos_token_id",None))
```
regisss, line 87:
```suggestion
    kwargs["needs_tensor_output"] = needs_tensor_output(kwargs.get("token_idx",None), kwargs.get("ignore_eos",True), kwargs.get("eos_token_id",None))
```
Let's simply call this kwarg the same

optimum/habana/transformers/generation/utils.py
```diff
@@ -115,6 +115,13 @@ def incrementor(bucket_size, prompt_len):
             "need_expansion": need_expansion,
         }
 
+def get_final_stopping_crit(x):
```
regisss, line 118:
```suggestion
def get_final_stopping_criteria(x):
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -115,6 +115,13 @@ def incrementor(bucket_size, prompt_len):
...
+    if type(x) == type(True):
```
regisss, line 119:
```suggestion
    if isinstance(x, bool):
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -115,6 +115,13 @@ def incrementor(bucket_size, prompt_len):
...
+        return x
+    elif torch.is_tensor(x):
+        return all(x)
+    else:
+        assert False
```
regisss, line 124:
```suggestion
        raise TypeError(f"The stopping criteria should be either a boolean or a torch.tensor but got {type(x)}.")
```

optimum/habana/transformers/generation/utils.py
```diff
@@ -743,6 +750,7 @@ def generate(
         if generation_config.static_shapes:
             # Pad inputs to have static shapes during generation, this gives better performance than dynamic shapes on HPUs
             # In encoder_decoder models, Inputs are already padded
+            assert generation_config.ignore_eos, "For gaudi we pad input_ids with EOS, so this option isnt available for now"
```
regisss, line 753: Not sure what you mean here. That `ignore_eos == False` is not compatible with static shapes?

ssarkar2: In some cases pad token and eos are same (like in gpt2). `ignore_eos==False` means a stopping criteria is to look for sentences generation eos token and stop. However if we pad with eos tokens, we will make a false stop. I put this assert under another check now `if generation_config.pad_token_id == generation_config.eos_token_id:`. For some models like llama where eos token != pad token, we can have ignore_eos=False

ssarkar2:
@regisss , do you know why style check isnt running for this PR?

regisss (approved):

regisss:
> @regisss , do you know why style check isnt running for this PR?

Because it is only called for PRs where the target branch is `main`.

PR #1019, opened 2024-05-29
Sasarkar/fix first token time

ssarkar2:
# What does this PR do?

In this PR: https://github.com/huggingface/optimum-habana/pull/899
We do not count the time between calling `model.generate` and the time that the first while loop starts

Also we need a sync point after first loop is done, if we are collecting timing data, else the model will not really execute and we will get a shorter first token time than reality

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -867,7 +867,7 @@ class GaudiGenerationMixin(GenerationMixin):
                 return_dict_in_generate=generation_config.return_dict_in_generate,
                 synced_gpus=synced_gpus,
                 streamer=streamer,
-                iteration_times=iteration_times,
+                hb_gen_time=hb_gen_time,
                 **model_kwargs,
             )
         if generation_mode == GenerationMode.GREEDY_SEARCH:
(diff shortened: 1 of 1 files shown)


hsubramony (approved):

optimum/habana/transformers/generation/utils.py
```diff
@@ -1449,9 +1455,7 @@ def greedy_search(
         if token_idx is not None:
             # Update cur_len in case of static shapes
             cur_len = token_idx.item()
-        if iteration_times is not None:
-            hb_gen_time = HabanaGenerationtime(iteration_times=iteration_times)
-            hb_gen_time.start()
+        first_done = False
```
regisss, line 1458: Let's use a more explicit variable name like `time_to_first_token_done` or so

regisss (commented):
Please run `make style`

optimum/habana/transformers/generation/utils.py
ssarkar2: Oh shiv already pushed a change. LGTM by me then.

regisss (approved):

PR #1029, opened 2024-05-31
fp8 textgen ci structure update

jiminha:
# What does this PR do?
Restructure fp8 text-gen CI structure to include more parameters including num_hpus, max_output_token, max_input_token, batch_size and etc.  Most of these tests were running with just batch1, we are updating this to cover more real test cases scenarios.


tests/test_text_generation_example.py
```diff
@@ -159,58 +154,58 @@ def _test_text_generation(
...
-
-                    with open(Path(tmp_dir) / "results.json") as fp:
-                        results = json.load(fp)
-
-                    # Ensure performance requirements (throughput) are met
-                    assert results["throughput"] >= (2 - TIME_PERF_FACTOR) * baseline
-                return
 
         command = [x for y in command for x in re.split(pattern, y) if x]
+        print(f"\n\nCommand to test: {' '.join(command)}\n")
```
regisss, line 208:
```suggestion
        print(f"\n\nCommand to test: {' '.join(command[:-2])}\n")
```
We need to avoid printing the token here too.

tests/test_text_generation_example.py
```diff
@@ -159,58 +154,58 @@ def _test_text_generation(
...
-            command = [x for y in command for x in re.split(pattern, y) if x]
-            subprocess.run(command, env=env_variables)
+            env_variables["TQDM_DISABLE"] = "1"
+            if measure_command:
+                measure_command.append(f"--token {token.value}")
+                env_variables["QUANT_CONFIG"] = os.path.join(
+                    path_to_example_dir, "text-generation/quantization_config/maxabs_measure_include_outputs.json"
+                )
+                measure_command = [x for y in measure_command for x in re.split(pattern, y) if x]
+                print(f"\n\nMeasure Command to test: {' '.join(measure_command)}\n")
```
regisss, line 200:
```suggestion
                print(f"\n\nMeasure Command to test: {' '.join(measure_command[:-2])}\n")
```
We need to avoid printing the token to prevent a leak.

tests/test_text_generation_example.py
```diff
@@ -159,58 +154,58 @@ def _test_text_generation(
...
+                subprocess.run(measure_command, env=env_variables)
```
regisss, line 201:
```suggestion
                proc = subprocess.run(measure_command, env=env_variables)
                
                # Ensure the run finished without any issue
                    # Use try-except to avoid logging the token if used
                    try:
                        assert proc.returncode == 0
                    except AssertionError as e:
                        if "'--token', 'hf_" in e.args[0]:
                            e.args = (f"The following command failed:\n{' '.join(measure_command[:-2])}",)
                        raise
```
This will make it easier to get the command to reproduce a potential issue.

tests/test_text_generation_example.py
```diff
@@ -159,58 +154,58 @@ def _test_text_generation(
...
+            command.insert(-2, "--flash_attention_recompute")
+
+        global prev_quant_model_name
+        global prev_quant_rank
+        measure_command = None
+        # FP8 Measurement only needed
+        if (not prev_quant_model_name) or (prev_quant_model_name != model_name) or (prev_quant_rank != world_size):
+            measure_command = [
+                x for x in command if not x.startswith("--max_new_tokens")
+            ]  # Remove max_new_tokens for mesurement
```
regisss, line 172:
```suggestion
            ]  # Remove max_new_tokens for measurement
```

tests/test_text_generation_example.py
```diff
@@ -159,58 +154,58 @@ def _test_text_generation(
...
+            measure_command = [
+                x if not x.startswith("--batch_size") else "--batch_size 1" for x in measure_command
+            ]  # Remove batch_size for mesurement
```
regisss, line 175:
```suggestion
            ]  # Remove batch_size for measurement
```

regisss (commented):
I left several comments, the main ones being about not printing the token used for gated models to avoid leaks.

Also, the baselines you updated were obtained with Synapse v1.16 and Transformers 4.40 right?

regisss (approved):

PR #1050, opened 2024-06-07
No shapes in profilings by default (#233)

astachowiczhabana:
1. Add option profiling_record_shapes in run_generation (default False) - reducing profiling files size
2. Remove option with_stack in profiler - reducing profiling files size

### DIFF
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -118,6 +118,8 @@ class GaudiTrainingArguments(TrainingArguments):
             Number of steps to ignore for profling.
         profiling_steps (`int`, *optional*, defaults to 0):
             Number of steps to be captured when enabling profiling.
+        profiling_record_shapes (`bool`, *optional*, defaults to True):
+            Record shapes when enabling profiling.
     """
 
     use_habana: Optional[bool] = field(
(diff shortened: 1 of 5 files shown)


optimum/habana/transformers/training_args.py
```diff
@@ -118,6 +118,8 @@ class GaudiTrainingArguments(TrainingArguments):
             Number of steps to ignore for profling.
         profiling_steps (`int`, *optional*, defaults to 0):
             Number of steps to be captured when enabling profiling.
+        profiling_record_shapes (`bool`, *optional*, defaults to True):
+            Record shapes when enabling profiling.
```
regisss, line 122: I don't think we should add this as there is no new argument defined in `training_args.py.

regisss (commented):
Please run `make style`

optimum/habana/transformers/training_args.py
libinta: where should we add this?

regisss: IT's already added in the docstrings of `generation/utils.py` and in the help section of the new arg of the text-generation example. This is good enough.

ssarkar2: removed

regisss (approved):

Issue #240, opened 2023-05-23
Can not run text-generation with bloom deepspeed?


ZhaiFeiyue:
### System Info

```shell
ubuntu 20.04
docker 1.9
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

1. refer https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation to setup
2. `python ../gaudi_spawn.py --use_deepspeed --world_size 8 run_generation.py --model_name_or_path bigscience/bloom-560m --bf16 --max_new_tokens 10 --batch_size 1  --use_kv_cache --do_sample`

### Expected behavior

running correctly

ZhaiFeiyue:
@regisss I'm now trying to run text-generation with  bloom in deepspeed mode, but error comes 
![image](https://github.com/huggingface/optimum-habana/assets/80079571/8cb33f74-00be-44d2-a382-b09dd29c4419)

after debug I found the issue is relevant to `token_idx` which is a scalar tensor cannot be expanded.

the model I used is bloom-560m since bloom is too large to do testing.

PR #1064, opened 2024-06-11
Fix lm_evaluation_harness to specific commit (#240)

astachowiczhabana:
* Fix lm_evaluation_harness to specific commit

* Different commit in requirements_lm_eval

Why this change?
1. We should have fixed reference instead of branch which could change in the future
2. By fixing to specific commit hash we make pip download requirements.txt a lot faster


regisss (approved):
LGTM

PR #1068, opened 2024-06-12
Add rouge metric evalution for llama 70B with orca datasets

sushildubey171:
* Add rouge metric evalution for llama 70B with orca datasets

use rouge metric to evaluate the corretness of the model, it uses openorca dataset


examples/text-generation/evaluation.py
```diff
@@ -0,0 +1,115 @@
...
+###################### Habana internal code ##################################
+ACC_TARGET = {"rouge1": 44.4312, "rouge2": 22.0352, "rougeL": 28.6162}
+
+# See https://github.com/mlcommons/inference/pull/1583
+##############################################################################
+
+def get_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--checkpoint-path", default="/mnt/weka/data/pytorch/llama2/Llama-2-70b-chat-hf",
+                        help="Path to Llama2-70b-hf-chat checkpoint")
```
libinta, line 17: please remove the default path as current. and put None

examples/text-generation/evaluation.py
```diff
@@ -0,0 +1,115 @@
...
+    parser.add_argument("--accuracy-file", default="output/accuracy.json", help="path to accuracy.json")
+    parser.add_argument("--dataset-file", default="/mnt/weka/data/mlperf_inference/llama2/processed-data.pkl",
+                        help="path to processed openorca validation set")
```
libinta, line 20: remove the default as this

examples/text-generation/run_generation.py
```diff
@@ -85,6 +87,12 @@ def setup_parser(parser):
         type=str,
         help="Optional argument if you want to assess your model on a given dataset of the HF Hub.",
     )
+    parser.add_argument(
+        "--dataset",
+        default="/mnt/weka/data/mlperf_inference/llama2/processed-data.pkl",
```
libinta, line 92: remove default, and put None, but do a check later

examples/text-generation/run_generation.py
```diff
@@ -313,8 +321,164 @@ def main():
...
+                })
+            return queries
+
+        ds = get_ds(args)
+        input_sentences = get_input(ds, args.batch_size)
+
+        def generate(input_tokens, size=None, reduce_recompile=False):
+            """Generates sequences from the input sentences and returns them."""
+
+            t0 = time.perf_counter()
```
libinta, line 355: please check how the 1st token latency is done here, and do the same https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/run_generation.py#L381

examples/text-generation/run_generation.py
```diff
@@ -313,8 +321,164 @@ def main():
...
+                hpu_graphs=args.use_hpu_graphs,
+                profiling_steps=args.profiling_steps,
+                profiling_warmup_steps=args.profiling_warmup_steps,
+            ).cpu()
+            outputs = outputs.tolist()
+            for i in range(len(outputs)):
+                outputs[i] = outputs[i][args.max_input_tokens:]
+            duration = time.perf_counter() - t0
+            print(f"Total E2E time of this batch is {duration:.3f}s", flush=True)
+            return outputs
```
ssarkar2, line 379: This is a LOT of code duplication, which can cause errors later due to failing to remember to propagate chaneges in both branches etc.


For example, you could reuse the older "generate" function, just by adding:
```
def generate(input_tokens=None, size=None, reduce_recompile=False):
.....
    if input_tokens is None:  # ADDING THIS
				# Tokenization
				if args.max_input_tokens > 0:
					input_tokens = tokenizer.batch_encode_plus(
						input_sentences,
						return_tensors="pt",
						padding="max_length",
						max_length=args.max_input_tokens,
						truncation=True,
					)
				else:
					input_tokens = tokenizer.batch_encode_plus(input_sentences, return_tensors="pt", padding=True)
```

And you could perform the output runcation outside generate:
```
for i in range(len(outputs)):
                outputs[i] = outputs[i][args.max_input_tokens:]
```

examples/text-generation/run_generation.py
```diff
@@ -313,8 +321,164 @@ def main():
...
 
-    if args.dataset_name is None:
+        # compilation stage disable profiling
+        HabanaProfile.disable()
+        # Compilation
+        logger.info("Graph compilation...")
+        dyn_prompt_lens = args.simulate_dyn_prompt
+        t0 = time.perf_counter()
+        # The first three iterations take longer because of graph compilation
+        if dyn_prompt_lens is None or len(set(dyn_prompt_lens)) == 1:
```
ssarkar2, line 390: There is no difference between these code and the already existing one in the else branch (except for the call to generate? If not lets not duplicate. You can just call generate differently:

`generate(input_sentences[0] if args.dataset_name == "openorca" else None, dyn_prompt_lens[0], args.reduce_recompile)`

examples/text-generation/run_generation.py
```diff
@@ -313,8 +321,164 @@ def main():
...
+        total_new_tokens_generated = 0
+        logger.info("Running generate...")
+        t0 = time.perf_counter()
+        # Benchmark over n_iterations iterations
+        N = len(input_sentences)
+        if dyn_prompt_lens is None:
+            for i in range(args.n_iterations):
+                results = []
+                b = 1
+                for sentence in input_sentences:
```
ssarkar2, line 425: Minor: More pythonic to write:
`for b, sentence in enumerate(input_sentences)`
then we can remove the b=1, b+=1 lines

examples/text-generation/run_generation.py
```diff
@@ -313,8 +321,164 @@ def main():
...
+                        generate(input_sentences[0], sz - 1, args.reduce_recompile)
+        torch_hpu.synchronize()
+        compilation_duration = time.perf_counter() - t0
+        HabanaProfile.enable()
+        total_new_tokens_generated = 0
+        logger.info("Running generate...")
+        t0 = time.perf_counter()
+        # Benchmark over n_iterations iterations
+        N = len(input_sentences)
+        if dyn_prompt_lens is None:
```
ssarkar2, line 421: What is the purpose of the dataset here? I suppose it is for accuracy eval. Then Why warmup? Once you have gone thru the dataset once and collected the sentences for accuracy, you dont need to go over the dataset n_iter times again, as far as I understand.

ssarkar2: Similarly teh whole apparatus of dynamic prompts warmup etc are also probably not used for orca eval? in which case we should delete all these extraneous if-elses that dynamic prompt gives rise to

ssarkar2 (changes requested):
This change is better suited in a separate file.

run_generation is meant for perf eval. Its apparatus for warmup and iterating multiple times over a same sentence is not needed for an accuracy eval usecase.

Also current PR has too much code duplications, and unused code paths (dynamic prompts branches)

examples/text-generation/evaluation.py
sushildubey171: default file name is to help running with correct dataset file since it is already preprocessed, no other dataset is used here. user may end up running with incorrect or may not find the correct dataset if removed.

examples/text-generation/run_generation.py
sushildubey171: same as above

examples/text-generation/run_generation.py
sushildubey171: please note this for evaluation and not for latency test still existing behaviour is retained

examples/text-generation/run_generation.py
sushildubey171: please note we need to run the measurement also, we need other functionalities from this file therefore existing behaviours are retained, if any code duplication can be handled as part of later cleanup/code refactoring.

examples/text-generation/run_generation.py
sushildubey171: there are difference related the data loading from different datasets and invoking the generate functions on openorca datasets in the for loop

examples/text-generation/run_generation.py
sushildubey171: two different datasets one for generating measurement and one for running the quantization for evaluation. Please note here number of iteration is 1 for running this evaluation, we have already specified the command for running the evaluation, I have kept to retain the n_itr argument. dynamic path is yet no tested for rouge eval, need to check with validation team else it can be clean up as part of code refactoring.

examples/text-generation/evaluation.py
```diff
@@ -0,0 +1,115 @@
+import argparse
+from transformers import AutoTokenizer
+import nltk
+import evaluate
+import numpy as np
+import json
+
+###################### Habana internal code ##################################
```
libinta, line 8: can you put proper head?

PR #1070, opened 2024-06-12
Fix throughput calculation in image-to-text example

regisss:
# What does this PR do?

The throughput calculation in the image-to-text example is currently misleading as it assumes that `args.max_new_tokens` will always be generated. However, this is not necessarily the case when setting `ignore_eos` to True.

This PR fixes that and adds a new `ignore_eos` argument.

## Before submitting

### DIFF
--- a/examples/image-to-text/run_pipeline.py
+++ b/examples/image-to-text/run_pipeline.py
@@ -86,6 +86,12 @@ def main():
     parser.add_argument("--batch_size", type=int, default=1, help="Input batch size.")
     parser.add_argument("--warmup", type=int, default=3, help="Number of warmup iterations for benchmarking.")
     parser.add_argument("--n_iterations", type=int, default=5, help="Number of inference iterations for benchmarking.")
+    parser.add_argument(
+        "--ignore_eos",
+        default=True,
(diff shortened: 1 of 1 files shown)


examples/image-to-text/run_pipeline.py
```diff
@@ -86,6 +86,12 @@ def main():
     parser.add_argument("--batch_size", type=int, default=1, help="Input batch size.")
     parser.add_argument("--warmup", type=int, default=3, help="Number of warmup iterations for benchmarking.")
     parser.add_argument("--n_iterations", type=int, default=5, help="Number of inference iterations for benchmarking.")
+    parser.add_argument(
+        "--ignore_eos",
+        default=True,
```
mandy-li, line 91: I would prefer the default is False for backward compatibility

mandy-li (commented):
should we set default to False?

examples/image-to-text/run_pipeline.py
regisss: It was designed to do the opposite but I'm fine with either way. I just pushed a new commit so that it behaves as you described.

mandy-li (approved):
LGTM

PR #1055, opened 2024-06-09
Fix the error of running run_pipeline.py of text_generation example

yuanwu2017:
# What does this PR do?

python run_pipeline.py --model_name_or_path mistralai/Mistral-7B-v0.3  --use_kv_cache --max_new_tokens 20 --prompt "He is working on"
Fix the following error:
06/09/2024 04:11:28 - INFO - __main__ - device: hpu, n_hpu: 0, bf16: False
06/09/2024 04:11:28 - INFO - __main__ - Model initialization took 10.879s
Traceback (most recent call last):
  File "/workspace/optimum-habana/examples/text-generation/text-generation-pipeline/run_pipeline.py", line 63, in <module>
    main()
  File "/workspace/optimum-habana/examples/text-generation/text-generation-pipeline/run_pipeline.py", line 44, in main
    pipe = GaudiTextGenerationPipeline(args, logger)
  File "/workspace/optimum-habana/examples/text-generation/text-generation-pipeline/pipeline.py", line 8, in __init__
    self.model, self.tokenizer, self.generation_config = initialize_model(args, logger)
ValueError: too many values to unpack (expected 3)

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1079, opened 2024-06-14
fix bug in initialize_model in text-generation-pipeline

rbrugaro:
# fix ValueError caused by setting wrong number of returned values in initialize_model 

example to reproduce the error:
python optimum-habana/examples/text-generation/text-generation-pipeline/run_pipeline.py --model_name_or_path meta-llama/Llama-2-7b-hf --use_hpu_graphs --use_kv_cache --max_new_tokens 100 --do_sample --prompt "Here is my prompt"

error:     
self.model, self.tokenizer, self.generation_config = initialize_model(args, logger)
ValueError: too many values to unpack (expected 3)

initialize_model returns 4 values (model, assistant_model, tokenizer, generation_config)


regisss:
Hi @rbrugaro, this is a known issue which should be fixed now as I just merged #1055.

PR #1086, opened 2024-06-18
Change the device_id for FSDP plugin

ckvermaAI:
# What does this PR do?

1) Return a ``torch.device`` for the specified ``device_id`` (This change is required for FSDP).
2) Similar to CUDA variant [_get_device_from_device_id](https://github.com/pytorch/pytorch/blob/main/torch/distributed/fsdp/_init_utils.py#L795).  

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/accelerate/accelerator.py
+++ b/optimum/habana/accelerate/accelerator.py
@@ -433,7 +433,7 @@ class GaudiAccelerator(Accelerator):
                         "param_init_fn": fsdp_plugin.param_init_fn,
                         "ignored_modules": fsdp_plugin.ignored_modules,
                         "limit_all_gathers": fsdp_plugin.limit_all_gathers,
-                        "device_id": torch.device("hpu"),
+                        "device_id": torch.device("hpu", torch.hpu.current_device()),
                     }
                     model = FSDP(model, **kwargs)
                     if fsdp_plugin.activation_checkpointing:
--- a/optimum/habana/accelerate/utils/dataclasses.py
+++ b/optimum/habana/accelerate/utils/dataclasses.py
@@ -142,5 +142,5 @@ class GaudiFullyShardedDataParallelPlugin(FullyShardedDataParallelPlugin):
         self.activation_checkpointing = str_to_bool(os.environ.get(prefix + "ACTIVATION_CHECKPOINTING", "False")) == 1
 
         if self.sync_module_states:
-            device = torch.device("hpu")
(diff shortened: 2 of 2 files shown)


vivekgoe:
@regisss @libinta This PR is needed for FSDP to work with Synapse 1.17.  Added new label "Synapse 1.17_dependency".

yafshar:
@ckvermaAI @vivekgoe how do you test this PR. Is "hpu:X" notation is working as expected on 1.17.0?

ckvermaAI:
@yafshar I've tested this PR for MPT (1B & 70B) in both eager & torch.compile **with FSDP**. I don't know the behavior of this API in allocating the tensor on a particular rank.

cc @vivekgoe

vivekgoe:
@yafshar This change is applicable only for FSDP related code where some parts of the code expect hpu:X notation. For any other use-case, such as example you are trying, correct way is to use 'hpu' or 'hpu:0' irrespective of device id.

yafshar:
> @yafshar I've tested this PR for MPT (1B & 70B) in both eager & torch.compile **with FSDP**. I don't know the behavior of this API in allocating the tensor on a particular rank.
> 
> cc @vivekgoe

@ckvermaAI can you add your testing command.

yafshar:
> @yafshar This change is applicable only for FSDP related code where some parts of the code expect hpu:X notation. For any other use-case, such as example you are trying, correct way is to use 'hpu' or 'hpu:0' irrespective of device id.

@vivekgoe if the functionality is not available yet, then what difference does it make to use `device = torch.device("hpu", torch.hpu.current_device())` or `device = torch.device("hpu")`. It makes the current PR not useful for the current stack. Later you are right, it is necessary for FSDP, but now it does not use it! Am I missing st here?

vivekgoe:
@yafshar Change in this PR is required to align with changes done to FSDP code changes in SynapseAI 1.17.0 release. I have added a label "Synapse 1.17_dependency" indicating the same, which means this should be merged only after SynapseAI 1.17.0 is available. Without these changes FSDP test-cases (such as https://github.com/huggingface/optimum-habana/blob/59d182dfa9a4c0ba068a1fd118b96b8ff0057c96/tests/test_examples.py#L793) will break. You can check this test with the PR once Synapse AI 1.17.0 is available.

yafshar:
@vivekgoe I checked my simple example on 1.17.0 and it did not work, but will check this PR in the test FSDP and will update here

regisss (approved):
LGTM!

PR #1096, opened 2024-06-25
TE FP8 integration

SanjuCSudhakaran:
# What does this PR do?

Refactoring `transformer_engine` FP8 support in optimum-habana. This removes model specific changes and makes the code more model agnostic.

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/language-modeling/README.md
+++ b/examples/language-modeling/README.md
@@ -574,6 +574,46 @@ python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
   --flash_attention_causal_mask True
 ```
 
+- Multi-card finetuning of Llama2-70B with DeepSpeed ZeRO-3 optimization, LoRA and FP8 precision:
+
+  > The following command requires Habana DeepSpeed 1.13.0 or later.
+
+```bash
+PT_HPU_MAX_COMPOUND_OP_SIZE=10 \
+python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
+  --model_name_or_path meta-llama/Llama-2-70b-hf \
+  --deepspeed llama2_ds_zero3_config.json \
+  --dataset_name tatsu-lab/alpaca \
+  --bf16 True \
+  --output_dir ./lora_out \
+  --num_train_epochs 2 \
+  --max_seq_len 2048 \
+  --per_device_train_batch_size 10 \
+  --per_device_eval_batch_size 10 \
+  --gradient_checkpointing \
+  --evaluation_strategy epoch \
+  --eval_delay 2 \
+  --save_strategy no \
+  --learning_rate 0.0018 \
+  --warmup_ratio 0.03 \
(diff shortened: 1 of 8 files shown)


examples/language-modeling/README.md
```diff
@@ -574,6 +574,46 @@ python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
   --flash_attention_causal_mask True
 ```
 
+- Multi-card finetuning of Llama2-70B with DeepSpeed ZeRO-3 optimization, LoRA and FP8 precision:
```
vivekgoe, line 577: Maybe we can just modify existing command for FP8 precision (instead of adding a new one).

optimum/habana/transformers/training_args.py
```diff
@@ -291,14 +291,6 @@ class GaudiTrainingArguments(TrainingArguments):
         metadata={"help": "Whether to use fp8 for training."},
     )
 
-    fp8_recipe_format: Optional[str] = field(
```
vivekgoe, line 294: @libinta please note we deleted this because we do not want to expose low level numeric selection to user as of now. In future we can revisit this.

vivekgoe:
@regisss @libinta please help review this PR. This is a cleaned up version of TE integration (tested with Llama-7B, Llama-70B, Mixtral-7B fine-tuning tasks).  Changes were done mainly for following reasons,
-  Align to newer PEFT version (0.10.0) which helps reducing custom changes to support LoRA layers in fp8,
-  Add generic fp8 support for models using activation_checkpointing, 
-  Align TE code to accelerator repo to extent possible. Limit fp8 configuration interface exposed to user (allow only well tested configuration for now)

vivekgoe:
Related PR with tests https://github.com/huggingface/optimum-habana/pull/1106

examples/language-modeling/README.md
```diff
@@ -574,6 +574,46 @@ python3 ../gaudi_spawn.py --use_deepspeed  --world_size 8  run_lora_clm.py \
...
+
+  > The following command requires Habana DeepSpeed 1.13.0 or later.
+
+```bash
+PT_HPU_MAX_COMPOUND_OP_SIZE=10 \
```
libinta, line 582: why do we still need PT_HPU_MAX_COMPOUND_OP_SIZE=10 ?

SanjuCSudhakaran: There is a spike in the loss curve when the flag is removed, which results in a slight drop in overall accuracy for Llama2-70B.

So for now we want to keep this flag, while we investigate and fix this issue.

examples/language-modeling/README.md
regisss: I agree with Vivek

SanjuCSudhakaran: Done. Updated the readme.

vivekgoe (approved):

libinta (approved):

regisss (approved):

PR #1099, opened 2024-06-25
Adds a helm chart, dockerfile, and instructions for running examples using a Kubernetes cluster

dmsuehir:
# What does this PR do?

This PR adds a helm chart, Dockerfile/docker compose, and instructions for running the optimum-habana examples using Kubernetes.

I added an `examples/kubernetes` directory, which is similar to `examples/multi-node-training` in that it's a general example that can be used with any of the other examples.

Since the docker images isn't published to a repository anywhere at this time, the user will need to build and push the container before deploying the helm chart. The Dockerfile uses the Habana Gaudi docker image and it's base and then installs optimum-habana, Habana's fork of DeepSpeed, and a `requirements.txt` file. The `requirements.txt` file is just installing the `requirements.txt` file from the other examples directories (like `language-modeling/requirements.txt`). The user can adjust this depending on which example they want to run.

The helm chart can be used to run single card and multi card jobs on a single node. I can post future update for multi node jobs using the kubeflow MPIJob (I don't have access to multiple nodes at this time, so I don't have it yet).

## Before submitting

I manually tested this on a vanilla kubernetes cluster with a Gaudi2 node and version 1.16.1 of the Habana software/driver. The use cases that I validated are noted in the README.


examples/kubernetes/Chart.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v2
+name: Deploy Optimum for Intel® Gaudi® Accelerators Examples to Kubernetes
```
yafshar, line 2: @dmsuehir, please follow the best practices for the name, here you added description instead of the name. Also the directory containing the chart must have the same name as the chart itself

examples/kubernetes/Chart.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
```
yafshar, line 13:
```suggestion
```
Is there a need for the type?

examples/kubernetes/Chart.yaml
```diff
@@ -0,0 +1,25 @@
...
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.12.0"
+
```
yafshar, line 25: ! suggestion is to add the minimum Kubernetes version that you have tested with it. I usually find having `kubeVersion: ` very useful here.

examples/kubernetes/Dockerfile
```diff
@@ -0,0 +1,21 @@
+ARG BASE_IMAGE_NAME=vault.habana.ai/gaudi-docker/1.16.1/ubuntu22.04/habanalabs/pytorch-installer-2.2.2
+ARG BASE_IMAGE_TAG=latest
+
+FROM ${BASE_IMAGE_NAME}:${BASE_IMAGE_TAG} AS optimum-habana
```
yafshar, line 4:
```suggestion
FROM vault.habana.ai/gaudi-docker/1.16.1/ubuntu22.04/habanalabs/pytorch-installer-2.2.2:latest AS optimum-habana
```

Personal style, but I am wondering if you need to introduce these extra arguments?

examples/kubernetes/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+ARG GAUDI_SW_VER=1.16.1
```
yafshar, line 6: if you introduce this argument then you should use it every where! even on your base

examples/kubernetes/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ARG OPTIMUM_HABANA_VER=1.12.0
+
+RUN pip install --no-cache-dir optimum-habana==${OPTIMUM_HABANA_VER} && \
+    pip install --no-cache-dir git+https://github.com/HabanaAI/DeepSpeed.git@${GAUDI_SW_VER}
+
+FROM optimum-habana AS optimum-habana-examples
+
+ARG OPTIMUM_HABANA_VER=1.12.0
```
yafshar, line 14:
```suggestion
```

No need to reintroduce the arg again

examples/kubernetes/Dockerfile
dmsuehir: My thought was that a user might have a different version of the Habana software on their machine, so instead of asking them to edit the Dockerfile, I defined it as an arg. Also, having a `BASE_IMAGE_NAME` and `BASE_IMAGE_TAG` allows for doing testing with a different base image (for example a pre-release image that is in a local/internal container registry). 

However, based on your other comment, it sounds like you would prefer something like this?
```
ARG GAUDI_SW_VER=1.16.1
ARG OS=ubuntu22.04
ARG TORCH_VER=2.2.2

FROM vault.habana.ai/gaudi-docker/${GAUDI_SW_VER}/${OS}/habanalabs/pytorch-installer-${TORCH_VER}:latest AS optimum-habana
```

examples/kubernetes/requirements.txt
```diff
@@ -0,0 +1,3 @@
+huggingface_hub==0.23.0
```
yafshar, line 1: @dmsuehir we have pinned a version < 0.23.0 for `huggingface_hub`. Is there a reason you need this specific version?

examples/kubernetes/docker-compose.yaml
```diff
@@ -0,0 +1,34 @@
+services:
+  optimum-habana:
+    build:
+      args:
+        http_proxy: ${http_proxy}
+        https_proxy: ${https_proxy}
+        no_proxy: ""
+        BASE_IMAGE_NAME: ${BASE_IMAGE_NAME:-vault.habana.ai/gaudi-docker/1.16.1/ubuntu22.04/habanalabs/pytorch-installer-2.2.2}
+        BASE_IMAGE_TAG: ${BASE_IMAGE_TAG:-latest}
+        GAUDI_SW_VER: 1.16.1
```
yafshar, line 10:
```suggestion
        GAUDI_SW_VER: ${GAUDI_SW_VER:-1.16.1}
```

examples/kubernetes/docker-compose.yaml
```diff
@@ -0,0 +1,34 @@
...
+        OPTIMUM_HABANA_VER: 1.12.0
```
yafshar, line 11:
```suggestion
        OPTIMUM_HABANA_VER: ${OPTIMUM_HABANA_VER:-1.12.0}
```

yafshar:
@dmsuehir I will test this PR in a day or 2 and will add more comments here. Thanks

examples/kubernetes/Dockerfile
dmsuehir: To use the arg in multiple stages, it needs to be redefined (source: https://docs.docker.com/reference/dockerfile/#scope). However, I can try making it just `ARG OPTIMUM_HABANA_VER` (without the `=1.12.0`)

examples/kubernetes/Dockerfile
yafshar: @dmsuehir thanks, if you are using ARGS, then I usually find it easier to have them at the start, like here and use them everywhere.

examples/kubernetes/requirements.txt
dmsuehir: Yes, [`HF_TOKEN_PATH`](https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#hftokenpath) was introduced in 0.23.0. The alternative is using `HF_HOME`, but that ended up being problematic because mounting the token to that location made `HF_HOME` read only. That caused an issue when other things tried to write to `HF_HOME` (like model/dataset files).

examples/kubernetes/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Requirements
+
+### Client System Requirements
+
+In order to build the docker images and deploy a job to Kubernetes you will need:
+
+* [Docker](https://docs.docker.com/engine/install/)
+* [Docker compose](https://docs.docker.com/compose/install/)
+* [kubectl](https://kubernetes.io/docs/tasks/tools/) installed and configured to access a cluster
```
yafshar, line 17:
```suggestion
* [kubectl](https://kubernetes.io/docs/tasks/tools/) installed and configured to access a cluster
* [Minikube](https://minikube.sigs.k8s.io/docs/start/) installed and configured to access a Kubernetes cluster on a single node
```

@dmsuehir would you provide a pointer for anyone using a single node and wants to test your PR to use Minikube.

```
If you are intending to run a Kubernetes cluster on a single node (locally), you will need a tool like [Minikube](https://minikube.sigs.k8s.io/docs/start/) to be installed first
```

examples/kubernetes/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG GAUDI_SW_VER=1.16.1
+ARG OS=ubuntu22.04
+ARG TORCH_VER=2.2.2
```
yafshar, line 3:
```suggestion
ARG TORCH_VER=2.2.2
ARG OPTIMUM_HABANA_VER=1.12.0
```

examples/kubernetes/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+FROM vault.habana.ai/gaudi-docker/${GAUDI_SW_VER}/${OS}/habanalabs/pytorch-installer-${TORCH_VER}:latest AS optimum-habana
+
+ARG GAUDI_SW_VER=1.16.1
+ARG OPTIMUM_HABANA_VER=1.12.0
```
yafshar, line 8:
```suggestion
ARG GAUDI_SW_VER
ARG OPTIMUM_HABANA_VER
```

dmsuehir:
@yafshar  Thanks, I've pushed those changes

libinta:
@dmsuehir can you update based on review comments?

dmsuehir:
> @dmsuehir can you update based on review comments?

Is there a certain comment in particular? I thought that I have address all of the comments.

yafshar:
I have one more test to do before finishing the review. Will update by tomorrow

yafshar (approved):
LGTM! 

@regisss could you please look at this PR when you have time!

regisss (approved):
LGTM!

PR #1102, opened 2024-06-26
Adding fused qkv support along with config

bhargaveede:
Support for fused_qkv in llama.
fused_qkv is single linear for all 3 q,k and v proj instead of separate linears.

Add "fused_qkv": true to config.json of model file to enable.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/llama/configuration_llama.py
+++ b/optimum/habana/transformers/models/llama/configuration_llama.py
@@ -26,6 +26,7 @@ class LlamaConfig(LlamaConfig):
         attention_bias=False,
         attention_dropout=0.0,
         mlp_bias=False,
+        fused_qkv=False,
         **kwargs,
     ):
         super().__init__(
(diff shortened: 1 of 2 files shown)


bhargaveede:
@regisss Please review the changes.

These are needed to bring fused_qkv changes to optimum-habana.
So that others can reproduce the MLperf results on Gaudi after 1.17 is released
By this we can have latest changes for MLPerf benchmark without needed to rebase/cherry-pick.

libinta (commented):
@bhargaveede  which training use case actually use it? if training used fused version, how the inference without fused version?

bhargaveede:
@libinta We use it for MLPerf training case and for inference, model is different and hence it won't use fused_version.
By default, we don't use fused_version. We've to explicitly change the config of model to use this.
Numerically, there is no difference. It's just the parameters are fused together instead of separate parameters each for q,k and v.

Model : https://huggingface.co/regisss/llama2-70b-fused-qkv-mlperf

libinta (approved):

regisss (approved):

PR #1106, opened 2024-06-27
Add a test for llama finetuning with FP8 precision

SanjuCSudhakaran:
# What does this PR do?

Adding example test case for the combination of Llama2-7B + MPI + LoRA + FP8 precision

Fixes # (issue)

## Before submitting

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -192,7 +192,7 @@ class ExampleTestMeta(type):
     """
 
     @staticmethod
-    def to_test(model_name: str, multi_card: bool, deepspeed: bool, example_name: str, fsdp: bool):
+    def to_test(model_name: str, multi_card: bool, deepspeed: bool, example_name: str, fsdp: bool, fp8: bool):
         models_with_specific_rules = [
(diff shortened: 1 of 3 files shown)


tests/test_examples.py
```diff
@@ -241,7 +241,7 @@ def to_test(model_name: str, multi_card: bool, deepspeed: bool, example_name: st
             return True
         elif "bridgetower" in model_name and IS_GAUDI2:
             return True
-        elif "falcon" in model_name and IS_GAUDI2 and not fsdp:
+        elif "falcon" in model_name and IS_GAUDI2 and not fsdp and not fp8:
             return True
         elif "bloom" in model_name and deepspeed and not IS_GAUDI2:
```
regisss, line 246: We should modify line 211 above as follows:
```
if (fsdp or fp8) and not IS_GAUDI2:
```
to make sure the test will not be run on Gaudi1.

SanjuCSudhakaran: Done

regisss (approved):

PR #1123, opened 2024-07-04
Update examples batchsize in README

shepark:
Update examples' batchsize referring to [Performance Data](https://developer.habana.ai/get-started/habana-models-performance/) and validation.

## Before submitting

### DIFF
--- a/examples/stable-diffusion/README.md
+++ b/examples/stable-diffusion/README.md
@@ -31,7 +31,7 @@ python text_to_image_generation.py \
     --model_name_or_path runwayml/stable-diffusion-v1-5 \
     --prompts "An image of a squirrel in Picasso style" \
     --num_images_per_prompt 20 \
-    --batch_size 4 \
+    --batch_size 7 \
     --image_save_dir /tmp/stable_diffusion_images \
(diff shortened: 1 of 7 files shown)


examples/stable-diffusion/README.md
```diff
@@ -31,7 +31,7 @@ python text_to_image_generation.py \
     --model_name_or_path runwayml/stable-diffusion-v1-5 \
     --prompts "An image of a squirrel in Picasso style" \
     --num_images_per_prompt 20 \
```
dsocek, line 33: @shepark @regisss If  If we use bs=7, we shouldn't we also set `--num_images_per_prompt` to a multiple of `7`, say `28` so sample runs 4 full batches here, since warmup is 3?

regisss: Yes, that does make sense.
@shepark You would get better throughput after warmup so I agree with @dsocek that running at least 4 steps should be favored.

imangohari1: I am opening a follow up PR on this. 
Will change the `--num_images_per_prompt` in this readme as well.

PR #1028, opened 2024-05-31
sarkar/cache_optimization

ssarkar2:
# What does this PR do?
PR-942 brought upto speed to main, with some fixes

test using:
```
python run_generation.py --model_name_or_path /mnt/weka/data/llama_inference/Llama-2-7b-hf --use_hpu_graphs --use_kv_cache --max_input_tokens 1024 --max_new_tokens 3072 --batch_size 39 --attn_softmax_bf16 --trim_logits --bf16 --warmup 2 --n_iterations 2 --limit_hpu_graphs --bucket_internal --bucket_size 128
```
note no `--reuse_cache` in above cmd

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1137, opened 2024-07-16
Support `bucket_internal` for MPT

pk1d3v:
# What does this PR do?

Adds support for `--bucket_internal` for MPT model.

Measurements:
| Param                                  | In/Out tokens | Bath size | Throughput t/s |  Memory allocated |
|-------------------------------|----------------|------------|----------------|-------------------|
| `--bucket_internal`               | 128/128         | 128           | 6019.4            | 37.04 GB           |
| **NO** `--bucket_internal`   | 128/128         | 128           | 2919.8           | 37.28 GB           |
| `--bucket_internal`               | 128/256         | 64             | 4268.5           | 31.28 GB           |
| **NO** `--bucket_internal`   | 128/256         | 64            | 3265.1           | 37.03 GB          |
| `--bucket_internal`               | 128/512         | 32             | 2529.7           | 28.03 GB           |
| **NO** `--bucket_internal`   | 128/512         | 32            | 2543.2          | 43.16 GB          |

Command line used: `python run_generation.py --model_name_or_path mosaicml/mpt-7b --use_hpu_graphs --use_kv_cache --max_input_tokens 128 --max_new_tokens <num> --trim_logits --bf16 --warmup 2 --n_iterations 2 --limit_hpu_graphs --batch_size=<num> --bucket_size 128 --bucket_internal`

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/mpt/modeling_mpt.py
+++ b/optimum/habana/transformers/models/mpt/modeling_mpt.py
@@ -67,9 +67,9 @@ def gaudi_mpt_attention_forward(
             else:
                 key_states = torch.cat([past_key_value[0], key_states], dim=2)
                 value_states = torch.cat([past_key_value[1], value_states], dim=2)
-        past_key_value = (key_states, value_states)
+        past_key_value = [key_states, value_states]
     else:
-        past_key_value = (key_states, value_states)
+        past_key_value = [key_states, value_states]
 
     attention_scores = torch.matmul(query_states, key_states.transpose(-1, -2)) * self.softmax_scale
 
@@ -289,7 +289,9 @@ class GaudiMptForCausalLM(MptForCausalLM):
         - add new args token_idx
         - add token_idx into model_inputs
         - from step2 when enable KV cache, slice next_input_ids from input_ids base on the token_idx
+        - support for internal bucketing
         """
+        bucket_internal = kwargs.get("bucket_internal")
         # only last tokens for input_ids if past is not None
         if past_key_values is not None:
             if token_idx is None:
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/mpt/modeling_mpt.py
```diff
@@ -305,6 +307,13 @@ def prepare_inputs_for_generation(
                 input_ids = input_ids[:, remove_prefix_length:]
             else:
                 input_ids = torch.index_select(input_ids, 1, token_idx - 1)
+            # Converting back to tuples as it should be, so there's no type mismatch when calling graph
+            past_key_values = tuple([tuple(kv) for kv in past_key_values])
+        elif bucket_internal and token_idx is not None:
```
yafshar, line 312:
```suggestion
        elif all([bucket_internal, token_idx]):
```

Suggestion to use all to make sure all variables are truthy (i.e., not None, not an empty string, not zero, etc.)

imangohari1, line 312: I've compared this to the llama/qwen2 (below) and I think this line should be 



```suggestion
        elif (reuse_cache or bucket_internal) and token_idx is not None:
```

https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/models/qwen2/modeling_qwen2.py#L890C14-L890C72
https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/models/llama/modeling_llama.py#L1117C9-L1117C73

imangohari1 (changes requested):
Looks good so far. 
I think there should be  a change. 
I tested this with `mpt-30b` as well below and seems to be running fine. 
```bash
python run_generation.py --model_name_or_path mosaicml/mpt-30b --use_hpu_graphs --use_kv_cache --max_input_tokens 128 --max_new_tokens 128 --trim_logits --bf16 --warmup 2 --n_iterations 2 --limit_hpu_graphs --batch_size 32 --bucket_size 128 --bucket_internal
```
Result: `715.6375115391292 tokens/second`

```bash
python run_generation.py --model_name_or_path mosaicml/mpt-30b --use_hpu_graphs --use_kv_cache --max_input_tokens 128 --max_new_tokens 128 --trim_logits --bf16 --warmup 2 --n_iterations 2 --limit_hpu_graphs --batch_size 32 --bucket_size 128
```
Result: `471.0239768753105 tokens/second`

optimum/habana/transformers/models/mpt/modeling_mpt.py
pk1d3v: @imangohari1, thanks for review!

There's no `reuse_cache` support for MPT yet. That's why I didn't add it to the condition.

imangohari1 (approved):
LGTM! 

@regisss 
Could you take a final look here?

optimum/habana/transformers/models/mpt/modeling_mpt.py
ssarkar2: historical context: reuse_cache came first, and then this change: https://github.com/huggingface/optimum-habana/pull/1028

PR 1028 removes the need for reuse_cache. So for new model optimizations I think it is fine to only make changes in line with PR1028, and leave out reuse_cache related changes.

Only in older models, where we already had reuse_cache code, we accomodate both

regisss (approved):
LGTM!

PR #1027, opened 2024-05-31
Upgrade to Transformers 4.40

regisss:
# What does this PR do?

As per title.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1138, opened 2024-07-17
gpt_bigcode: added FusedSDPA kernel

mgonchar:
Added support of following options to gpt_bigcode (starcoderbase) model use_flash_attention,
flash_attention_recompute,
flash_attention_fast_softmax,
flash_attention_causal_mask

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
+++ b/optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
@@ -1,12 +1,149 @@
+import math
 from typing import List, Optional, Tuple, Union
 
 import torch
+import torch.nn.functional as F
 import torch.utils.checkpoint
 from torch.nn import CrossEntropyLoss
 from transformers.modeling_outputs import BaseModelOutputWithPastAndCrossAttentions, CausalLMOutputWithCrossAttentions
 from transformers.models.gpt_bigcode.modeling_gpt_bigcode import GPTBigCodeForCausalLM
 
 from ...modeling_attn_mask_utils import GaudiAttentionMaskConverter
+
+
+try:
+    from habana_frameworks.torch.hpex.kernels import FusedSDPA
+except ImportError:
+    print("Not using HPU fused scaled dot-product attention kernel.")
+    FusedSDPA = None
+
+import habana_frameworks.torch.core as htcore
+
+
+def gaudi_flash_attn_v1(
+    query_layer, key_layer, value_layer, attention_mask, dropout_rate, is_causal, scale, softmax_mode, q_block_size
+):
+    """
+    Gaudi version of Flash Attention V1 to support long sequence at prompt phase
+    Causal mask is not supported in this optimization
+    """
+    if is_causal:
+        raise ValueError("Causal mask is not supported for long input sequences")
+
+    q_len = query_layer.size(-2)
+    q_tiles = (q_len // q_block_size) if (q_len % q_block_size == 0) else math.ceil(q_len / q_block_size)
+    q_padding = q_tiles * q_block_size - q_len
+    query_layer = F.pad(query_layer, (0, 0, 0, q_padding), "constant", 0)
+    if attention_mask is not None:
+        attention_mask = F.pad(attention_mask, (0, 0, 0, q_padding), "constant", -10000.0)
+    row_o_list = []
+    for i in range(q_tiles):
+        s, e = i * q_block_size, (i + 1) * q_block_size
+        row_q = query_layer[:, :, s:e, :]
+        row_mask = attention_mask[:, :, s:e, :]
+        attn_output_partial = FusedSDPA.apply(
+            row_q, key_layer, value_layer, row_mask, dropout_rate, is_causal, scale, softmax_mode
+        )
+        row_o_list.append(attn_output_partial)
+    attn_output = torch.cat(row_o_list, dim=-2)
+    if q_padding != 0:
+        attn_output = attn_output[:, :, :-q_padding, :]
+    return attn_output
+
+
+def apply_FusedSDPA(
+    self,
+    query,
+    key,
+    value,
+    attention_mask=None,
+    flash_attention_recompute=False,
+    flash_attention_fast_softmax=False,
+    flash_attention_causal_mask=False,
+):
+    """
+    Copied from GPTBigCodeSdpaAttention._attn: https://github.com/huggingface/transformers/blob/v4.40-release/src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
+    The only differences are:
+    - replaced torch.nn.functional.scaled_dot_product_attention with Habana's FusedSDPA
+    - removed WA for key and value tensor expanding over heads dimension. That WA also works but dramatically drops throughput
+    - added args use_flash_attention, flash_attention_recompute, flash_attention_fast_softmax, flash_attention_causal_mask to control parameters of FusedSDPA
+    - added special case handling for input larger 8192 with function gaudi_flash_attn_v1
+    """
+
+    scale = None
+    if not self.scale_attn_weights:
+        scale = 1
+
+    # MQA models: (batch_size, query_length, num_heads * head_dim)
+    # MHA models: (batch_size, num_heads, query_length, head_dim)
+    query_shape = query.shape
+    batch_size = query_shape[0]
+
+    if self.multi_query:
+        query_length = query_shape[1]
+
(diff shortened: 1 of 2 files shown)


mgonchar:
Original implementation
```
python run_generation.py \
    --model_name_or_path bigcode/starcoderbase-3b \
    --use_hpu_graphs \
    --use_kv_cache \
    --batch_size 42 \
    --max_new_tokens 1024 \
    --do_sample\
    --bf16
```

Stats
```
---------------------------------------------------------------------------------------------------------------
Throughput (including tokenization) = 3970.2326582228557 tokens/second
Number of HPU graphs                = 15
Memory allocated                    = 84.78 GB
Max memory allocated                = 94.62 GB
Graph compilation duration          = 64.8335093039932 seconds
---------------------------------------------------------------------------------------------------------------
```

FusedSDPA
```
python run_generation.py \
    --model_name_or_path bigcode/starcoderbase-3b \
    --use_hpu_graphs \
    --use_kv_cache \
    --batch_size 42 \
    --max_new_tokens 1024 \
    --do_sample\
    --bf16 \
    --use_flash_attention
```

Stats
```
--------------------------------------------------------------------------------------------------------------
Throughput (including tokenization) = 4175.830644970393 tokens/second
Number of HPU graphs                = 15
Memory allocated                    = 13.41 GB
Max memory allocated                = 94.45 GB
Graph compilation duration          = 75.58716204599841 seconds
--------------------------------------------------------------------------------------------------------------
```

examples/text-generation/utils.py
```diff
@@ -515,6 +515,12 @@ def initialize_model(args, logger):
         "token": args.token,
         "trust_remote_code": args.trust_remote_code,
     }
+
+    # For starcoderbase model it is essential to setup proper flags for sdpa kernel at model level
+    # to avoid transpose of tensors at each attention call
+    if 'starcoderbase' in args.model_name_or_path:
```
vidyasiv, line 521: Does this also apply to bigcode/starcoder? If so is it better to check the model type is gpt_bigcode instead?

optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
```diff
@@ -150,222 +291,245 @@ def gaudi_gpt_bigcode_block_forward(
...
-    use_cache: Optional[bool] = None,
-    output_attentions: Optional[bool] = None,
-    output_hidden_states: Optional[bool] = None,
-    return_dict: Optional[bool] = None,
-    token_idx: Optional[torch.Tensor] = None,
-) -> Union[Tuple, BaseModelOutputWithPastAndCrossAttentions]:
+class GaudiGPTBigCodeModel(GPTBigCodeModel):
     """
-    Copied from GPTBigCodeModel.forward: https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
+    Inherits from GPTBigCodeModel: https://github.com/huggingface/transformers/blob/v4.40-release/src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
```
vidyasiv, line 296: Assuming this is v4.40.2 ? Might be better to be specific in case the branch moves on after we commit this.

vidyasiv (commented):
@mgonchar , thank you for your PR.

- Please run make style prior to future submissions, it takes care of code formatting fixes
- Please verify if bigcode/starcoder which also uses this model file runs alright for you. I am seeing  an error `RuntimeError: Graph compile failed. synStatus=synStatus 26 [Generic failure]. `  for `python  run_generation.py --model_name_or_path bigcode/starcoder --batch_size 1 --use_hpu_graphs --use_kv_cache  --max_new_tokens 100 --bf16 --use_flash_attention`. I suspect this has to do with "starcoderbase" check and it works if substituted with "starcoder" but we need a better match because we want to avoid starcoder2.
- Testing: please check if the performance of starcoder in tests/test_examples.py is improved following your change. Or you may add a new entry for starcoderbase.

I will do a another pass after we resolve these

optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
mgonchar: Last commit with transformers package version bump for this model is to ver 4.40, see [this](https://github.com/huggingface/optimum-habana/pull/1027) PR. What do you mean by "specific" in case of this link?

examples/text-generation/utils.py
mgonchar: hm, right it might be better approach. But need to plug in to at least two functions: ```setup_model``` and ```setup_distributed_model```

optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
mgonchar: you may just post the link you want to see here and I will use it

examples/text-generation/utils.py
mgonchar: ok, there is one problem with this approach. Currently (see [here](https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/utils.py#L214)) we don't extract ```config``` ahead model creation. I might add new function, which will extract config just to check model type and modify args. It might be usefull for other models in future as well. If you are fine with such approach, I can do corresponding change

optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
vidyasiv: Sorry ignore my comment I see we have `"transformers >= 4.40.0, < 4.41.0"`, so using the v4.40 release branch is fine

examples/text-generation/utils.py
vidyasiv: I think that sounds better. I too ran into issues when I tried to find the right place to insert this, so a refactor might be in order. The main requirement was for it to apply to gpt_bigcode models in general, example: starcoderbase and starcoder - see my comment: https://github.com/huggingface/optimum-habana/pull/1138#pullrequestreview-2184281140

vidyasiv: Capturing some options we discussed offline:

Another way that I tested is to pass it during initialize_model() in examples/text-generation/utils.py like so
```
    model_kwargs = {
        "revision": args.model_revision,
        "token": args.token,
        "trust_remote_code": args.trust_remote_code,
        "use_flash_attention": args.use_flash_attention,
    }
```
But if you find you can override the classes so flash_attention is passed in early enough and recognized i.e from setup_generation_config() of examples/text-generation/utils.py then that is probably best.

mgonchar:
@vidyasiv I've updated this PR, based on your feedback. Please have a look

optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
```diff
@@ -9,6 +11,128 @@
...
+        attn_output = attn_output[:, :, :-q_padding, :]
+    return attn_output
+
+def apply_FusedSDPA(self, query, key, value, attention_mask=None, flash_attention_recompute=False, flash_attention_fast_softmax=False, flash_attention_causal_mask=False):
+
+    """
+    Copied from GPTBigCodeSdpaAttention._attn: https://github.com/huggingface/transformers/blob/v4.40-release/src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
+    The only differences are:
+    - replaced torch.nn.functional.scaled_dot_product_attention with Habana's FusedSDPA
+    - removed WA for key and value tensor expanding over heads dimention. That WA also works but dramatically drops throughput
```
vidyasiv, line 57:
```suggestion
    - removed WA for key and value tensor expanding over heads dimension. That WA also works but dramatically drops throughput
```

vidyasiv (approved):
Minor typo.I was able to run tests and so far LGTM

optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
mgonchar: thanks, fixed

vidyasiv (changes requested):
Please re-run `make style`

mgonchar:
@vidyasiv done

vidyasiv (approved):

vidyasiv:
@regisss , please take a look

mgonchar:
PR rebased. I've rechecked rebased code - no regressions found. @regisss please have a look

regisss (approved):
LGTM!

PR #1141, opened 2024-07-18
clear help msg for ignore_eos to avoid misunderstanding

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting


sywangyi:
@regisss please help review

examples/language-modeling/README.md
```diff
@@ -746,7 +746,7 @@ python3 ../text-generation/run_generation.py \
     --use_kv_cache \
     --batch_size 1 \
     --use_hpu_graphs \
-    --ignore_eos \
+    --no-ignore_eos \
```
ssarkar2, line 749: Why change `--ignore_eos` to `--no-ignore_eos` only for this cmd?

examples/text-generation/run_generation.py
```diff
@@ -267,7 +267,7 @@ def setup_parser(parser):
         "--ignore_eos",
         default=True,
         action=argparse.BooleanOptionalAction,
-        help="Whether to ignore eos, set False to disable it",
+        help="Whether to disable stopping with eos token when calling `generate`. --no-ignore_eos to disable it",
```
ssarkar2, line 270: **Minor**: For `argparse.BooleanOptionalAction`, if we have `--foo`, then the arg parser automatically provides `--no-foo`. That's mentioned in argparse documentation: https://docs.python.org/3/library/argparse.html ,  so we might not need to explicitly add `--no-ignore_eos to disable it` here.

I think its fine and good to add it, so we can keep it. But maybe we should add it in the help line of `examples/image-to-text/run_pipeline.py` as well. Or we should remove mention of `--no-ignore_eos` from both `examples/text-generation/run_generation.py` and `examples/image-to-text/run_pipeline.py`

ssarkar2 (commented):
Thanks for clarifying the help msg. Some minor comments

examples/language-modeling/README.md
sywangyi: The cmd is shown in Prompt/Prefix/P-tuning section, which uses https://huggingface.co/datasets/ought/raft/viewer/twitter_complaints for finetune, whose label is expected as "complaint" or "no complaint", as a result, the generation output is expected as "complaint" or "no complaint" using the finetuned peft model as input. if not set --no-ignore_eos, extra work will be output.

examples/text-generation/run_generation.py
sywangyi: The `ignore_eos` arg in `examples/text-generation/run_generation.py` does not use `argparse.BooleanOptionalAction`, it's `action="store_true",`, default value is false. and adding `--ignore_eos` in command makes it true. so no need to mention --no-ignore_eos in `examples/text-generation/run_generation.py`

libinta:
@sywangyi  can you fix the review comments?

ssarkar2 (approved):

regisss (approved):

PR #1159, opened 2024-07-25
use fast softmax only on prefill

jaygala223:
use fast softmax only on prefill

### DIFF
--- a/optimum/habana/transformers/models/llama/modeling_llama.py
+++ b/optimum/habana/transformers/models/llama/modeling_llama.py
@@ -473,7 +473,7 @@ class GaudiLlamaAttention(LlamaAttention):
                 use_recompute = True if os.getenv("QUANT_CONFIG", "") else False
                 with ht.sdp_kernel(enable_recompute=use_recompute):
                     attn_output = self.fused_scaled_dot_product_attention(
-                        query_states, key_states, value_states, attention_mask, 0.0, False, None, 'None'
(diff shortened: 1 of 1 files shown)


dvarshney-habana (approved):

yafshar:
@jaygala223 can you point to the source for this change? or is there an issue? From literature, fast softmax can be used in prefill

yafshar:
If there is an accuracy issue raised by using fast softmax, can you point out to the issue, or bring it up here for the reference. It could be beneficial for other models and prevent others to root causing in similar cases.

jaygala223:
@yafshar apologies for the delayed response. There was a performance regression which got introduced in a patch revert for context manager. This PR fixes it

yafshar:
> @yafshar apologies for the delayed response. There was a performance regression which got introduced in a patch revert for context manager. This PR fixes it

@jaygala223 thanks, does this affect other models? Should we consider the same change for other models?

yafshar (approved):
LGTM!

@regisss please look at this PR

yafshar:
@libinta please correct the PR label

jaygala223:
> @jaygala223 thanks, does this affect other models? Should we consider the same change for other models?

No, it does not affect other models

regisss (approved):

regisss:
Please run `make style`

regisss (approved):

PR #1148, opened 2024-07-22
Refactor KV cache, Rope  , reduce common code 

abhilash1910:
# What does this PR do?

Reduces common KVCache class code across generation models,  refactors into modelling class.
cc @libinta @vidyasiv 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1160, opened 2024-07-25
Set KV Cache update as static method

ulivne:
This solves torch compile issue related to using self argument in HQT patched module

### DIFF
--- a/optimum/habana/transformers/models/falcon/modeling_falcon.py
+++ b/optimum/habana/transformers/models/falcon/modeling_falcon.py
@@ -229,7 +229,8 @@ class KVCache(torch.nn.Module):
     def forward(self, cur, dim, idx):
         return self.update(self.cache, cur, dim, idx, self.inp_seq_len)
 
-    def update(self, prev, cur, dim, idx, inp_seq_len):
+    @staticmethod
+    def update(prev, cur, dim, idx, inp_seq_len):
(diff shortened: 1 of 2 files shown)


yafshar:
@ulivne can you point to the issue if it is filed. Or provide more details on how to reproduce the issue?

ulivne:
> @ulivne can you point to the issue if it is filed. Or provide more details on how to reproduce the issue?

The issue was internal to Habana SW stack during enablement of torch.compile usage.
The issue is that the HQT uses  different module for KVCache quantization, and torch.compile triggers graph break due to usage in different 'self' argument (log - "... mismatch for bound method.. ").
Setting the update method as static solved this, and also there is no need for this method to be  non-static as it not uses any class members.

yafshar:
@ulivne thanks. I pointed out your comment in this [PR 1148](https://github.com/huggingface/optimum-habana/pull/1148) 
Does it makes sense to close this PR and have everything there?

ulivne:
> @ulivne thanks. I pointed out your comment in this [PR 1148](https://github.com/huggingface/optimum-habana/pull/1148) Does it makes sense to close this PR and have everything there?

Yes it make sense to me.
i will close this PR once the PR 1148 is merged

regisss (approved):
LGTM!

PR #1165, opened 2024-07-28
Support loading hugging face checkpoint

ulivne:
* Support loading checkpoint with INC

* load_cp explanation

* Add torch_dtype bf16 for the model

* Support getting quantized weight tensor


MrGeva (approved):

examples/text-generation/utils.py
```diff
@@ -203,6 +203,14 @@ def setup_model(args, model_dtype, model_kwargs, logger):
             torch_dtype=model_dtype,
             **model_kwargs,
         )
+    elif args.load_cp:
+        from neural_compressor.torch.quantization import load
```
yafshar, line 207: @ulivne sounds like `neural_compressor` is missing from the `requirements` consider adding it!

ulivne: > @ulivne sounds like `neural_compressor` is missing from the `requirements` consider adding it!

neural_compressor is installed automatically as part of habana software stack. it replaces habana_quantization_toolkit which was also not part of requirements.

ulivne:
Removed test as it is uses a huggingface model, and we are not sure that we can use in our open source code (in terms of licence, intel regulations)

regisss (commented):
I think `load_cp` is not explicit enough. Transformers has `load_in_4bit` which I think would be better here.

@libinta You'll let me know when this PR can be merged.

ulivne:
> I think `load_cp` is not explicit enough. Transformers has `load_in_4bit` which I think would be better here.
> 
> @libinta You'll let me know when this PR can be merged.

this flag triggers neural compressor load api, which also supports loading of high precision models, and is planned to support loading of fp8 models in the future. it is correct that currently for gaudi only 4bit is suppoerted, however i think we should keep the load_cp name to avoid changing it in the future when we add loading of fp8 models.

regisss:
> > I think `load_cp` is not explicit enough. Transformers has `load_in_4bit` which I think would be better here.
> > @libinta You'll let me know when this PR can be merged.
> 
> this flag triggers neural compressor load api, which also supports loading of high precision models, and is planned to support loading of fp8 models in the future. it is correct that currently for gaudi only 4bit is suppoerted, however i think we should keep the load_cp name to avoid changing it in the future when we add loading of fp8 models.

`load_in_4_bit` is just a proposal but honestly, if I'm a user and I don't know the arguments, `load_cp` doesn't tell me anything about what it does. I would prefer something more explicit like `load_quantized_model` or even better automatically detect if this is a 4/8-bit checkpoint.

regisss (approved):

examples/text-generation/run_generation.py
```diff
@@ -300,6 +300,11 @@ def setup_parser(parser):
         help="Whether to trust the execution of code from datasets/models defined on the Hub. This option should only be set to `True` for repositories you trust and in which you have read the code, as it will execute code present on the Hub on your local machine.",
     )
     parser.add_argument(
+        "--load_quantized_model",
+        action="store_true",
+        help="Whether to load model from hugging face checkpoint.",
+    )
+
```
libinta, line 307: one line missing here

PR #1163, opened 2024-07-26
Upgrade to Transformers 4.43

regisss:
# What does this PR do?

As per title.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1167, opened 2024-07-29
[Clip] Update media device

murthy-l-habana:
Fixes media device name and fw params for media's external reader


regisss:
Closing as it was integrated into #1163.

PR #1168, opened 2024-07-29
[Stable-Diffusion] Update media device changes

murthy-l-habana:
Fixes media device handling for stable diffusion and move reader to external reader op


examples/stable-diffusion/training/media_pipe_imgdir.py
```diff
@@ -228,7 +217,7 @@ def __init__(
         drop_last=True,
         queue_depth=5,
     ):
-        self.device = get_device_name()
+        self.device = "legacy"
```
shepark, line 220: What is the definition of device "legacy" ?

PR #1172, opened 2024-07-30
simplify the code for `cache_idx` calculation

Jing1Ling:
# What does this PR do?

Actually I think it’s better to handle `cache_idx` in `prepare_inputs_for_generation()`. But considering many models already implement `--bucket_internal`, I just simplified the implementation and tried to make it easier to understand.

## Before submitting

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -1626,7 +1626,6 @@ class GaudiGenerationMixin(GenerationMixin):
         )
         hb_profer.start()
         bucket_size = model_kwargs.get("bucket_size", -1)
-        prev_idx = -1  # avoiding calculate cache_idx when its value is not changing
         bucket_internal = model_kwargs.get("bucket_internal", None)
         reduce_recompile = model_kwargs.get("reduce_recompile", False)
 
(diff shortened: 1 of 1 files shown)


emascarenhas:
Please sync your PR with main/upstream and fix any merge conflicts. Thank you.

ssarkar2 (changes requested):
1. Is this a fix, or just a code rewrite? if its a fix, can you please provide the failing command that is fixed with this PR
2. This change is in a critical path (generation/utils.py), used by many code paths. Could you please provide more details what this PR tries to do and how it does it, possible perf gains or any other relevant details
3. There is another section where we have same code, so if we update contrastive search we should update sample as well .. https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/generation/utils.py#L2341

Jing1Ling:
> 1. Is this a fix, or just a code rewrite? if its a fix, can you please provide the failing command that is fixed with this PR
> 2. This change is in a critical path (generation/utils.py), used by many code paths. Could you please provide more details what this PR tries to do and how it does it, possible perf gains or any other relevant details
> 3. There is another section where we have same code, so if we update contrastive search we should update sample as well .. https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/generation/utils.py#L2341

This is a small rewrite of the code, trying to simplify the code logic for `cache_idx` calculation. `cache_idx` is used for kv_cache slice, which means upper boundary index of current kv_cache bucket. It will only be used when `bucket_internal` is enabled. 
I spent unnecessary time understanding this code earlier, so I simplified it.

Jing1Ling:
And I updated _sample() and _contrastive_search().

PR #1178, opened 2024-08-01
Enable fast softmax as default

wszczurekhabana:
Upstream of: https://github.com/HabanaAI/optimum-habana-fork/pull/219

Original description:

This PR enables by default fast softmax feature of FusedSDPA.

To do that it changes handling of flash attention related arguments inside run_generation.py
Now the flags can also except values such as "--use_flash_attention true/false" etc.

Flag "--flash_attention_fast_softmax" will be true if "--use_flash_attention" is true. It can be explicitly turned off by passing "--flash_attention_fast_softmax false" to the script.

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -42,6 +42,22 @@ logger = logging.getLogger(__name__)
 
 
 def setup_parser(parser):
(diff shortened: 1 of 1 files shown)


wszczurekhabana:
@libinta could you add 1.17_dependency label to this PR?

PR #1185, opened 2024-08-02
Enable t.compile for Granite20B

dvarshney-habana:
**Enable t.compile for Granite20B**

Granite20B model uses the GPTBigCode model type. In this model, model object uses model.transformer instead of model.model for the actual model. This is a pointed fix for this model, in future other models may need similar change.

# What does this PR do?

Fixes # (issue)

## Before submitting


libinta (approved):

regisss (approved):
LGTM!

PR #1199, opened 2024-08-05
Refactor SD3

dsocek:
# SD3 Cleanup

- Removed duplicate code function in SD3 which can be inherited
- Added default scheduler in text-to-image sample (SD3 uses default `FlowMatchEulerDiscreteScheduler`)


dsocek:
@regisss this is the code cleanup PR as discussed in #1153.  I also forgot to add `default` option for scheduler selection which is needed for SD3 so I added it in this PR

regisss (approved):
LGTM!

PR #1200, opened 2024-08-05
sarkar/remove inner private api calls in tests

ssarkar2:
# What does this PR do?

There are a bunch of tests of this nature:
```
out1=model.generate(mode="sample")
out2=model._sample()
assert out1==out2
```

Reasons for removing these tests
1. `generate` is the only public api, the rest are meant to be private and are not meant to be called
2. The test themselves are not testing much, because it exercises the same paths, and the outer generate ends up calling the inner ones themselves
3. the transformer repo itself does not have these tests
4. the inner/private apis can change at any time (like it changed in this upgrade), causing multiple tests to break

This PR also hooks up the stub contrastive search in OH, so it will fail now. so contrastive search tests are marked xfail

Current status of the branch on 1 test file that I tested t5:
=
FAILED tests/transformers/tests/models/t5/test_modeling_t5.py::T5ModelIntegrationTests::test_summarization - AssertionError: Lists differ: ['pro[360 chars]n in the occupied Palestinian territory . as m[703 chars]2 .'] != ['pro[360 chars]n in Palestinian territories...
FAILED tests/transformers/tests/models/t5/test_modeling_t5.py::T5ModelIntegrationTests::test_torch_quant - AttributeError: 'Linear' object has no attribute 'custom_name'

Fixes # (issue)

## Before submitting

### DIFF
--- a/tests/transformers/tests/models/t5/test_modeling_t5.py
+++ b/tests/transformers/tests/models/t5/test_modeling_t5.py
@@ -1421,7 +1421,10 @@ class T5ModelIntegrationTests(unittest.TestCase):
         translation = tok.decode(output[0], skip_special_tokens=True, clean_up_tokenization_spaces=False)
         self.assertEqual(translation, expected_translation)
 
+    # contrastive search is not supported and expected to fail
+    # In earlier versions it was passing because it was going down default implementation, and it just happened to pass
     @slow
+    @pytest.mark.xfail(reason="contrastive search is not implemented", raises=NotImplementedError)
     def test_contrastive_search_t5(self):
         article = (
             " New York (CNN)When Liana Barrientos was 23 years old, she got married in Westchester County, New York. A"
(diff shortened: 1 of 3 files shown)


tests/transformers/tests/models/t5/test_modeling_t5.py
```diff
@@ -1421,7 +1421,10 @@ def test_translation_en_to_ro(self):
         translation = tok.decode(output[0], skip_special_tokens=True, clean_up_tokenization_spaces=False)
         self.assertEqual(translation, expected_translation)
 
+    # contrastive search is not supported and expected to fail
+    # In earlier versions it was passing because it was going down default implementation, and it just happened to pass
     @slow
+    @pytest.mark.xfail(reason="contrastive search is not implemented", raises=NotImplementedError)
```
ssarkar2, line 1427: contrastive search will be merged, so this will need to be removed

imangohari1:
another reason to reflect these changes: 
- the eventual test_utils.py would much closer in test content to HF upstream for future transformers updates.

ssarkar2:
merged to https://github.com/huggingface/optimum-habana/pull/1163 via https://github.com/huggingface/optimum-habana/pull/1204/files#diff-7726d5e68b02d4909ece94ee45f0cc0beeb9b94c53cbd4c5cbf75476e6f36a89

PR #1205, opened 2024-08-06
Fix max_position_embedding error for Mixtral

jiminha:
PR 1202 is breaking Mixtral since it doesn't pass config.
This change fixes  both for Mixtral and Llama.


regisss (commented):
@jiminha Do we still need this change? https://github.com/huggingface/optimum-habana/pull/1202/files

jiminha:
> @jiminha Do we still need this change? https://github.com/huggingface/optimum-habana/pull/1202/files

no we don't need. I thought you already merged, so I removed that code in this PR.

PR #1225, opened 2024-08-07
Remove the default env variable to trust remote code by default

yafshar:
# What does this PR do?

Remove the default env variable to trust remote code by default.

This is wrong to set the env variable by default in the code. 
```py
os.environ.setdefault("HF_DATASETS_TRUST_REMOTE_CODE", "true")
```

User should set this in their env if they trust remote dataset. User can do `export HF_DATASETS_TRUST_REMOTE_CODE=true` if they trust the remote dataset.

## Before submitting

### DIFF
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -573,6 +573,8 @@ deepspeed --num_gpus 8 run_lm_eval.py \
 -o eval.json
 ```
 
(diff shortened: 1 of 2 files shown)


yafshar:
@regisss can you please check this. This env variable is set wrongly and should be removed from the code

yafshar:
@libinta can you add a label for this minor fix. I removed the HF_DATASETS_TRUST_REMOTE_CODE env variable to be true for all datasets

regisss (approved):
Yes, I agree with you. I also pushed a commit to add a note in the README to document this.

PR #1228, opened 2024-08-07
work around llama torch.compile performance issue

libinta:
# What does this PR do?
In modeling_llama.py, the position_embedding check along with logger.warning_once causes the graph recompilation during decoding phase. As we are not using position_embeddings, remove the logic to check position_embeddings for now.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/llama/modeling_llama.py
+++ b/optimum/habana/transformers/models/llama/modeling_llama.py
@@ -537,16 +537,20 @@ class GaudiLlamaAttention(LlamaAttention):
                     else:
                         kv_seq_len = past_key_value[0].shape[-2]
 
-        if position_embeddings is None:
-            logger.warning_once(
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/llama/modeling_llama.py
```diff
@@ -537,16 +537,20 @@ def pre_attn_forward(
                     else:
                         kv_seq_len = past_key_value[0].shape[-2]
 
-        if position_embeddings is None:
```
deepvars, line 540: this code is coming from upstream https://github.com/huggingface/transformers/blob/v4.43.0/src/transformers/models/llama/modeling_llama.py 

Instead of commenting out this code which looks ugly, similar result should be achieved if you move this check inside the constructor. Can you pls try that?

deepvars (changes requested):

regisss (approved):

PR #1233, opened 2024-08-08
Add _reorder_cache back to Llama for HPU

jiminha:
Transformer 4.43 has moved to use new Cache mechanism(https://github.com/huggingface/transformers/pull/31898) , and removed _reorder_cache from llama code while HPU is still using the previous mechanism.  Adding back _reorder_cache to the HPU llama code. 

Without this PR, below test case is failing. 

 RUN_SLOW=true  GAUDI2_CI=1 python -m pytest tests/transformers/tests/models/llama/test_modeling_llama.py -v -s -k test_beam_search_generate

### DIFF
--- a/optimum/habana/transformers/models/llama/modeling_llama.py
+++ b/optimum/habana/transformers/models/llama/modeling_llama.py
@@ -1381,6 +1381,17 @@ class GaudiLlamaForCausalLM(LlamaForCausalLM):
         )
         return model_inputs
 
+    # Transformer4.43 use new Cache mechanism while Gaudi is not.
+    # Adding _reorder_cache back to support HPU.
+    @staticmethod
+    def _reorder_cache(past_key_values, beam_idx):
+        reordered_past = ()
+        for layer_past in past_key_values:
+            reordered_past += (
+                tuple(past_state.index_select(0, beam_idx.to(past_state.device)) for past_state in layer_past),
+            )
+        return reordered_past
+
 
 def apply_customized_rope(q, k, cos, sin, position_ids):
     if q.device.type == "hpu" and has_fused_rope:


imangohari1:
tested this changes. 
reverts the pytests to the know `1.16`+`transformer4.40` state 
```
GAUDI2_CI=1 RUN_SLOW=1 python -m pytest tests/transformers/tests/models/llama/test_modeling_llama.py -s -v
 .
.

======================================================================================================= short test summary info =======================================================================================================
FAILED tests/transformers/tests/models/llama/test_modeling_llama.py::LlamaModelTest::test_contrastive_generate_low_memory - TypeError: expected Tensor as element 0 in argument 0, but got list
FAILED tests/transformers/tests/models/llama/test_modeling_llama.py::LlamaModelTest::test_generate_from_inputs_embeds_decoder_only - RuntimeError: The input size 2, plus negative padding 0 and -12 resulted in a invalid output size, Check dimension 0 of your input.
======================================================================================== 2 failed, 58 passed, 26 skipped, 6 warnings in 30.07s ========================================================================================

yafshar:
@jiminha is this the common approach? Should we consider adding this to this PR https://github.com/huggingface/optimum-habana/pull/1148

jiminha:
> @jiminha is this the common approach? Should we consider adding this to this PR #1148

It seems #1148 is just moving KVCache to common place so we don't need to multiple define that in each code.  

This _reorder_cache needs to be part of  LlamaForCausalLM.   Original Transformer implements this method for each model separately. This call flow though only getting called with beamsearch Unittest and I don't think this is get invoked from run_generation examples.  If other models (which removed this _reorder_cache from TF4.43 upgrade) also need to support this test case, then we will also need to add this back to each model files.

imangohari1:
I've pulled this PR in this integration branch for more comprehensive testing. Will update when that test is done. 
https://github.com/huggingface/optimum-habana/compare/main...imangohari1:optimum-habana:ig/ci_20240808?expand=1

abhilash1910:
@yafshar @jiminha 
Upon a closer look, it does seem that this reordering logic is somewhat similar to a class of generative models - llama/opt/mpt/ qwen etc. Probably a case wise analysis needs to be made but a conditional control flow can ensure this can be reused across modules. Maybe after this gets merged.

yafshar:
> @yafshar @jiminha Upon a closer look, it does seem that this reordering logic is somewhat similar to a class of generative models - llama/opt/mpt/ qwen etc. Probably a case wise analysis needs to be made but a conditional control flow can ensure this can be reused across modules. Maybe after this gets merged.

I agree. When I looked at it I thought it can be useful for a class of models. That is why I asked @jiminha .

vidyasiv (approved):
LGTM, verified failing tests on main pass with this PR on Gaudi2:

- tests/transformers/tests/models/llama/test_modeling_llama.py::LlamaModelTest::test_beam_search_generate
- tests/transformers/tests/models/llama/test_modeling_llama.py::LlamaModelTest::test_beam_search_generate_dict_outputs_use_cache)

vidyasiv:
@regisss , please take a look

regisss (approved):

PR #1251, opened 2024-08-14
Fix FSDP graph error due to Tranformer 4.43 update

jiminha:
- Update CPU_RAM_EFFICIENT_LOADING to take it from config
- Disable FSDP flat_param upcast

# What does this PR do?
With transformer 4.43 update(Accelerator 0.33.0), FSDP functionality is broken. 
- Flag was not set correctly due to the argument update. This cause FSDP not fully enabled causing Llama70b test fail with Host OOM. 
- Upcast logic(https://github.com/huggingface/accelerate/pull/2674) has been added on the update, but this is causing FSDP graph compiler failure.

### DIFF
--- a/examples/language-modeling/fsdp_config.json
+++ b/examples/language-modeling/fsdp_config.json
@@ -9,5 +9,5 @@
     "fsdp_use_orig_params": true,
     "transformer_layer_cls_to_wrap": "GaudiLlamaDecoderLayer",
     "fsdp_activation_checkpointing": false,
-    "cpu_ram_efficient_loading": true
+    "fsdp_cpu_ram_efficient_loading": true
 }
--- a/examples/language-modeling/fsdp_config.json
+++ b/examples/language-modeling/fsdp_config.json
@@ -8,5 +8,6 @@
     "fsdp_sync_module_states": true,
     "fsdp_use_orig_params": true,
     "transformer_layer_cls_to_wrap": "GaudiLlamaDecoderLayer",
-    "fsdp_activation_checkpointing": false
+    "fsdp_activation_checkpointing": false,
+    "cpu_ram_efficient_loading": true
 }
--- a/examples/question-answering/fsdp_config.json
+++ b/examples/question-answering/fsdp_config.json
@@ -8,5 +8,6 @@
     "fsdp_sync_module_states": true,
     "fsdp_use_orig_params": true,
     "transformer_layer_cls_to_wrap": "BertLayer",
-    "fsdp_activation_checkpointing": false
+    "fsdp_activation_checkpointing": false,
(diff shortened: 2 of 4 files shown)


optimum/habana/accelerate/accelerator.py
```diff
@@ -485,6 +485,12 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                             ),
                             auto_wrap_policy=fsdp_plugin.auto_wrap_policy,
                         )
+
+                """
+                TODO: Temporarily disable this upcast due to FSDP graph compile issue.
```
vivekgoe, line 490: Please add link to PR where this code was added to "accelerate" repo. Will help to understand the context when issue is debugged in future.

examples/language-modeling/fsdp_config.json
```diff
@@ -8,5 +8,6 @@
     "fsdp_sync_module_states": true,
     "fsdp_use_orig_params": true,
     "transformer_layer_cls_to_wrap": "GaudiLlamaDecoderLayer",
-    "fsdp_activation_checkpointing": false
```
vivekgoe, line 11: Please also update peft version in requirement file (ignore if it is already done)

examples/question-answering/fsdp_config.json
```diff
@@ -8,5 +8,6 @@
     "fsdp_sync_module_states": true,
     "fsdp_use_orig_params": true,
     "transformer_layer_cls_to_wrap": "BertLayer",
-    "fsdp_activation_checkpointing": false
+    "fsdp_activation_checkpointing": false,
```
vivekgoe, line 11: Please also update peft version in requirement file (ignore if it is already done)

vivekgoe (changes requested):
Looks good to me except for couple of minor comments.

examples/language-modeling/fsdp_config.json
jiminha: For PEFT, the current version is actually 0.12.0 from the docker itself. 
Single pytest, and diffuser test install peft to 0.10.0 and 0.7.0 for their test, but default is 1.12.0 and also in language-modeling requirements also specify peft as 0.12.0 so I think we are ok.

examples/question-answering/fsdp_config.json
jiminha: same as above comment.

optimum/habana/accelerate/accelerator.py
jiminha: Updated .

imangohari1:
@yafshar @emascarenhas  FYI regarding fsdp.

regisss (approved):

PR #1249, opened 2024-08-13
Update README.md for trust_remote_code and fix model name

shepark:
Fixes README.md validation failures
 . added --trust_remote_code in contrastive-image-text test cases
. fixed BiomedCLIP-PubMedBERT_256-vit_base_patch16_224 name
. removed PEFT model example (goliaro/llama-2-7b-lora-full model is no longer available, will add back when we find available PEFT model)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1259, opened 2024-08-15
Update README and more changes, rebase to main

shepark:
Fixes README.md validation failures
. added --trust_remote_code in contrastive-image-text test cases
. fixed BiomedCLIP-PubMedBERT_256-vit_base_patch16_224 name
. removed PEFT model example (goliaro/llama-2-7b-lora-full model is no longer available, will add back when we find available PEFT model)
. and some more changes

## Before submitting


shepark:
@regisss Sorry for late, this is the rebased version of PR: https://github.com/huggingface/optimum-habana/pull/1249 to the main.

regisss:
There are still changes we don't want to merge into main here, like `setup.py`, `requirements.txt`, `version.py`.
I think it may be easier to start from a branch natively based on main.
As a general rule, we never directly merge anything into the release branches so everything should be done in main. And then I cherry-pick the necessary commits from main into the release branch if needed.

shepark:
> There are still changes we don't want to merge into main here, like `setup.py`, `requirements.txt`, `version.py`. I think it may be easier to start from a branch natively based on main. As a general rule, we never directly merge anything into the release branches so everything should be done in main. And then I cherry-pick the necessary commits from main into the release branch if needed.

Done, thanks.

regisss (approved):

PR #1260, opened 2024-08-15
Gpt bigcode fused sdpa

yeonsily:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -638,8 +638,8 @@ def initialize_model(args, logger):
     )
     if model.config.model_type == "gpt_bigcode" and args.torch_compile and args.use_flash_attention:
         logger.warning_once(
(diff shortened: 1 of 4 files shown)


examples/text-generation/utils.py
```diff
@@ -636,6 +636,12 @@ def initialize_model(args, logger):
         if not args.parallel_strategy == "tp"
         else setup_distributed_model_tp(args, model_dtype, model_kwargs, logger, cache_dir)
     )
+    if model.config.model_type == "gpt_bigcode" and args.torch_compile and args.use_flash_attention:
```
libinta, line 639: can we add TODO: since we need to remove it later

yeonsily: Removed this WA and updated README.

regisss (approved):

PR #1270, opened 2024-08-19
Idefics2

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting


sywangyi:
|  throught(tokens/s)| A100 | Gaudi2 |
|-------|------|----------|
| BF16 | 11.7 |  55.2|
|FP32 |15.3 | 29.8|

cmd:
python3 run_pipeline.py     --model_name_or_path HuggingFaceM4/idefics2-8b     --use_hpu_graphs     --bf16

optimum/habana/transformers/models/idefics2/modeling_idefics2.py
```diff
@@ -0,0 +1,340 @@
...
+                # Remove padding images from the mask/pP p
+                pixel_attention_mask = pixel_attention_mask.view(
+                    batch_size * num_images, *pixel_attention_mask.shape[2:]
+                )
+                pixel_attention_mask = pixel_attention_mask[real_images_inds].contiguous()
+
+            patch_size = self.config.vision_config.patch_size
+            patches_subgrid = pixel_attention_mask.cpu().unfold(dimension=1, size=patch_size, step=patch_size)
+            patches_subgrid = patches_subgrid.cpu().unfold(dimension=2, size=patch_size, step=patch_size)
+            patch_attention_mask = (patches_subgrid.sum(dim=(-1, -2)) > 0).bool()
```
ssarkar2, line 127: Hi, can you please let me know the values of patch_size,
and values/shape of pixel_attention_mask
Trying to see if we can rewrite it in some way so that it stays on HPU

ssarkar2 (commented):
Looks good otherwise, have a question regd unfold

optimum/habana/transformers/models/idefics2/modeling_idefics2.py
sywangyi: the patch_size is 14 for idefics2-8b. see https://huggingface.co/HuggingFaceM4/idefics2-8b/blob/main/config.json#L28
for the pixel_attention_mask. it's mask about pixel attention, shape not fixed, shape is (batch*image_num, image_height, image_width], image heigtht/width is rescale to the scale according to longest edge and shorted edge in config see https://huggingface.co/HuggingFaceM4/idefics2-8b/blob/main/preprocessor_config.json#L22-L25 and code https://github.com/huggingface/transformers/blob/main/src/transformers/models/idefics2/image_processing_idefics2.py#L48. image_num is 5 since image spit in on in config, see https://huggingface.co/HuggingFaceM4/idefics2-8b/blob/main/preprocessor_config.json#L3

ssarkar2: I think we can do this with a convolution, something like the following. Note I just tried a small example with some random numbers, and did not verify it on HPU ir did not verify on the idefics pipeline. Could you check if it provides any perf benefit (and hopefully doesn't cause any accuracy issues)


```

import torch

bs = 2
patch_size = 14
image_num =5
W=128
H=128

pixel_attention_mask = torch.rand([bs*image_num, W, H])

patches_subgrid = pixel_attention_mask.cpu().unfold(dimension=1, size=patch_size, step=patch_size)
patches_subgrid = patches_subgrid.cpu().unfold(dimension=2, size=patch_size, step=patch_size)
res1 = patches_subgrid.sum(dim=(-1, -2))


filter = torch.ones([1, 1, patch_size, patch_size])

res2 = torch.nn.functional.conv2d(pixel_attention_mask.unsqueeze(1), filter, stride=patch_size).squeeze()

print('max abs', torch.max(torch.abs(res1 - res2)))
```

sywangyi: res2 = torch.nn.functional.conv2d(pixel_attention_mask.unsqueeze(1), filter, stride=patch_size).squeeze()
NotImplementedError: convolution_overrideable not implemented. You are likely triggering this with tensor backend other than CPU/CUDA/MKLDNN, if this is intended, please use TORCH_LIBRARY_IMPL to override this function

ssarkar2: so my above impl works on CPU, and I meant it as a reference. didnt really try it in hpu

for hpu i suppose u have to create a conv layer (nn.module) in the init of the model, set it to weights=1 and then use that instead of the functional API for conv

If that seems too much for this PR, we can skip this, and we can try it out later for a separate PR, so I'll approve this one for now. But if u do run the expt mentione dabove (conv nnmodule in the init instead of torch.nn.functional.conv2d) and get some speed up, then please let us know here, and we can then incorporate it into this PR itself.

ssarkar2 (approved):

optimum/habana/transformers/models/idefics2/modeling_idefics2.py
sywangyi: yes, agree. we could skip this now. meanwhile, I add finetune example of image-to-text. which is used for idecfics finetune. please also help review it.
some issue find during enabling.
1) the image size from dataset varies. so the visual model input is varying. so I move the visual modeling handing to prepare input(only impact 1th token generation) 
2) if ignore_eos is False and batch_size>1. the eos is handled incorrectly now, which will cause accuracy in the evaluation low. I fix it in generation code.
3) I also add testcase for finetune case.

sywangyi:
> Looks good otherwise, have a question regd unfold

Hi, I have replace unfold with conv2D. cpu fallback is not needed. finetune got speedup @ssarkar2

sywangyi:
has found the recent PR which upgrade test_image_to_text_example.py. add "use_kv_cache", whose default value is false if not indicated explicitly in command line. which will lead image-to-text ci test  to fail. fix it.

sywangyi:
also found the similar PR which mean to fix eos issue if batch>1, please refer https://github.com/huggingface/optimum-habana/pull/1287,
 the root cause of it is the  needs_tensor_output is incorrect set

optimum/habana/transformers/generation/stopping_criteria.py
```diff
@@ -79,11 +80,7 @@ def gaudi_EosTokenCriteria_call(
 
 
 def needs_tensor_output(token_idx, ignore_eos, eos_token_id) -> bool:
-    if token_idx is None:
-        return not ignore_eos and eos_token_id is not None
-    else:
-        # token_idx is present, so we have static shapes, so using single boolean
-        return False
+    return not ignore_eos and eos_token_id is not None
```
ssarkar2, line 83: token_idx input is not used, could it be removed?

ssarkar2, line 83: token_idx is unused, could we remove it

sywangyi:
@regisss could you review and merge it?

regisss (approved):

PR #1271, opened 2024-08-19
Fix memory regression for modeling llama

libinta:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/llama/modeling_llama.py
+++ b/optimum/habana/transformers/models/llama/modeling_llama.py
@@ -170,10 +170,16 @@ class GaudiLlamaRotaryEmbedding(torch.nn.Module):
         if seq_len > self.max_seq_len_cached:
             self._set_cos_sin_cache(seq_len=seq_len, device=x.device, dtype=x.dtype)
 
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/llama/modeling_llama.py
```diff
@@ -170,10 +170,16 @@ def forward(self, x, seq_len=None):
         if seq_len > self.max_seq_len_cached:
             self._set_cos_sin_cache(seq_len=seq_len, device=x.device, dtype=x.dtype)
 
-        return (
-            self._cos_cached[:seq_len].to(dtype=x.dtype) * self.attention_scaling,
-            self._sin_cached[:seq_len].to(dtype=x.dtype) * self.attention_scaling,
-        )
+        if self.attention_scaling is None or self.attention_scaling == 1.0:
```
regisss, line 173:
```suggestion
        if self.attention_scaling == 1.0:
```
I don't think `self.attention_scaling` can be equal to None

regisss (approved):

PR #1274, opened 2024-08-19
Fix BERT FSDP test

vidyasiv:
# What does this PR do?

Fixes Test failure from transformers_4.43 update. :  `tests/test_fsdp_examples.py::test_fsdp_bf16[token0-bert-base-uncased-Habana/bert-base-uncased-3516.322-85.5503-question-answering-24-8-run_qa.py-full_shard]` 
Error:
```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/root/optimum-habana/examples/question-answering/run_qa.py", line 733, in <module>
[rank0]:     main()
[rank0]:   File "/root/optimum-habana/examples/question-answering/run_qa.py", line 679, in main
[rank0]:     train_result = trainer.train(resume_from_checkpoint=checkpoint)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/optimum/habana/transformers/trainer.py", line 553, in train
[rank0]:     return inner_training_loop(
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/optimum/habana/transformers/trainer.py", line 1052, in _inner_training_loop
[rank0]:     self._maybe_log_save_evaluate(tr_loss, _grad_norm, model, trial, epoch, ignore_keys_for_eval)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/optimum/habana/transformers/trainer.py", line 1269, in _maybe_log_save_evaluate
[rank0]:     self._save_checkpoint(model, trial, metrics=metrics)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/optimum/habana/transformers/trainer.py", line 1331, in _save_checkpoint
[rank0]:     self._save_optimizer_and_scheduler(output_dir)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/optimum/habana/transformers/trainer.py", line 1409, in _save_optimizer_and_scheduler
[rank0]:     save_fsdp_model(
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/accelerate/utils/fsdp_utils.py", line 63, in save_fsdp_model
[rank0]:     with FSDP.state_dict_type(
[rank0]:   File "/usr/lib/python3.10/contextlib.py", line 142, in _exit_
[rank0]:     next(self.gen)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torch/distributed/fsdp/fully_sharded_data_parallel.py", line 828, in state_dict_type
[...]

PR #1281, opened 2024-08-21
Fix BERT FSDP test

regisss:
# What does this PR do?

This PR fixes a bug where the FSDP wrapper of the model is lost at some point in Accelerate:
- Distributed wrappers (i.e. DDP, FSDP, DeepSpeed) are removed [here](https://github.com/huggingface/accelerate/blob/v0.33.0/src/accelerate/utils/other.py#L90),
- At some point after that operation is performed, the FSDP wrapper is lost, which makes the BERT FSDP test fail.

This PR fixes that issue by keeping a buffer variable to not lose track of the distributed model and making sure that it is set it back at the end.

## Before submitting

### DIFF
--- a/optimum/habana/accelerate/utils/other.py
+++ b/optimum/habana/accelerate/utils/other.py
@@ -9,6 +9,15 @@ from accelerate.utils.versions import is_torch_version
 
 
 def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive: bool = False):
+    """
+    Adapted from: https://github.com/huggingface/accelerate/blob/v0.33.0/src/accelerate/utils/other.py#L56
+
+    Changes:
+    - add a `distributed_model` variable to keep track of the distributed wrapper
+      and not lose it when setting it back at the end (for compiled models)
+
+    See https://github.com/huggingface/optimum-habana/pull/1281 for more information.
+    """
     options = (torch.nn.parallel.DistributedDataParallel, torch.nn.DataParallel)
 
     is_compiled = is_compiled_module(model)
--- a/optimum/habana/accelerate/utils/other.py
+++ b/optimum/habana/accelerate/utils/other.py
@@ -35,6 +35,8 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive
 
         options += (FSDP,)
 
+    # Keep track of the distributed wrapper
+    # TODO: to revisit as lines 44 to 71 are now useless
     distributed_model = model
     while isinstance(model, options):
         model = model.module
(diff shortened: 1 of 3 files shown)


regisss:
Test fails because I get a throughput of 3234.769 and it should be larger than 3516.322 though

vidyasiv:
Thanks @regisss , I see the test passing and performance is as follows:
```
***** train metrics *****
  epoch                       =        2.0
  max_memory_allocated (GB)   =       6.85
  memory_allocated (GB)       =        0.4
  total_flos                  =  4047423GF
  total_memory_available (GB) =      94.62
  train_loss                  =     1.3817
  train_runtime               = 0:01:02.51
  train_samples               =      88524
  train_samples_per_second    =   3602.047
  train_steps_per_second      =     18.803
***** eval metrics *****
  epoch                       =        2.0
  eval_exact_match            =    76.8874
  eval_f1                     =    85.3401
  eval_runtime                = 0:00:13.00
  eval_samples                =      10784
  eval_samples_per_second     =    829.419
  eval_steps_per_second       =     12.998
  max_memory_allocated (GB)   =       6.85
  memory_allocated (GB)       =       0.76
  total_memory_available (GB) =      94.62
```
I have closed my other PR: https://github.com/huggingface/optimum-habana/pull/1274

yafshar (approved):
LGTM!

thanks @regisss, very nice catch! I missed that the FSDP wrapper is lost!

optimum/habana/accelerate/utils/other.py
```diff
@@ -0,0 +1,75 @@
...
+        from deepspeed import DeepSpeedEngine
+
+        options += (DeepSpeedEngine,)
+
+    if is_torch_version(">=", FSDP_PYTORCH_VERSION) and is_torch_distributed_available():
+        from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
+
+        options += (FSDP,)
+
+    distributed_model = model
```
vivekgoe, line 38: Looks good solution to me for now. But we will have problem in future if some use-case needs functionality from L42 - L69, with this change we are effectively skipping those lines. Maybe good to add a comment warning about this.

vivekgoe (commented):
@regisss reviewed change, please check comments. Thanks.

yafshar:
@regisss do you know or can you comment why the original solution is not an issue on non-Gaudi devices?

regisss:
> @regisss do you know or can you comment why the original solution is not an issue on non-Gaudi devices?

It's still unclear to me why it works on GPU, I'll need to take a closer look at it.

vivekgoe (commented):
LGTM.

PR #1284, opened 2024-08-22
Boost SDXL speed with initialized schedule step reset

dsocek:
# Set SDXL scheduler starting timestep index

In current code, we are not setting the starting timestep for Euler type schedulers (e.g. `EulerDiscreteScheduler`, `GaudiEulerDiscreteScheduler`, etc).  Consequently, this value is `None` which causes slower timesteps reset process (e.g. in multiple batches scenarios for Euler schedulers).  When initial value is set to `0` then reset is performed as simple index switch (faster). (For example, this can be seen [here](https://github.com/huggingface/diffusers/blob/v0.30.0/src/diffusers/schedulers/scheduling_euler_discrete.py#L486-L491))

This PR fixes the aforementioned issue.  This gives notable speed boost in SDXL pipeline for both `default` and `euler_discrete` scheduler selections when processing multiple batches.

## Example

Ran the following command before and after the fix:
```bash
python text_to_image_generation.py \
    --model_name_or_path stabilityai/stable-diffusion-xl-base-1.0 \
    --prompts "Sailing ship painting by Van Gogh" \
    --num_images_per_prompt 16 \
    --batch_size 4 \
    --image_save_dir ./sdxl_images \
    --scheduler default \
    --use_habana \
    --gaudi_config Habana/stable-diffusion \
    --bf16
```

Output log **before** this PR's fix:
```bash
...
Speed metrics: {'generation_runtime': 269.7971, 'generation_samples_per_second': 0.233, 'generation_steps_per_second': 0.233}
```

Output log **after** this PR's fix:
```bash
Speed metrics: {'generation_runtime': 236.4827, 'generation_samples_per_second': 0.718, 'generation_steps_per_second': 0.718}
```

Performance is boosted from from **0.233 images/sec** to **0.718 images/sec** 

> Note: Rebased to latest main branch and ran `make style`

### DIFF
--- a/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
+++ b/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
@@ -651,6 +651,8 @@ class GaudiStableDiffusionXLPipeline(GaudiDiffusionPipeline, StableDiffusionXLPi
             t1 = t0
 
             self._num_timesteps = len(timesteps)
+            if hasattr(self.scheduler, "set_begin_index"):
+                self.scheduler.set_begin_index()
 
             hb_profiler = HabanaProfile(
                 warmup=profiling_warmup_steps,
@@ -687,8 +689,6 @@ class GaudiStableDiffusionXLPipeline(GaudiDiffusionPipeline, StableDiffusionXLPi
                 timestep_cond = self.get_guidance_scale_embedding(
                     guidance_scale_tensor, embedding_dim=self.unet.config.time_cond_proj_dim
                 ).to(device=device, dtype=latents.dtype)
-
-            self._num_timesteps = len(timesteps)
 
             # 8.3 Denoising loop
             throughput_warmup_steps = kwargs.get("throughput_warmup_steps", 3)
--- a/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_img2img.py
+++ b/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_img2img.py
@@ -536,6 +536,8 @@ class GaudiStableDiffusionXLImg2ImgPipeline(GaudiDiffusionPipeline, StableDiffus
                 ).to(device=device, dtype=latents.dtype)
 
             self._num_timesteps = len(timesteps)
+            if hasattr(self.scheduler, "set_begin_index"):
+                self.scheduler.set_begin_index()
 
             # 8.3 Denoising loop
             throughput_warmup_steps = kwargs.get("throughput_warmup_steps", 3)
--- a/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_inpaint.py
+++ b/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_inpaint.py
@@ -744,6 +744,8 @@ class GaudiStableDiffusionXLInpaintPipeline(GaudiDiffusionPipeline, StableDiffus
                 ).to(device=device, dtype=latents.dtype)
 
             self._num_timesteps = len(timesteps)
+            if hasattr(self.scheduler, "set_begin_index"):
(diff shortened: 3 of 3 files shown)


skavulya (approved):
Looks good to me. Thanks @dsocek. Impressive perf improvement.

imangohari1 (approved):
LGTM! 

I've ran the readme examples, `test_diffusers.py -k sdxl` and all passed. 

Also the timing checks out. 
running with bs=1, #prompt: 4, the expected is sample/s: 0.2, which checks with measured. 
```
python text_to_image_generation.py     --model_name_or_path stabilityai/stable-diffusion-xl-base-1.0     --prompts "Sailing ship painting by Van Gogh"     --num_images_per_prompt 4     --batch_size 1     --image_save_dir /tmp/stable_diffusion_xl_images     --scheduler euler_discrete     --use_habana     --use_hpu_graphs     --gaudi_config Habana/stable-diffusion     --bf16
```
```
████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [02:43<00:00, 40.88s/it]
[INFO|pipeline_stable_diffusion_xl.py:831] 2024-08-22 22:14:44,145 >> Speed metrics: {'generation_runtime': 163.531, 'generation_samples_per_second': 0.198, 'generation_steps_per_second': 0.793}
08/22/2024 22:14:55 - INFO - __main__ - Saving images in /tmp/stable_diffusion_xl_images...

dsocek:
@mandy-li @libinta @regisss If possible, could you also review this fix?  Thank you

skavulya:
@dsocek @regisss @libinta Now that the default diffuser schedulers are working well, should we consider deprecating the Gaudi-specific ones in future releases? This would ease upgrades and also streamline the process for enabling new models.

ssarkar2 (approved):

regisss (approved):

regisss:
@dsocek It seems there is a test consistently failing after this PR: https://github.com/huggingface/optimum-habana/actions/runs/10820064759/job/30019296923
Any idea why?

dsocek:
@regisss Looks like img2img SDXL pipeline test "test_stable_diffusion_xl_img2img_euler" checks for exactness of denoised output, assuming same randomness process.  When I added the reset step in the batch loop, it is called every `j`.  To avoid test failure, likely we can keep the current test values if we skip reset for `j=0`, or alternatively update expected vals array in the test.  I am out next week but back the following one, meanwhile @imangohari1 might be able to help to unblock.

dsocek:
> @dsocek CC @regisss Daniel, I added a `j != 0` in the `if` condition below and the test still failed. I even took this whole if out and no success. https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_img2img.py#L564 Is that were you thought of adding the skip for `j=0`?
> 
> I was able to get it passed by changing the expected values, but now the code is divergent from the upstream. We probably need to take a closer look at why this is happening. #1332

Hmm taking whole if out should have effectively reverted to a state before PR.  We can analyze more closely to root cause

imangohari1:
> > @dsocek CC @regisss Daniel, I added a `j != 0` in the `if` condition below and the test still failed. I even took this whole if out and no success. https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_img2img.py#L564 Is that were you thought of adding the skip for `j=0`?
> > I was able to get it passed by changing the expected values, but now the code is divergent from the upstream. We probably need to take a closer look at why this is happening. #1332
> 
> Hmm taking whole if out should have effectively reverted to a state before PR. We can analyze more closely to root cause

Taking the if outside the j loop, aka, https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_img2img.py#L539, reverts the PR to the previous state and it passes the test. 

Taking the reset if inside the loop, aka https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_img2img.py#L564, didn't help much.

PR #1285, opened 2024-08-22
Update FusedSDPA calling method as Gaudi documentation

yeonsily:
# What does this PR do?
According to the https://docs.habana.ai/en/latest/PyTorch/Reference/Python_Packages.html#hpex-kernels-fusedsdpa , 
we should pass 'recompute_mode' as API parameters.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
+++ b/optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
@@ -21,7 +21,16 @@ import habana_frameworks.torch.core as htcore
 
 
 def gaudi_flash_attn_v1(
-    query_layer, key_layer, value_layer, attention_mask, dropout_rate, is_causal, scale, softmax_mode, q_block_size
+    query_layer,
+    key_layer,
+    value_layer,
+    attention_mask,
+    dropout_rate,
+    is_causal,
+    scale,
+    softmax_mode,
+    enable_recompute,
+    q_block_size,
 ):
     """
     Gaudi version of Flash Attention V1 to support long sequence at prompt phase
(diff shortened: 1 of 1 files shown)


yafshar:
@yeonsily can you add more info on the issue? according to the document if you do not pass recompute_mode in the interface it reads it from the context API variable and it is given here.

libinta (approved):

regisss:
> @yeonsily can you add more info on the issue? according to the document if you do not pass recompute_mode in the interface it reads it from the context API variable and it is given here.

I agree, there is
```py
with ht.sdp_kernel(enable_recompute=enable_recompute):
```
currently in the codebase, so the behavior should be the same no?

yeonsily:
> > @yeonsily can you add more info on the issue? according to the document if you do not pass recompute_mode in the interface it reads it from the context API variable and it is given here.
> 
> I agree, there is
> 
> ```python
> with ht.sdp_kernel(enable_recompute=enable_recompute):
> ```
> 
> currently in the codebase, so the behavior should be the same according to the documentation no?

@yafshar  @regisss  We have an internal ticket which the current method causes performance degradation in torch.compile run as there will be graph breaks. This issue is torch.compile specific but I think we can pass recompute_mode as a parameter regardless of mode to keep the topology code same.
Please let me know if you want to have this change only for torch.compile mode.

regisss:
> > > @yeonsily can you add more info on the issue? according to the document if you do not pass recompute_mode in the interface it reads it from the context API variable and it is given here.
> > 
> > 
> > I agree, there is
> > ```python
> > with ht.sdp_kernel(enable_recompute=enable_recompute):
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > currently in the codebase, so the behavior should be the same according to the documentation no?
> 
> @yafshar @regisss We have an internal ticket which the current method causes performance degradation in torch.compile run as there will be graph breaks. This issue is torch.compile specific but I think we can pass recompute_mode as a parameter regardless of mode to keep the topology code same. Please let me know if you want to have this change only for torch.compile mode.

Thanks for the exaplanation, let's merge it then :slightly_smiling_face:

regisss (approved):

PR #1313, opened 2024-09-05
use native checkpointing under compile mode

xinyu-intel:
Change-Id: I49d2dafefd4d34f8c0165855611720fd7beecd4d

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -682,7 +682,7 @@ class GaudiTrainer(Trainer):
 
             import transformers.modeling_utils
 
-            if args.deepspeed:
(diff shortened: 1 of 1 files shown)


regisss:
Does it lead to an error without this fix? If yes, can you share a command to reproduce it?

bsochack:
> Does it lead to an error without this fix? If yes, can you share a command to reproduce it?

There is no error. This is performance issue with torch.compile. DeepSpeed activation checkpointing is not traced by torch.compile and everything fallbacks to eager. PyTorch took care of supporting their activation checkpointing, so it is traced and produces correct deterministic results.

regisss (approved):

PR #1322, opened 2024-09-10
unify_measurements.py script support to unify PCQ 70B 8x

Yantom1:
Change-Id: Ia23b3049bd54dd18c8ae779e027901cf0cd1de8b

This PR supports a case we have measurements in PCQ mode and we have multi weights and outputs values

### DIFF
--- a/examples/text-generation/quantization_tools/unify_measurements.py
+++ b/examples/text-generation/quantization_tools/unify_measurements.py
@@ -79,9 +79,24 @@ def unify_measurements(
                 for i in range(0, len(max_inputs)):
                     max_inputs[i] = max(measurement_json[node_name]["inputs"][i], max_inputs[i])
                 if max_outputs is not None:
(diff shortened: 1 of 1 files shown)


regisss:
Please add a description to this PR to explain a bit its purpose as the title is not very explicit.

Yantom1:
> Please add a description to this PR to explain a bit its purpose as the title is not very explicit.

done

regisss:
Is this compatible with Synapse 1.17 ?
And sorry for my ignorance, what's PCQ mode?

Yantom1:
> Is this compatible with Synapse 1.17 ? And sorry for my ignorance, what's PCQ mode?

it's not. PCQ is per channel quantization that means that each channel have its own quantization info.

regisss (approved):
LGTM!
Let's wait for the release of SynapseAI 1.18 to merge it.

PR #1335, opened 2024-09-16
Fix the deepspeed init_inference reserves device memory as model size

yeonsily:
For text-generation, set 'keep_module_on_host' to True to save device memory.

For quantization run,  https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html#enabling-and-running-inc-in-pytorch-models, says "If DeepSpeed is used, INC should be called after deepspeed.init_inference. "
But when deepspeed.init_inference() is called, it reserves the device memory as much as bf16 model size because it's initialized as bf16 model dtype first. This reserves extra device memory.

This change is fix for that.

# What does this PR do?

Fixes # (issue)

## Before submitting


regisss:
@yeonsily @libinta Should this be in the v1.14 release?

yeonsily:
> @yeonsily @libinta Should this be in the v1.14 release?

No. the change from deepspeed side is not ready for 1.18. Let me close this PR.

PR #1358, opened 2024-09-25
Fix GPT_neox incorrect output with batch query

Jianhong-Zhang:


vivekgoe:
@dvarshney-habana @libinta can you please help with review? change is in text-generation, you may have a better idea about effect of changes here.

optimum/habana/transformers/generation/utils.py
```diff
@@ -2354,6 +2361,12 @@ def _sample(
 
             if token_idx is not None:
                 idx = token_idx + model_kwargs.get("inputs_embeds_offset", 0)
+                if batch_size > 1 and has_eos_stopping_criteria:
+                    for i in range(batch_size):
+                        if next_tokens[i] == eos_token_id:
+                            idx_bs[i] = token_idx.item()
+                        if token_idx > idx_bs[i]:
+                            next_tokens[i] = pad_token_id
```
libinta, line 2369: please do not add any logic inside the generation loop, it will impact performance. instead you can do it after generation  to return right content based on eos

Jianhong-Zhang: moved logic after generation loop and before return

optimum/habana/transformers/generation/utils.py
```diff
@@ -2437,6 +2437,17 @@ def _sample(
...
+        if batch_size > 1 and has_eos_stopping_criteria:
+            eos_token_id = generation_config.eos_token_id
+            idx_bs = generation_config.max_length
+            for i in range(batch_size):
+                for idx in range(len(input_ids[i])):
+                    if input_ids[i][idx] == eos_token_id:
+                        idx_bs = idx
+                    if idx > idx_bs:
+                        input_ids[i][idx] = pad_token_id
+                idx_bs = generation_config.max_length
```
libinta, line 2449: do we still want to return padding token in this case or just stop here?

Jianhong-Zhang: according to customer requirement, we need to padding token after <|endoftext|>

regisss (approved):

PR #1373, opened 2024-09-27
Minor UX enhancement

MohitIntel:
# What does this PR do?
As part of UX enhancement drive,  got feedback to make this flag completely optional and not needing explicit  True/False when used.

## Before submitting


regisss (approved):
LGTM!

PR #1374, opened 2024-09-28
Remove additional float/clone() for perf 

jiminha:
# What does this PR do?

Remove float() from 4.45 upgrade due to perf issue
Extra aten::item, cast is causing perf degradation

DONOTMERGE NOW : Need accuracy test to see if there is any accuracy drop.

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -705,7 +705,6 @@ class GaudiGenerationMixin(GenerationMixin):
         Changes:
         - change the default from DynamicCache to tuples
         """
-
         cache_name = "past_key_values" if "mamba" not in self.__class__.__name__.lower() else "cache_params"
         requires_cross_attention_cache = (
             self.config.is_encoder_decoder or model_kwargs.get("encoder_outputs") is not None
@@ -2814,7 +2813,7 @@ class GaudiGenerationMixin(GenerationMixin):
                 else:
                     next_token_logits = torch.index_select(outputs.logits, -2, token_idx - 1).squeeze(-2)
             else:
-                next_token_logits = outputs.logits[:, -1, :].clone()
+                next_token_logits = outputs.logits[:, -1, :]
 
             next_token_scores = torch.nn.functional.log_softmax(
                 next_token_logits, dim=-1
@@ -3260,7 +3259,7 @@ class GaudiGenerationMixin(GenerationMixin):
                 else:
                     next_token_logits = torch.index_select(outputs.logits, -2, token_idx - 1).squeeze(-2)
             else:
-                next_token_logits = outputs.logits[:, -1, :].clone()
+                next_token_logits = outputs.logits[:, -1, :]
 
             next_token_scores = torch.nn.functional.log_softmax(
                 next_token_logits, dim=-1
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -1801,7 +1801,7 @@ class GaudiGenerationMixin(GenerationMixin):
                         logit_for_next_step = torch.index_select(outputs.logits, -2, token_idx - 1).squeeze(-2)
                 else:
                     # .float() is needed to retain precision for later logits manipulations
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/generation/utils.py
```diff
@@ -3580,8 +3579,7 @@ def _assisted_decoding(
 
             # 2.3. Process the new logits
             # .float() is needed to retain precision for later logits manipulations
```
mgonchar, line 3581: well, I think you should at least remove this comment

regisss:
@jiminha Have you got time to check if we get the same output sequences removing all these `.float()`?

jiminha:
> @jiminha Have you got time to check if we get the same output sequences removing all these `.float()`?

I tested all text_generation and test_encoder_decoder test cases from pytest and compared the outputs and it all looked the same. We'd like to understand more though why the original transformer added this float() for all logits' last output computation, what specific test cases that they ran and found this float() is needed.  Would you be able to check with them?

regisss:
@jiminha It's explained in the first message of this issue: https://github.com/huggingface/transformers/issues/30860

Basically, we don't want to cast all the logits to float in the forward causal-lm modelsand we only do it:
- if labels are provided to the forward
- and in `generate` for the last logit which is needed for computing the next token

So we should probably remove this float [here](https://github.com/huggingface/optimum-habana/blob/b73e250eea1c69fafc23dfca2e1492deab04fd25/optimum/habana/transformers/models/llama/modeling_llama.py#L1313) and keep the ones in `generate`. It's not done in this version of Transformers because it's a breaking change (the model outputs won't have the same type anymore) but they'll do it in v4.46.

jiminha:
> @jiminha It's explained in the first message of this issue: [huggingface/transformers#30860](https://github.com/huggingface/transformers/issues/30860)
> 
> Basically, we don't want to cast all the logits to float in the forward causal-lm modelsand we only do it:
> 
> * if labels are provided to the forward
> * and in `generate` for the last logit which is needed for computing the next token
> 
> So we should probably remove this float [here](https://github.com/huggingface/optimum-habana/blob/b73e250eea1c69fafc23dfca2e1492deab04fd25/optimum/habana/transformers/models/llama/modeling_llama.py#L1313) and keep the ones in `generate`. It's not done in this version of Transformers because it's a breaking change (the model outputs won't have the same type anymore) but they'll do it in v4.46.

Thanks for the link and explanation. So are we actually having double float cast on this v4.45 in the code since we added float() on the generation/utils.py and also kept the same logic in the model(causal LM) file as well.  The perf regression what I saw was in this t5 model (t5 test_encoder_decoder test) which doesn't use the causalLM and doesn't have float conversion to begin with, so the duration increased with this change. 

https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5/modeling_t5.py#L1768C8-L1768C50

optimum/habana/transformers/generation/utils.py
```diff
@@ -2370,7 +2369,7 @@ def _sample(
                     next_token_scores = logits_processor(input_ids, next_token_logits)
             else:
                 # .float() is needed to retain precision for later logits manipulations
-                next_token_logits = outputs.logits[:, -1, :].float()
```
libinta, line 2373: consider keeping this

optimum/habana/transformers/generation/utils.py
```diff
@@ -2814,7 +2813,7 @@ def expand_if_needed(tensor, new_size, value, dim=-1):
                 else:
                     next_token_logits = torch.index_select(outputs.logits, -2, token_idx - 1).squeeze(-2)
             else:
-                next_token_logits = outputs.logits[:, -1, :].float()
```
libinta, line 2817: we don't normally run to here

regisss:
For T5 it would be in `T5ForConditionalGeneration` but it's true that I don't see any cast to float in the forward.
Let me try with a check on the model type in `generate`.

jiminha:
Closing this PR since it's not adding much perf improvement for most of the model. T5 has small regression, but it's very little.

PR #1384, opened 2024-09-30
use model.generation_config instead of model.config

hsubramony:
running   test_run_speech_recognition_seq2seq_whisper-small_multi_card produces the following error

[rank0]: Traceback (most recent call last):
[rank0]:   File "/root/github/tmp/optimum-habana/examples/speech-recognition/run_speech_recognition_seq2seq.py", line 664, in <module>
[rank0]:     main()
[rank0]:   File "/root/github/tmp/optimum-habana/examples/speech-recognition/run_speech_recognition_seq2seq.py", line 583, in main
[rank0]:     model.config.save_pretrained(training_args.output_dir)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/transformers/configuration_utils.py", line 383, in save_pretrained
[rank0]:     raise ValueError(
[rank0]: ValueError: Some non-default generation parameters are set in the model config. These should go into either a) `model.generation_config` (as opposed to `model.config`); OR b) a GenerationConfig file (https://huggingface.co/docs/transformers/generation_strategies#save-a-custom-decoding-strategy-with-your-model)


vidyasiv (approved):

regisss:
This is an issue in Transformers and I guess it will be patched soon.
In the meantime, I just pushed a commit so that the line triggering the error is commented. `model.generation_config` will be saved when the model is saved anyway so I don't think we need to call it.

regisss (approved):

PR #1400, opened 2024-10-06
Fixed recursion error when uses both wrapped PEFT and DeepSpped

kplau1128:
- Recursion error in compute_loss. The crux of the issue lies in circular references introduced by the combined wrappers.


optimum/habana/sentence_transformers/st_gaudi_trainer.py
```diff
@@ -262,11 +262,17 @@ def call_model_init(self, trial=None) -> "SentenceTransformer":
...
+            # Check if the current model is already the wrapped model to prevet redundant assignments
+            if loss.model != model:
                 loss.model = model
-            elif isinstance(child, torch.nn.Module):
-                setattr(loss, name, self.override_model_in_loss(child, model))
+        # Recursively traverse child modules without replacing already wrapped models
+        for name, child in loss.named_children():
+            if isinstance(child, torch.nn.Module):
+                # Avoid replacing the model again if it's already the desired model
+                if not (name == "model" and child is model):
```
yafshar, line 274:
```suggestion
                if name != "model" or child is not model:
```

can we simplify the condition! (Personal suggestion)

optimum/habana/sentence_transformers/st_gaudi_trainer.py
```diff
@@ -317,13 +323,14 @@ def compute_loss(
         # Insert the wrapped (e.g. distributed or compiled) model into the loss function,
         # if the loss stores the model. Only called once per process
         # from https://github.com/UKPLab/sentence-transformers/blob/v3.1.0/sentence_transformers/trainer.py#L337
+        # Override the model in the loss function with the original model to prevent recursion
         if (
             model == self.model_wrapped
             and model != self.model  # Only if the model is wrapped
             and hasattr(loss_fn, "model")  # Only if the loss stores the model
-            and loss_fn.model != model  # Only if the wrapped model is not already stored
+            and loss_fn.model != self.model  # Assign the original model instead
```
yafshar, line 331: @kplau1128 this does not sound correct to me! Here the goal is to insert the wrapped model (distributed or compiled) into the loss function, not the original model

yafshar: I see that without this fix the recursion happen, but the workaround is different form the intention

yafshar: Here it does not do anything since `loss_fn.model is self.model`

kplau1128: You are right.

The crux of the issue lies in circular references introduced by the combined wrappers. I will rework the workaround only to modify the `override_model_in_loss` function to skip override loss when combined wrappers.

kplau1128:
Re-work the workaround to check PEFT-wrapped model and skip override model in loss.
This will prevent recursion error.

kplau1128:
Reworked as @nngokhale recommended,  just add a check condition in `compute_loss` for override `loss_fn`.

yafshar:
@kplau1128 this does not sound correct to me. Let me check it more, I will update here

yafshar:
@kplau1128 I made another PR https://github.com/huggingface/optimum-habana/pull/1428 which I thin root causes this issue. Please take a look and close this PR if you agree with that solution.

kplau1128:
> @kplau1128 I made another PR #1428 which I think root causes this issue. Please take a look and close this PR if you agree with that solution.

Thanks @yafshar. Looks similar to my previous rework tried to do https://github.com/huggingface/optimum-habana/commit/92ce0c9c7ecff65f37a8bad66a6d0577e025638f, but your PR is better. I will close this one.

PR #1411, opened 2024-10-10
Tests for text gen output text

vidyasiv:
# What does this PR do?

Fixes # (issue)
- Fix to `run_generation.py` so `results.json` stores all outputs instead of just the last when batch_size > 1
- Functional verification of select text gen models to be run for every PR (fast tests)
- Duration of added fast tests on Gaudi2: ` 7 passed, 1 warning in 1991.08s (0:33:11)`
- Functional verification added to slow test `test_text_generation_example.py` for select BF16 1HPU cases
- Found issues in bigcode/starcoder output in fast(passes) vs slow(fails due to bucketing options), will file ticket to fix that.

## Before submitting

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ style: clean
 fast_tests:
 	python -m pip install .[tests]
 	python -m pytest tests/test_gaudi_configuration.py tests/test_trainer_distributed.py tests/test_trainer.py tests/test_trainer_seq2seq.py
+	python -m pytest test_functional_text_generation_example.py
 
 # Run unit and integration tests related to Diffusers
 fast_tests_diffusers:
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,8 @@ style: clean
 fast_tests:
 	python -m pip install .[tests]
 	python -m pytest tests/test_gaudi_configuration.py tests/test_trainer_distributed.py tests/test_trainer.py tests/test_trainer_seq2seq.py
-	python -m pytest test_functional_text_generation_example.py
+# TODO enable when CI has more servers
+#	python -m pytest test_functional_text_generation_example.py
 
 # Run unit and integration tests related to Diffusers
 fast_tests_diffusers:
--- a/tests/test_functional_text_generation_example.py
+++ b/tests/test_functional_text_generation_example.py
@@ -11,16 +11,34 @@ from optimum.habana.utils import set_seed
 
 
 MODELS_TO_TEST = [
-    # ("bigcode/starcoder",""),
-    # ("bigcode/starcoder2-15b", ""),
-    # ("google/gemma-2b", ""),
+    (
+        "bigcode/starcoder",
+        "DeepSpeed is a machine learning framework for fast and accurate deep learning.\n\nDeepSpeed is a deep learning optimization library that makes distributed training easy, efficient, and effective.\n\nDeepSpeed is the engine that powers [Megatron-LM](https://github.com/NVIDIA/Megatron-LM) and [DeepSpeedExamples](https://github.com/microsoft/DeepSpeedExamples).\n\nDeepSpeed is built on PyTorch and TorchScript, and can run models on CPUs, GPUs",
+    ),
+    (
+        "bigcode/starcoder2-15b",
+        "DeepSpeed is a machine learning framework that optimizes transformer models using model parallelism and low-precision arithmetic. It is built to be easy to use and efficient to run.\n\nDeepSpeed is a research framework that provides both researchers and engineers with the ability to explore new ideas in distributed training.\n\nDeepSpeed is a community-driven project. We welcome your contributions and feedback.\n\n## Installation\n\nDeepSpeed can be installed from PyPI:\n\n```\npip install deepspeed\n```\n\n",
+    ),
+    (
+        "google/gemma-2b",
+        "DeepSpeed is a machine learning framework that uses a novel approach to accelerate training of deep neural networks. It is based on a novel technique called <em>gradient-based quantization</em>, which is a technique that quantizes the weights and activations of a neural network in a way that preserves the gradient flow. This allows for faster training and better accuracy.\n\nDeepSpeed is a framework that uses a novel approach to accelerate training of deep neural networks. It is based on a novel technique called <em>gradient-based quantization</em>, which is a",
+    ),
     (
         "meta-llama/Llama-2-7b-hf",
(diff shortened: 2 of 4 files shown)


vidyasiv:
@regisss , these tests add 30min to fast tests, let me know if you'd like me to cut down on the models

jiminha:
@vidyasiv  this is good, but is it possible we actually add this accuracy check also as part of the test_text_generation_example.py? We already run so many models test there for token generation, so I think it will not add much time to just compare the 1st or last token outputs along with checking throughputs.

vidyasiv:
> @vidyasiv this is good, but is it possible we actually add this accuracy check also as part of the test_text_generation_example.py? We already run so many models test there for token generation, so I think it will not add much time to just compare the 1st or last token outputs along with checking throughputs.

It won't be part of every PR check then i.e fast tests.

jiminha:
> > @vidyasiv this is good, but is it possible we actually add this accuracy check also as part of the test_text_generation_example.py? We already run so many models test there for token generation, so I think it will not add much time to just compare the 1st or last token outputs along with checking throughputs.
> 
> It won't be part of every PR check then i.e fast tests.

Got it. Then, can we keep this as it is for fast test, but also add similar logic for the existing text-gen test?

tests/test_text_generation_example.py
```diff
@@ -19,32 +21,38 @@
...
+            ("bigcode/starcoder2-3b", 1, False, 261.07213776344133, True),
+            ("adept/persimmon-8b-base", 4, False, 366.73968820698406, False),
+            ("Qwen/Qwen1.5-7B", 4, False, 490.8621617893209, False),
+            ("google/gemma-7b", 1, False, 109.70751574382221, True),
+            ("state-spaces/mamba-130m-hf", 1536, False, 5385.511100161605, False),
+            ("Deci/DeciLM-7B", 1, False, 120, False),
+            ("Qwen/Qwen2-7B", 512, False, 9669.45787, True),
+            ("Qwen/Qwen1.5-MoE-A2.7B", 1, True, 44.25834541569395, False),
+            ("EleutherAI/gpt-neo-2.7B", 1, False, 257.2476416844122, False),
         ],
```
jiminha, line 56: So only the model with "True" will have output tested? If we want to add more output check later on, all we need to do is change this last value to True, and add the values 1st token output to the "MODEL_OUTPUTS" table below?

tests/test_text_generation_example.py
```diff
@@ -89,41 +97,51 @@
             ("gpt2-xl", 1, False, 51.61471298016438),
         ],
     }
+    MODEL_OUTPUTS = {
```
jiminha, line 100: Can this be also used for fp8 test?

tests/test_text_generation_example.py
vidyasiv: yes that is right

tests/test_text_generation_example.py
vidyasiv: i have not tried but perhaps it can be a later effort

jiminha: Yeah I think we should add it for fp8 output check as well - with fp8, and int4. Maybe for another PR.

jiminha:
@vidyasiv  Is this time including model downloading time? 
Duration of added fast tests on Gaudi2:  7 passed, 1 warning in 1991.08s (0:33:11)

vidyasiv:
@jiminha , i deleted cache and kicked off a run, will let u know.

tests/test_functional_text_generation_example.py
```diff
@@ -0,0 +1,75 @@
...
+        "bigcode/starcoder2-3b": 'def print_hello_world():\n    print("Hello World")\n\ndef print_hello_world_with_name(name):\n    print("Hello World, " + name)\n\ndef print_hello_world_with_name_and_age(name, age):\n    print("Hello World, " + name + ", " + str(age))\n\ndef print_hello_world_with_name_and_age_and_gender(name, age, gender):\n    print("Hello',
+        "google/gemma-7b": "DeepSpeed is a machine learning framework that enables training of large-scale models on commodity hardware. It is designed to be a drop-in replacement for PyTorch, and it is compatible with the existing PyTorch ecosystem. DeepSpeed is designed to be easy to use, and it provides a number of features that make it easy to train large-scale models.\n\nDeepSpeed is a machine learning framework that enables training of large-scale models on commodity hardware. It is designed to be a drop-in replacement for PyTorch, and",
```
vidyasiv, line 17: @regisss , a common theme I observe in the output (except starcoder/starcoder2) is repeated sentences, not sure if that is some other bug or expected

vidyasiv:
@jiminha  can u approve/ apply run-test?

jiminha:
@regisss  Could you check this PR?  This will run basic accuracy check on the fast test (48min added) to PR. 
Is this ok?

jiminha (approved):

tests/test_functional_text_generation_example.py
regisss: I'm not surprised, greedy search tends to produce very repetitive outputs as it only looks for the most likely token to generate in each iteration. Tweaking some generation parameters (for example to penalize more the sequences of tokens that are already part of the output) and/or using sampling (to consider several candidates and not only the most likely one) usually helps to give the model more flexibility and to get more realistic outputs.

Makefile
```diff
@@ -35,6 +35,7 @@ style: clean
 fast_tests:
 	python -m pip install .[tests]
 	python -m pytest tests/test_gaudi_configuration.py tests/test_trainer_distributed.py tests/test_trainer.py tests/test_trainer_seq2seq.py
+	python -m pytest test_functional_text_generation_example.py
```
regisss, line 38: I think it's a bit tricky to add this one to the fast tests now as it's going to be quite slow with the current infrastructure. Maybe we can comment this line and add a TODO comment to uncomment it when we have more servers dedicated to CI.

regisss (commented):
Very nice PR!

I would just not add this type of check to the fast tests yet, see my comment below. The reason is that I recently migrated the fast tests from DL1 instances to Gaudi2. And I have only one Gaudi2 server for the whole CI (i.e. all the fast tests of all the PRs + nightly tests) vs up to 4 DL1 instances in parallel before. I'm afraid this would increase the queue of workflows quite a lot. Happy to reconsider it if/when I have at least one more server for CI. In the meantime, these checks will still be performed in the nightly tests.

Makefile
vidyasiv: updated

regisss (approved):

PR #1364, opened 2024-09-26
Load INC GPTQ checkpoint & rename params

HolyFalafel:
Added load options for different GPTQ checkpoints.
Modified the load params accordingly

PR #1412, opened 2024-10-14
Load INC GPTQ checkpoint & rename params

HolyFalafel:
Added load options for different GPTQ checkpoints.
Modified the load params accordingly

Same branch as https://github.com/huggingface/optimum-habana/pull/1364 but for main

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ slow_tests_diffusers: test_installs
 
 # Run text-generation non-regression tests
 slow_tests_text_generation_example: test_installs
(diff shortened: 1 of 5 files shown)


regisss:
@HolyFalafel I'm going to cherry-pick in main all the changes that were merged directly into the `v1.14-release`. So, if this PR is the same as #1364, it shouldn't be needed.

HolyFalafel:
> @HolyFalafel I'm going to cherry-pick in main all the changes that were merged directly into the `v1.14-release`. So, if this PR is the same as #1364, it shouldn't be needed.

OK, no problem.
Closing.
Thanks!

PR #1432, opened 2024-10-17
Fix scoped linear all-reduce for starcoder model

skavulya:
A bug in scoped linear all-reduce implementation for starcoder2 model caused incorrect output as shown below:

`python ../gaudi_spawn.py --use_deepspeed --world_size 2 run_generation.py --model_name_or_path bigcode/starcoder2-15b --use_hpu_graphs --trust_remote_code --attn_softmax_bf16 --trim_logits --use_kv_cache --use_flash_attention --flash_attention_recompute --max_new_tokens 128 --batch_size 1 --bf16 --prompt "def is_prime():"`

Output:
Input/outputs:
input 1: ('def is_prime():',)
output 1: ('def is_prime(): ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( ( (',)

Output after the fix:
Input/outputs:
input 1: ('def is_prime():',)
output 1: ('def is_prime():\n    for i in range(2, int(math.sqrt(n)) + 1):\n        if n % i == 0:\n            return False\n    return True\n\ndef is_palindrome():\n    return str(n) == str(n)[::-1]\n\ndef is_lychrel():\n    for i in range(50):\n        n = n + int(str(n)[::-1])\n        if is_palindrome():\n            return False\n    return True\n\ncount = 0\nfor i in range(10000):\n    if is_lychrel():',)

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/modeling_all_models.py
+++ b/optimum/habana/transformers/models/modeling_all_models.py
@@ -164,5 +164,7 @@ class ScopedLinearAllReduce(torch.nn.Module):
             dist.inference_all_reduce(input, group=self.mp_group)
 
     def post_all_reduce(self, input):
-        output = input + self.bias if (self.bias is not None) else input
-        return output
+        # inplace addition needed for correct results
+        if self.bias is not None:
+            input += self.bias
+        return input
--- a/optimum/habana/transformers/models/starcoder2/modeling_starcoder2.py
+++ b/optimum/habana/transformers/models/starcoder2/modeling_starcoder2.py
@@ -61,17 +61,19 @@ logger = logging.get_logger(__name__)
 
 class GaudiStarcoder2MLP(Starcoder2MLP):
     def pre_mlp_forward(self, x):
-        inputs = self.act_fn(self.gate_proj(x)) * self.up_proj(x)
-        output = self.down_proj(inputs)
+        x = self.c_fc(x)
+        x = self.act(x)
+        x = self.c_proj(x)
+        output = F.dropout(x, p=self.residual_dropout, training=self.training)
         return output
 
     def mlp_all_reduce(self, x):
-        if hasattr(self.down_proj, "all_reduce"):
-            self.down_proj.all_reduce(x)
+        if hasattr(self.c_proj, "all_reduce"):
+            self.c_proj.all_reduce(x)
 
     def post_mlp_forward(self, x):
-        if hasattr(self.down_proj, "post_all_reduce"):
-            return self.down_proj.post_all_reduce(x)
+        if hasattr(self.c_proj, "post_all_reduce"):
+            return self.c_proj.post_all_reduce(x)
         return x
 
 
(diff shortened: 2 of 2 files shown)


optimum/habana/transformers/models/starcoder2/modeling_starcoder2.py
```diff
@@ -61,17 +61,19 @@
 
 class GaudiStarcoder2MLP(Starcoder2MLP):
     def pre_mlp_forward(self, x):
-        inputs = self.act_fn(self.gate_proj(x)) * self.up_proj(x)
-        output = self.down_proj(inputs)
+        x = self.c_fc(x)
+        x = self.act(x)
+        x = self.c_proj(x)
+        output = F.dropout(x, p=self.residual_dropout, training=self.training)
         return output
```
yafshar, line 68:
```suggestion
        x= F.dropout(x, p=self.residual_dropout, training=self.training)
        return x
```

@skavulya why do you create a new memory here, is there a reason you do not overwrite x?

optimum/habana/transformers/models/modeling_all_models.py
```diff
@@ -164,5 +164,7 @@ def all_reduce(self, input):
             dist.inference_all_reduce(input, group=self.mp_group)
 
     def post_all_reduce(self, input):
-        output = input + self.bias if (self.bias is not None) else input
-        return output
+        # inplace addition needed for correct results
```
yafshar, line 167: @skavulya very nice catch! 

Does this affect any other tests or models (falcon, llama, gemma, qwen2, qwen2_moe)?

skavulya: I only looked at llama and qwen2. They were not affected because they didn't use a bias. I can test falcon, and qwen2 moe.

optimum/habana/transformers/models/starcoder2/modeling_starcoder2.py
skavulya: Good catch. I will overwrite x similar to the original implementation

https://github.com/huggingface/transformers/blob/f51ac9e059a78049362803c1d606a2c6a8160ee4/src/transformers/models/starcoder2/modeling_starcoder2.py#L199

optimum/habana/transformers/models/modeling_all_models.py
yafshar: @skavulya if llama and qwen2 did not use a bias, then the output is wrong isn't it? since based on the original code `output=input + input`

skavulya: For llama and qwen2, bias was None so the addition is skipped.

skavulya:
@yafshar  The results of the transformer tests on main and this pr are the same

`GAUDI2_CI=1 RUN_SLOW=true python -m pytest tests/transformers/tests/models/`
main: commit f98688d5c172 ==== 41 failed, 993 passed, 401 skipped, 101 warnings in 1531.33s (0:25:31) ====
this pr: ==== 41 failed, 993 passed, 401 skipped, 101 warnings in 1296.93s (0:21:36) ====

yafshar (approved):
LGTM!

@regisss this PR is ready. Would you check it.

regisss (approved):
Good catch!

skavulya:
Thanks @regisss Please include this fix in the 1.14 release. We need it for the TGI gaudi release

regisss:
> Thanks @regisss Please include this fix in the 1.14 release. We need it for the TGI gaudi release

Just added it to the release branch: https://github.com/huggingface/optimum-habana/commits/v1.14-release

PR #1448, opened 2024-10-22
Create CI Eager/Lazy for Language Modeling

Luca-Calabria:
# What does this PR do?

Add a test to CI suite to check/validate LLM training/finetuning as Eager and Lazy Mode. This PR adds Gemma Eager Mode test case.

How to run it manually:
` root@id:~/optimum-habana# GAUDI2_CI=1 RUN_SLOW=1 python -m pytest tests/test_examples.py -v -s -k "single_card"`


Luca-Calabria:
I need an help to catch a "segmentation fault" event like this one (it was the main reason to create this test case)
`mme_descriptor_generator_base.cpp::1114 function: setParams, failed condition: (0), message: input element types should match
mme_descriptor_generator_base.cpp::1114 function: setParams, failed condition: (0), message: input element types should match
mme_descriptor_generator_base.cpp::1114 function: setParams, failed condition: (0), message: input element types should match
mme_descriptor_generator_base.cpp::1114 function: setParams, failed condition: (0), message: input element types should match
mme_descriptor_generator_base.cpp::1114 function: setParams, failed condition: (0), message: input element types should match
Internal Error: Received signal - Segmentation fault` 

It doesn't return to main thread and I can't complete the test run

Luca-Calabria:
@emascarenhas this is the test to add to CI suite. Please take a look when you have time. Thank you

tests/test_language_modeling_example.py
```diff
@@ -0,0 +1,92 @@
...
+    token: str,
+    dataset_name: str,
+    gaudi_config_name: str,
+    use_lazy_mode: bool = False,
+    per_device_train_batch_size: int = 16,
+    per_device_eval_batch_size: int = 16,
+    num_train_epochs: int = 1,
+):
+    command = ["python3"]
+    path_to_example_dir = Path(__file__).resolve().parent.parent / "examples"
```
Chris-Sigopt, line 38: Very minor style point, not blocking. 

Have `Path(__file__).resolve().parent.parent` set to something like OPTIMUM_ROOT as a constant, and then set path to example dir from that constant. Just to make it clearer if someone rearranges the directory structure.

tests/test_language_modeling_example.py
```diff
@@ -0,0 +1,92 @@
...
+        f"--per_device_eval_batch_size {per_device_eval_batch_size}",
+        f"--dataset_name {dataset_name}",
+        "--use_habana",
+        "--do_train",
+        "--do_eval",
+        f"--gaudi_config_name {gaudi_config_name}",
+        "--gradient_checkpointing",
+        "--bf16",
+        f"--num_train_epochs {num_train_epochs}",
+        f"--use_lazy_mode {use_lazy_mode}",
```
Chris-Sigopt, line 57: If we are setting lazy mode here why do we need to set it in the env up above?

tests/test_language_modeling_example.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+prev_quant_model_name = None
+prev_quant_rank = 0
+
+if os.environ.get("GAUDI2_CI", "0") == "1":
+    # Gaudi2 CI baselines
+    MODELS_TO_TEST = {
+        "bf16_1x": [
+            ("google/gemma-2b-it", "mamamiya405/finred", "Habana/gpt2", True, 9.5, 31.5),
```
Chris-Sigopt, line 20: Where do these baseline_train and baseline_eval numbers come from? Just observation of running? Do we have a feel for what the acceptable dispersion of results is? 

Just worried about false positives leading to alert fatigue here.

Chris-Sigopt (approved):
Minor quibbles, otherwise looks good to me.

tests/test_language_modeling_example.py
Luca-Calabria: Actually I just copy-pasted the same structure from the others existent tests. Should I use a different strategy here?

tests/test_language_modeling_example.py
Luca-Calabria: you are right! I remove --use_lazy_mode from commandline

tests/test_language_modeling_example.py
Luca-Calabria: yes baseline come from observation of running

Luca-Calabria:
@Chris-Sigopt any idea about this issue? https://github.com/huggingface/optimum-habana/pull/1448#issuecomment-2429105477

emascarenhas:
> @Chris-Sigopt any idea about this issue? [#1448 (comment)](https://github.com/huggingface/optimum-habana/pull/1448#issuecomment-2429105477)


> @emascarenhas this is the test to add to CI suite. Please take a look when you have time. Thank you

@Luca-Calabria , Can you make this part of test_examples.py instead of a new file?
I think it will run as part of CI if you make it part of test_examples.py and there are already language modeling tests in there.

tests/test_language_modeling_example.py
Chris-Sigopt: Then it's probably not worth it. It would be slightly easier to change in the future but if they have to make the changes anyway probably easier to just have it be the same change every step of the way.

Luca-Calabria:
> > @Chris-Sigopt any idea about this issue? [#1448 (comment)](https://github.com/huggingface/optimum-habana/pull/1448#issuecomment-2429105477)
> 
> > @emascarenhas this is the test to add to CI suite. Please take a look when you have time. Thank you
> 
> @Luca-Calabria , Can you make this part of test_examples.py instead of a new file? I think it will run as part of CI if you make it part of test_examples.py and there are already language modeling tests in there.

@emascarenhas sure, I can. This actually was a question I would ask to you because I saw similar test cases in other existent files. I'll make the new test as part of test_examples.py

Chris-Sigopt:
> as a question I would ask to you because I saw similar test cases in other existent files. I'll make the new test as part of test_examples.py

I have not encountered that particular error message before. Based off the message, my guess would be that some python type coercion is happening incorrectly, which is a problem we've encountered regularly, but that's just a guess.

Luca-Calabria:
@emascarenhas @Chris-Sigopt I moved the new test to test_examples.py as part of Causal Language Modeling test case.
I see that in this test file it covers just Lazy Mode run. Do you think it makes sense to create a test case to cover also the Eager Mode run?

emascarenhas (commented):
@Luca-Calabria , 
About adding Eager test, yes I think you should, because that was one of the failures?

Otherwise, code looks good to me. If you add Eager test, I can do a quick re-review. I suppose you ran the test and it worked without issues and you didn't hit the crash from earlier? Please confirm.

Luca-Calabria:
> @Luca-Calabria , About adding Eager test, yes I think you should, because that was one of the failures?
> 
> Otherwise, code looks good to me. If you add Eager test, I can do a quick re-review. I suppose you ran the test and it worked without issues and you didn't hit the crash from earlier? Please confirm.

@emascarenhas , yep after the fix on command line suggested by @Chris-Sigopt  here https://github.com/huggingface/optimum-habana/pull/1448#discussion_r1828197229 I'm able to run both Lazy and Eager successufully also on older Synapse versions. 
I'll add a test on Egear mode anyway because it is not covered by current CI.

tests/test_examples.py
```diff
@@ -546,6 +570,7 @@ class ExampleTesterBase(TestCase):
         "train_samples_per_second": (TestCase.assertGreaterEqual, 2 - TIME_PERF_FACTOR),
         "eval_samples_per_second": (TestCase.assertGreaterEqual, 2 - TIME_PERF_FACTOR),
     }
+    EAGER_MODE = None
```
Chris-Sigopt, line 573: Should this be False, not None? Just seems like it is trying to be the opposite of True.

Chris-Sigopt (approved):
Minor clarifying issue, other than that looks good to me.

regisss (approved):
LGTM!

regisss:
> The code quality check failed, please run `make style`.

The fail may not be related to this PR, please sync your branch with main and run `make style`.

Luca-Calabria:
@libinta @regisss I merged main branch and fixed "make style" issue

PR #1464, opened 2024-10-30
fixed measurement path bug

Yantom1:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/quantization_tools/unify_measurements.py
+++ b/examples/text-generation/quantization_tools/unify_measurements.py
@@ -28,8 +28,9 @@ def unify_measurements(
 
     # save all the jsons paths in the given measurement group
     for measurement in measurement_group:
-        measurement_path = find_measurement_path(measurement, measurements_dir_path, scales, groups_size)
-        measurements_paths.append(measurement_path)
+        measurement_path = find_measurement_path(measurement, measurements_dir_path, groups_size)
+        if measurement_path is not None:
+            measurements_paths.append(measurement_path)
         group_name += measurement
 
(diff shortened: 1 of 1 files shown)


examples/text-generation/quantization_tools/unify_measurements.py
```diff
@@ -28,8 +28,9 @@ def unify_measurements(
 
     # save all the jsons paths in the given measurement group
     for measurement in measurement_group:
-        measurement_path = find_measurement_path(measurement, measurements_dir_path, scales, groups_size)
-        measurements_paths.append(measurement_path)
+        measurement_path = find_measurement_path(measurement, measurements_dir_path, groups_size)
+        if measurement_path is not None:
+            measurements_paths.append(measurement_path)
```
libinta, line 33: what should we do if the measurement_path is None?

Yantom1: It won't add the path and will not iterate over it in next steps. we shouldn't do anything

libinta: @Yantom1  it seems you didn't change the function find_measurement_path. can you update that on line 9 def find_measurement_path(measurement, measurements_dir_path, scales, group_size): ?

HolyFalafel:
@libinta see here:
https://github.com/huggingface/optimum-habana/pull/1496

examples/text-generation/quantization_tools/unify_measurements.py
libinta: seems this needs PR1496

PR #1502, opened 2024-11-20
add fused kernel config support for run_clm.py

ranzhejiang:
we need to add some fused kernel  config support for run_clm.py（for customer demands), and also add config support for some existing models

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -611,6 +611,12 @@ def setup_tokenizer(args, model, assistant_model, logger):
         )
         model.generation_config.eos_token_id = model.generation_config.eos_token_id[-1]
 
+    if model.config.model_type == "mpt":
+        if tokenizer.pad_token is None:
+            tokenizer.pad_token = tokenizer.eos_token
+        if model.generation_config.pad_token_id is None:
+            model.generation_config.pad_token_id = tokenizer.eos_token_id
+
     # Some models like GPT2 do not have a PAD token so we have to set it if necessary
     if tokenizer.pad_token is None:
         tokenizer.pad_token = tokenizer.eos_token


jiminha:
Can you add one test case using these new parameters and also update on Readme for example? 
https://github.com/huggingface/optimum-habana/blob/main/tests/test_examples.py

jiminha:
@ranzhejiang  can you update based on my comment above if you want this to be included 1.19 release?

ranzhejiang:
> @ranzhejiang can you update based on my comment above if you want this to be included 1.19 release?

Sorry for late reply, I will update this pr follow your suggestions these two days, I need to spend some time writing tests

examples/language-modeling/run_clm.py
```diff
@@ -165,7 +165,50 @@ class ModelArguments:
             )
         },
     )
-
+    attn_softmax_bf16: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Whether to run attention softmax layer in bf16 precision for fine-tuning. The current support is limited to Llama only."
```
emascarenhas, line 172: Besides llama there are other models that support this options, so best to remove the 2nd sentence for this and the next option.

emascarenhas:
@ranzhejiang , We are looking for your updates to move this PR along. Please resync to latest upstream.
Also make sure that "make style" is not generating any new errors. And the previous request above was to add new tests if necessary and make changes to README if needed. Need this in the next few days otherwise we may have to postpone this to a later release and/or close it.

ranzhejiang:
> @ranzhejiang , We are looking for your updates to move this PR along. Please resync to latest upstream. Also make sure that "make style" is not generating any new errors. And the previous request above was to add new tests if necessary and make changes to README if needed. Need this in the next few days otherwise we may have to postpone this to a later release and/or close it.

Thank you for review, I will update it tomorrow.

ranzhejiang:
for some code out of date, I will reopen it when resolved

PR #1503, opened 2024-11-20
fix lora padding loss problem

ranzhejiang:
We must set its label to IGNORE_INDEX for padding token, otherwise the loss can't align with unpadded way


emascarenhas:
@ranzhejiang ,
Please run "pip install -U ruff; make style" and check and fix any reported errors.
Also run tests/ci/fast_tests.sh and all the slow tests related to LoRA language modeling e.g., "GAUDI2_CI=1 RUN_SLOW=1 python -m pytest tests/test_custom_file_input.py" . Also slow tests test_fsdp_examples.py and test_fp8_examples.py. Report results here.

PR #1533, opened 2024-11-27
Avoid negative values in eval metrics

deepak-gowda-narayana:
# What does this PR do?

Fixes the issue of negative throughputs being observed with evaluation while finetuning LLM with OH, which was due to the number of samples being computed as negative when number of warmup steps was greater than number of evaluation steps.

PR does:
1. Clause to check if eval steps <= warmup steps , and set samples and steps used in eval metrics accordingly.

Warmup steps are taken into account for the throughput calculation because the number of evaluation steps ({eval_steps}) is smaller than the number of warmup steps ({self.args.throughput_warmup_steps}).

Example Case : 
Finetuning on 8 card system with : total available samples = 405 , warmup steps = 3 , batch size = 24

Samples used to measure throughput is computed as = ( 405 - 3 * 24 * 8 ) , this gives **-171 samples** resulting in negative evaluation results for throughput

Follow is the eval result prior to PR:

***** eval metrics *****
  eval_accuracy               =     0.6456
  eval_loss                   =     1.6202
  eval_model_preparation_time =     0.0372
  eval_runtime                = 0:00:17.89
  eval_samples                =        405
  **eval_samples_per_second     =     -9.554**
  **eval_steps_per_second       =        0.0**
  max_memory_allocated (GB)   =      44.99
  memory_allocated (GB)       =      26.82
  perplexity                  =     5.0539
  total_memory_available (GB) =      94.62

Eval Result with the Fix :

***** eval metrics *****
  eval_accuracy               =     0.6456
  eval_loss                   =     1.6202
  eval_model_preparation_time =     0.0383
  eval_runtime                = 0:00:18.80
  eval_samples                =        405
  **eval_samples_per_second     =     21.536
  eval_steps_per_second       =      0.112**
  max_memory_allocated (GB)   =      44.99
  memory_allocated (GB)       =      26.82
  perplexity                  =     5.0539
  total_memory_available (GB) =      94.62

## Before submitting


optimum/habana/transformers/trainer.py
```diff
@@ -1763,6 +1763,14 @@ def evaluate(
         num_samples = output.num_samples - self.args.throughput_warmup_steps * total_batch_size
         num_steps = math.ceil(output.num_samples / total_batch_size) - self.args.throughput_warmup_steps
 
+        eval_steps = output.num_samples / total_batch_size
+        if eval_steps <= self.args.throughput_warmup_steps:
+            logger.warning(
+                f" Ignoring warmup steps as number of samples is less to exceed {self.args.throughput_warmup_steps} warmup evaluation steps"
```
regisss, line 1769: Not super clear to me. If I understand correctly, what we do here is to actually compute metrics for the warmup steps since there are only warmup steps right?

deepak-gowda-narayana: @regisss 
Yes we ignore subtracting the samples processed during warmup stage as this leads to negative number of samples when there are only warmup steps.

regisss: Yeah, can we put a clearer message?
Something like "Warmup steps are taken into account for the throughput calculation because the number of evaluation steps ({eval_steps}) is smaller than the number of warmup steps({self.args.throughput_warmup_steps})."

regisss: Thanks, it's very clear! I would just like that you update the message in `logger.warning(...)` because it is not clear there IMO.

deepak-gowda-narayana: My Bad, I thought the PR description was unclear. Updated the warning message in trainer and also rounding up the eval_steps to next integer value instead of using a decimal value

regisss: Thanks!

regisss:
@deepak-gowda-narayana Can you merge the up-to-date main branch into your branch and run `make style` please? And then everything should be good to merge :slightly_smiling_face:

deepak-gowda-narayana:
> @deepak-gowda-narayana Can you merge the up-to-date main branch into your branch and run `make style` please? And then everything should be good to merge 🙂

@regisss synched the branch and ran make style, everything looks good.

ruff check . setup.py --fix
All checks passed!
ruff format . setup.py
401 files left unchanged

regisss (approved):

PR #1512, opened 2024-11-22
add Qwen2-VL static generation

Spycsh:
# What does this PR do?

Add Qwen2-VL static generation.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

PR #1542, opened 2024-12-02
Add Qwen2-VL

nngokhale:
# What does this PR do?
Adds optimized Qwen2-VL to Optimum habana.
Uses HF static cache.
Adds OH static shapes.
Fixes graph recompilations.
Adds FusedSDPA with use_flash_attention kwarg
Removed _"Adds mark_step in eos stopping criterion for reduced graph recompilation time"_
Removed _"Accuracy WA"_
Add Qwen2-VL to examples/image-to-text/run_pipeline.py and Readme
Add Qwen2-VL to tests/test_image_to_text_example

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/image-to-text/README.md
+++ b/examples/image-to-text/README.md
@@ -33,7 +33,6 @@ python3 run_pipeline.py \
 ```
 > SDPA may introduce [reduced precison](https://pytorch.org/docs/stable/notes/numerical_accuracy.html#reduced-precision-reduction-for-fp16-and-bf16-in-scaled-dot-product-attention-sdpa)
 
-
 ### Multi-cards inference with BF16
 
 Use the following commands to run Llama-3.2-90B-Vision-Instruct BF16 inference with FusedSDPA on 8 HPUs:
--- a/examples/image-to-text/README.md
+++ b/examples/image-to-text/README.md
@@ -33,6 +33,7 @@ python3 run_pipeline.py \
 ```
 > SDPA may introduce [reduced precison](https://pytorch.org/docs/stable/notes/numerical_accuracy.html#reduced-precision-reduction-for-fp16-and-bf16-in-scaled-dot-product-attention-sdpa)
 
+
 ### Multi-cards inference with BF16
 
 Use the following commands to run Llama-3.2-90B-Vision-Instruct BF16 inference with FusedSDPA on 8 HPUs:
--- a/optimum/habana/transformers/generation/stopping_criteria.py
+++ b/optimum/habana/transformers/generation/stopping_criteria.py
@@ -17,6 +17,7 @@
 import time
 from typing import Union
 
+import habana_frameworks.torch.core as htcore
 import torch
 
 from optimum.utils import logging
@@ -67,6 +68,7 @@ def gaudi_MaxTimeCriteria_call(
 def gaudi_EosTokenCriteria_call(
     self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs
 ) -> Union[torch.BoolTensor, bool]:
+    htcore.mark_step()
     self.eos_token_id = self.eos_token_id.to(input_ids.device)
     token_idx = kwargs.get("token_idx", None)
     if token_idx is not None:
--- a/optimum/habana/transformers/generation/stopping_criteria.py
+++ b/optimum/habana/transformers/generation/stopping_criteria.py
@@ -17,7 +17,6 @@
 import time
 from typing import Union
 
-import habana_frameworks.torch.core as htcore
 import torch
 
 from optimum.utils import logging
@@ -68,7 +67,6 @@ def gaudi_MaxTimeCriteria_call(
 def gaudi_EosTokenCriteria_call(
     self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs
 ) -> Union[torch.BoolTensor, bool]:
-    htcore.mark_step()
     self.eos_token_id = self.eos_token_id.to(input_ids.device)
     token_idx = kwargs.get("token_idx", None)
     if token_idx is not None:
--- a/optimum/habana/transformers/models/qwen2_vl/modeling_qwen2_vl.py
+++ b/optimum/habana/transformers/models/qwen2_vl/modeling_qwen2_vl.py
@@ -1,0 +1,774 @@
+# coding=utf-8
+# Copyright 2024 the HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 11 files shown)


optimum/habana/transformers/generation/stopping_criteria.py
```diff
@@ -67,6 +68,7 @@ def gaudi_MaxTimeCriteria_call(
 def gaudi_EosTokenCriteria_call(
     self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs
 ) -> Union[torch.BoolTensor, bool]:
+    htcore.mark_step()
```
ssarkar2, line 71: this is a general change, and may affect performance in other models.

Can you add some numbers for the compile times u have with and without this change?

ssarkar2: is it possible to add teh markstep at teh end of the qwen model or something like that so the effect of this is isolated or qwen only?


otherwise we'd have to do a larger study to make sure it doesnt have any bad effect on any other model/usecase

nngokhale: mark_step at the end of the model doesn't have the same effect. 
There is 2x improvement in warmup time. 
I agree this should be independent PR with more tests.
I will remove this particular change.

ssarkar2 (approved):

optimum/habana/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -0,0 +1,763 @@
...
+        attn_output = self.proj(attn_output)
+        del attention_mask
+        return attn_output
+
+
+GAUDI_QWEN2_VL_VISION_ATTENTION_CLASSES = {
+    "eager": VisionAttention,
+    "flash_attention_2": VisionFlashAttention2,
+    "sdpa": GaudiQwen2VisionSdpaAttention,
+}
```
libinta, line 107: why do we need this? can you check how other model is done? only overwrite the function/class needed

optimum/habana/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -0,0 +1,763 @@
...
+        attn_output = self.o_proj(attn_output)
+
+        return attn_output, None, past_key_value
+
+
+GAUDI_QWEN2_VL_ATTENTION_CLASSES = {
+    "eager": Qwen2VLAttention,
+    "flash_attention_2": Qwen2VLFlashAttention2,
+    "sdpa": GaudiQwen2VLSdpaAttention,
+}
```
libinta, line 300: please check how other model rewrite the attention layer

libinta (commented):
for each changed function, please indicate source and what's the change

optimum/habana/transformers/models/qwen2_vl/modeling_qwen2_vl.py
nngokhale: Removed

optimum/habana/transformers/models/qwen2_vl/modeling_qwen2_vl.py
nngokhale: This was part of original code, now removed, only Sdpa+FusedSdpa attention is retained.

regisss:
@nngokhale Is it the same as #1512 ?

nngokhale:
> @nngokhale Is it the same as #1512 ?

Hi @regisss, not the same, Implementation differs. This PR uses static cache, resolve graph recompilation / hpu graph issue, and adds fused sdpa.

Spycsh:
> @nngokhale Is it the same as #1512 ?

@regisss @libinta I've done some benchmarks and it shows @nngokhale 's PR is faster than mine (both warmup/inference stage for qwen2_vl 2B and 7B). I think one of the main reason is that he is able to wrap the whole model in the HPU graph, including Vision Block. If you look at my implementation [here](https://github.com/Spycsh/optimum-habana/blob/qwen2_vl/examples/image-to-text/run_pipeline.py#L325), I just wrap the HPU graph of the language model part. This is a workaround because wrap the whole HPU graph issue can cause a annoying `FATAL ERROR :: MODULE:PT_DEVMEM Allocation failed for size...` bug with the loop [test](https://github.com/Spycsh/qwen-vl-hpu/blob/main/loop_test.py). Therefore, the workaround is to simply wrap the language model part into the HPU graph. This PR actually does a better job because there seems no such bug (I think it may be because he uses the internal static cache `use_cache=True, cache_implementation="static", static_shapes=True,`, which is more robust than handwriting the past_key_values static updating when prefill/decoding like [mine](https://github.com/Spycsh/optimum-habana/blob/qwen2_vl/optimum/habana/transformers/models/qwen2_vl/modeling_qwen2_vl.py#L111).

Meanwhile, @nngokhale also added the FusedSDPA part which made it slightly faster.

Overall, I think @nngokhale 's PR is a better choice for you to merge. Please feel free to close my PR is you have no more questions!

examples/image-to-text/README.md
```diff
@@ -35,6 +35,7 @@ Models that have been validated:
   - [meta-llama/Llama-3.2-90B-Vision-Instruct](https://huggingface.co/meta-llama/Llama-3.2-90B-Vision-Instruct)
   - [tiiuae/falcon-11B-vlm](https://huggingface.co/tiiuae/falcon-11B-vlm)
   - [google/paligemma-3b-mix-224](https://huggingface.co/google/paligemma-3b-mix-224)
+  - [Qwen/Qwen2-VL-7B-Instruct](https://huggingface.co/Qwen/Qwen2-VL-7B-Instruct)
```
IlyasMoutawwakil, line 38: can you please add the new model type to the repo's README.md table as well as the docs/source/index.mdx

IlyasMoutawwakil (commented):
LGTM, need to add the model family to the tables in README and docs

IlyasMoutawwakil:
Also since static shapes is support is added as well, shouldn't `qwen2_vl` be added to `MODELS_OPTIMIZED_WITH_STATIC_SHAPES` in `optimum/habana/transformers/generation/utils.py`

examples/image-to-text/README.md
nngokhale: Done

nngokhale:
> Also since static shapes is support is added as well, shouldn't `qwen2_vl` be added to `MODELS_OPTIMIZED_WITH_STATIC_SHAPES` in `optimum/habana/transformers/generation/utils.py`

Added to MODELS_OPTIMIZED_WITH_STATIC_SHAPES.

examples/image-to-text/README.md
IlyasMoutawwakil: okay just saw it ! thanks !

examples/image-to-text/README.md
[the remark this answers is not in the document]
nngokhale: ok, done

regisss (approved):

PR #1549, opened 2024-12-04
Fix for single_card llama7b and falcon40b CI errors

MohitIntel:
fixes regression in single_card CI due to PR#1529


tests/baselines/Meta_Llama_3_1_8B.json
regisss, line 1: Why are we removing this from CI?

MohitIntel: Not removing "LLama_3_1_8b.json" (which is original file still kept in the baseline folder). This file named "Meta_LLama_3_1_8b.json" is a duplicate that I forgot to remove in PR #1529.

regisss: oh okay I had missed that!

regisss (approved):

PR #1519, opened 2024-11-22
Add support for optimized SDXL pipeline

sushildubey171:
# What does this PR do?

This PR adds support for sdxl to use optimized pipeline from mlperf.
Also enable fp8 support in sdxl for optimized pipeline.
Readme is updated for sdxl 

Fixes # (issue)
This also fixes perf reporting for default sdxl pipeline.

PR #1554, opened 2024-12-04
Add diffuser tests for optimized sdxl flow on HPU

sushildubey171:
To add diffuser tests for sdxl optimized flow for both bf16 and fp8 mode, verified and run locally 
jira: SW-211820
use case PR [#1519](https://github.com/huggingface/optimum-habana/pull/1519)


tests/test_diffusers.py
```diff
@@ -1329,6 +1329,144 @@ def test_stable_diffusion_xl_hpu_graphs(self):
...
+        logger = logging.get_logger(sd_pipe.__module__)
+        logger.setLevel(level=diffusers.logging.INFO)
+        sd_pipe.unet.set_default_attn_processor(sd_pipe.unet)
+        sd_pipe.to(torch.device("hpu"))
+
+        quant_config_path = os.getenv("QUANT_CONFIG")
+
+        original_dir = os.getcwd()
+        config_dir = os.path.dirname(os.path.abspath("../examples/stable-diffusion/quantization"))
+        os.chdir(config_dir)
```
regisss, line 1439: Is this absolutely necessary?

tests/test_diffusers.py
```diff
@@ -1329,6 +1329,144 @@ def test_stable_diffusion_xl_hpu_graphs(self):
...
+        )
+
+        model_name_or_path = "stabilityai/stable-diffusion-xl-base-1.0"
+
+        sd_pipe = StableDiffusionXLPipeline_HPU.from_pretrained(
+            model_name_or_path,
+            **kwargs,
+        )
+        logger = logging.get_logger(sd_pipe.__module__)
+        logger.setLevel(level=diffusers.logging.INFO)
```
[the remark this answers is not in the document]
sushildubey171: yes

tests/test_diffusers.py
sushildubey171: yes, this is required to avoid file not found error because we are executing inside from tests/, for below measurement file, it is present in examples/stable-diffusion/quantization
https://github.com/huggingface/optimum-habana/blob/main/examples/stable-diffusion/quantization/quant_config.json#L6

in the end I do restore the original directory

tests/test_diffusers.py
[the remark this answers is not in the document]
sushildubey171: ok, I will replace with print()

tests/test_diffusers.py
sushildubey171: there is no way to specify as an argument, measurement file path is specified in above config file which is read at runtime by required Intel neural compressor library (INC), so I am executing from that directory, to avoid file not found error. let me know if you see any issue here, other way could be by copying the quantization directory in current test directory and execute.

regisss: Okay, that's good enough, thanks for the explanation!

tests/test_diffusers.py
```diff
@@ -1329,6 +1329,141 @@ def test_stable_diffusion_xl_hpu_graphs(self):
...
+            "scheduler": scheduler,
+            "use_habana": True,
+            "use_hpu_graphs": True,
+            "gaudi_config": "Habana/stable-diffusion",
+            "torch_dtype": torch.bfloat16
+        }
+
+        os.environ["PATCH_SDPA"] = "1"
+        # Set QUANT_CONFIG environment variable
+        os.environ["QUANT_CONFIG"] = "./quantization/quant_config.json"
```
regisss, line 1418: The structure of the quantization folder has changed: https://github.com/huggingface/optimum-habana/tree/main/examples/stable-diffusion/quantization
Can you update your PR accordingly please?

sushildubey171: done

regisss:
@sushildubey171 I just pushed a change to make the `config_dir` more generic as we don't necessarily call the tests from the `optimum-habana/tests` folder. Please let me know if that still works on your side.

regisss (approved):

PR #1565, opened 2024-12-05
fusedsdpa for stable diffusion xl

skaulintel:
-Use fusedsdpa in stable diffusion xl pipeline
-modify sd and sdxl CI tests to use the above


optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
```diff
@@ -50,6 +56,48 @@ class GaudiStableDiffusionXLPipelineOutput(BaseOutput):
     images: Union[List[PIL.Image.Image], np.ndarray]
     throughput: float
 
+def set_attn_processor_hpu(self, processor: Union[AttentionProcessor, Dict[str, AttentionProcessor]]):
```
libinta, line 59: put these functions in utils and refer here

regisss: We should rather put it in `optimum.habana.diffusers.models.unet_2d_condition.UNet2DConditionModel`, to make it similar to Diffusers. It will be easier to find and maintain. I'm going to push a commit to do this.

PR #1573, opened 2024-12-06
Added mamba model support and test CI script

zzhang37:
# What does this PR do?
Added Mamba model support using custom op, and added the test on Mamba, the test command is
GAUDI2_CI=1 RUN_SLOW=1 python -m pytest tests/test_text_generation_example.py -s -v -k "mamba"
It passed on 1.18.0release and 1.19.0, build 497 with corresponding so files.

Fixes # (issue)

## Before submitting


zzhang37:
Just run the command
GAUDI2_CI=1 RUN_SLOW=1 python -m pytest tests/test_text_generation_example.py -s -v -k "mamba"
It will pass on 1.18 and 1.19 version

setup.py
```diff
@@ -36,6 +36,7 @@
     "diffusers >= 0.31.0, < 0.32.0",
     "huggingface_hub >= 0.24.7",
     "sentence-transformers == 3.2.1",
+    "wget",
```
regisss, line 39: We should not use that to download files on the HF hub but: https://huggingface.co/docs/huggingface_hub/v0.26.3/en/package_reference/file_download#huggingface_hub.hf_hub_download

examples/text-generation/README.md
```diff
@@ -198,6 +198,24 @@ python ../gaudi_spawn.py --use_deepspeed --world_size 8 run_generation.py \
...
+--use_hpu_graphs \
+--use_kv_cache \
+--batch_size 1024 \
+```
+We need to download two shared libraries and set two env variables for mamba, for 1.19 version use *_119 so files. For different release, use different set of these so files.
+```bash
+wget https://huggingface.co/Habana/mamba/resolve/main/hpu_custom_pscan_all.cpython-310-x86_64-linux-gnu_119.so
+wget https://huggingface.co/Habana/mamba/resolve/main/libcustom_tpc_perf_lib_119.so
+export HABANA_CUSTOM_OP_DIR=/path/of/hpu_custom_pscan_all.cpython-310-x86_64-linux-gnu_119.so located
+export GC_KERNEL_PATH=/path/to/libcustom_tpc_perf_lib_119.so:$GC_KERNEL_PATH
```
regisss, line 217: This should be done automatically in the script if the model type is Mamba

README.md
```diff
@@ -241,6 +241,7 @@ The following model architectures, tasks and device distributions have been vali
 | MiniCPM3 |   | <div style="text-align:left"><li>Single card</li></div> | <li>[text generation](https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation)</li> |
 | Baichuan2 |   | <div style="text-align:left"><li>Single card</li></div> | <li>[text generation](https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation)</li> |
 | DeepSeek-V2 |   | :heavy_check_mark: | <li>[text generation](https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation)</li> |
+| Mamba |   | <div style="text-align:left"><li>Single card</li></div> |  <li>[text generation](https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation)</li> |
```
regisss, line 244: Let's add the same to the doc

examples/text-generation/README.md
zzhang37: The test did the automatically download, but for users want to run the mamba, they can just download the so files form the Habana/mamba card.

zzhang37: I also remove the wget and use hf_hub_download in the test script.

README.md
zzhang37: Added.

examples/text-generation/README.md
regisss: I think this will discourage many users to use this model. Can't we put the code you added to the test file to donwload the .so in a new utils file in the Mamba modeling folder? And then we could call it here and in the test.

zzhang37: This is just readme file.

tests/test_text_generation_example.py
```diff
@@ -218,6 +217,38 @@ def _test_text_generation(
...
 
+    if "mamba" in model_name.lower():
+        from huggingface_hub import hf_hub_download
+        import subprocess
+
+        cmd1 = subprocess.Popen(["pip", "list"], stdout=subprocess.PIPE)
+        cmd2 = subprocess.Popen(["grep", "habana-torch-plugin"], stdin=cmd1.stdout, stdout=subprocess.PIPE)
+        cmd1.stdout.close()
+        version_no, _ = cmd2.communicate() 
+
```
jiminha, line 228: Did you try just using check_optimum_habana_min_version?

setup.py
zzhang37: This one is already removed.

zzhang37:
Fixed the make style issue.

examples/text-generation/README.md
regisss: I'm not sure what you mean. I think automatically downloading these .so files in the script if the model type is mamba would give a better user experience. Why not doing it there?

tests/test_text_generation_example.py
```diff
@@ -43,7 +43,7 @@
             ("Qwen/Qwen1.5-7B", 4, False, 490.8621617893209, False),
             ("google/gemma-7b", 1, False, 109.70751574382221, True),
             ("google/gemma-2-9b", 1, False, 92.302359446567, True),
-            ("state-spaces/mamba-130m-hf", 1536, False, 5385.511100161605, False),
+            ("state-spaces/mamba-130m-hf", 1536, False, 19283.0330042467, False),
```
vidyasiv, line 46: could you enable check_output=true and add functional testing output to MODEL_OUTPUTS
```suggestion
            ("state-spaces/mamba-130m-hf", 1536, False, 19283.0330042467, True),
```

examples/text-generation/README.md
jiminha: @zzhang37  it seems the mamba run code fails if we don't set this environment variable at all. Is this intended? Can customer run with/without custom code? If the intention is always run mamba with custom kernel, we should put all these download/setup code in modeling_mamba file at the beginning before you load the library.

zzhang37: @regisss Removed the all wget and added util to do the job.
@jiminha Just checked with latest build 538, works fine.

tests/test_text_generation_example.py
zzhang37: Done.

tests/test_text_generation_example.py
vidyasiv: does the test actually check the output,you didn't update MODEL_OUTPUTS dictionary with model and text output

zzhang37: It can pass with setting True. But I think we don't need to change it this time.

vidyasiv: @zzhang37, you misunderstand me, test "passes" cause it does not enter the loop as model_name is not in MODEL_OUTPUTS keys, please re-read that part of the code- line 364.
we can remove that to make it `if check_output:` instead as devs might make the same mistake you did.

@regisss up to you and @jiminha  if you want to enforce functional testing but this is the reason it was added, to encourage devs to update with models.

regisss: Yes I agree with @vidyasiv.
@zzhang37 You simply have to run the test with the `-s` flag to display the outputs. And then you can simply copy-paste the output of the model in the `MODEL_OUTPUTS` dictionary.

zzhang37: Done.

vidyasiv: thanks @zzhang37

regisss:
@zzhang37 There are some new issues now but these should be quite straightforward to fix:
- Calling `adapt_transformers_to_gaudi` will return an error if `GC_KERNEL_PATH` has not been specified before. I suggest to run [this piece of code](https://github.com/zzhang37/optimum-habana/blob/a87ed29b102e75095a854c9e912167fe8c46d537/optimum/habana/transformers/models/mamba/modeling_mamba.py#L18-L42) only if `GC_KERNEL_PATH` is in `os.environ`. If I understood correctly, `GC_KERNEL_PATH` should be specified if the user intends to use Mamba right?
- The test doesn't work because `GC_KERNEL_PATH` is not specified. You should specify it similarly to how it is done [here](https://github.com/zzhang37/optimum-habana/blob/a87ed29b102e75095a854c9e912167fe8c46d537/tests/test_text_generation_example.py#L229) if the tested model is Mamba.

zzhang37:
@regisss The problem is GC_KERNEL_PATH needs to be set before the start of the script since Synpase needs to use this path to load the TPC kernels. There is a default path for GC_KERNEL_PATH, but mamba required an additional path to add. So that is the reason if we run mamba model, we need to add additional path to GC_KERNEL_PATH beforehand rather than in the middle of the script.

PR #1500, opened 2024-11-20
add check_neural_compressor_min_version for 4 bit behavior

xin3he:
# What does this PR do?

add check_neural_compressor_min_version for 4 bit loading behavior to fix [SW-195965].
It is fixed when INC version >= 3.2

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1578, opened 2024-12-09
Revert "add check_neural_compressor_min_version for 4 bit behavior"

xin3he:
Reverts huggingface/optimum-habana#1500

Currently, `neural_compressor_pt` package misses packing the version info. So we need to revert it.
`
AttributeError: module 'neural_compressor' has no attribute '__version__'
 `

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -38,7 +38,6 @@ from optimum.habana.checkpoint_utils import (
 )
 from optimum.habana.utils import (
     check_habana_frameworks_version,
-    check_neural_compressor_min_version,
     check_optimum_habana_min_version,
     get_habana_frameworks_version,
     set_seed,
@@ -278,8 +277,9 @@ def setup_model(args, model_dtype, model_kwargs, logger):
             original_model=org_model,
             **model_kwargs,
         )
-        if not check_neural_compressor_min_version("3.2"):
-            model = model.to(model_kwargs["torch_dtype"])
+        # TODO: This will be removed in v1.19 Synapse release
+        # the loaded model should have the same dtype as original_model
+        model = model.to(model_kwargs["torch_dtype"])
     else:
         if args.assistant_model is not None:
             assistant_model = AutoModelForCausalLM.from_pretrained(
--- a/optimum/habana/utils.py
+++ b/optimum/habana/utils.py
@@ -385,15 +385,6 @@ def check_habana_frameworks_version(req_version):
     )
 
 
-def check_neural_compressor_min_version(req_version):
-    """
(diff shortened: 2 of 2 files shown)


examples/text-generation/utils.py
```diff
@@ -278,8 +277,9 @@ def setup_model(args, model_dtype, model_kwargs, logger):
             original_model=org_model,
             **model_kwargs,
         )
-        if not check_neural_compressor_min_version("3.2"):
-            model = model.to(model_kwargs["torch_dtype"])
+        # TODO: This will be removed in v1.19 Synapse release
+        # the loaded model should have the same dtype as original_model
+        model = model.to(model_kwargs["torch_dtype"])
```
jiminha, line 282: @libinta  we don't need this code at all for 1.19 docker, but need it for 1.18. Can we remove completely for 1.15 release?  Or do we need to have it?

jiminha: @xin3he  Can you confirm running this code on 1.19 has no effect(no harm)?

xin3he: no harm, but we'd better remove it for 1.19

regisss: We can merge this PR into the release branch first. And once Synapse 1.19 is released, I'll cherry-pick it in main.

I agree with @jiminha that we should remove this block completely for 1.19.

libinta: @regisss  let's merge this change now in case people are still using 1.18

regisss: I don't understand, this change is for 1.18 compatibility?

jiminha: @regisss  the problem is it seems "version.Version(neural_compressor.__version__)" doesn't work in either 18 or 19 docker. So we can't run any INT4 quantization now with an error (AttributeError: module 'neural_compressor' has no attribute '__version__') with main.   

In synapse_8, Xin is saying we need this code "model.to(model_kwargs["torch_dtype"])" since the model is running int32  instead of bf16 without this. On 19, it seems we don't need this at all.   

Main, we still need this change since it's not working. For 19 release, we can remove this whole thing.

regisss: Oh okay got it, so basically we should merge it into main but no need to cherry-pick this change into the release branch (since the release branch is tailored for 1.19) right?

jiminha: For the 1.19 branch, I think it's no harm to have this line, but it's better to remove it completely. I will discuss with Libin and submit patch if we want to go that way.

regisss (approved):

PR #1594, opened 2024-12-12
Revert common KVCache not to check token_idx

jiminha:
# What does this PR do?

Fixed common KVCache code NOT to have token_idx check to fix lm_eval with reuse_cache. (Readme validation). 
Added LlamaKVCache to have token_idx check handle llama prompt finetuning. 

Fixes # (issue)
Command for Falcon180b, Phi lm_Eval from Readme(https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/README.md#running-with-fp8) is failing
Command:
```
python run_lm_eval.py -o falcon.eval --model_name_or_path tiiuae/falcon-7B --use_hpu_graphs --use_kv_cache --trim_logits --batch_size 1 --bf16 --reuse_cache --use_flash_attention --flash_attention_recompute --flash_attention_causal_mask
```
Error 
```
2024-12-12 17:55:48.844] [COMPLEXGUID_LIB] [error] 'tpckernel.sdpa_recomp_core_fwd' op Broadcast of shape [1, 1, 384, 384] not possible at index 3. Dimension 384 incompatible with output shape dimension 769Broadcast of shape [1, 1, 384, 384] not possible at index 3. Dimension 384 incompatible with output shape dimension 769
[17:55:48.845182][SYN_RECIPE    ][error][tid:3635] compileGraph: Can not compile graph
[17:55:48.845204][SYN_RECIPE    ][error][tid:3635] addRecipeHandleAndCompileGraph: Can not compile
[17:55:48.845215][SYN_API       ][error][tid:3635] compileGraph: Failed to add recipe handle into Recipe-Singleton status 26[synFail]
[17:55:48.845224][SYN_API       ][error][tid:3635] synGraphCompile: SYN_API_CALL failed status: 26[synFail]. graphHandle 0x55812f4a6ce0
```

## Before submitting

### DIFF
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -487,7 +487,6 @@ QUANT_CONFIG=./quantization_config/maxabs_measure_include_outputs.json python ..
 --trim_logits \
 --batch_size 1 \
 --bf16 \
---reuse_cache \
 --use_flash_attention \
 --flash_attention_recompute \
 --flash_attention_causal_mask
--- a/optimum/habana/transformers/models/llama/modeling_llama.py
+++ b/optimum/habana/transformers/models/llama/modeling_llama.py
@@ -30,9 +30,48 @@ from ....distributed.tp import TPModule
 from ...modeling_attn_mask_utils import (
     _gaudi_prepare_4d_causal_attention_mask,
 )
-from ..modeling_all_models import KVCache, Matmul, apply_customized_rope_module
+from ..modeling_all_models import Matmul, apply_customized_rope_module
 from .configuration_llama import LlamaConfig
 
+class KVCache(torch.nn.Module):
+    def __init__(self):
+        super(KVCache, self).__init__()
+        self.cache = None
+        self.inp_seq_len = -1
+
+    def allocate(self, inp_seq_len, dtype, device, shape):
+        if self.cache is None or self.cache.shape != shape:
+            self.inp_seq_len = inp_seq_len
+            self.cache = torch.zeros(shape, dtype=dtype, device=device)
+        else:
+            assert (
+                self.inp_seq_len == inp_seq_len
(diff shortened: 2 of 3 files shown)


jiminha:
@ssarkar2  @sywangyi  @regisss  Please review. I'm running a few more test, but if there is better way to do it, please also comment here.

regisss:
A couple of questions:
- Why not checking if we are currently training instead of checking if the index is None?
- If Llama also raises an error, how does that PR solve it? Since the implementation for Llama would still be the same

sywangyi:
I will file a PR to fix llama case in lm eval.
there are three cases needed consideration
1.lm_eval case. not use model.generate. use model.forward. so token idx is None. if reuse_cache is used. self.cache is not None. prev.copy_(cur) should be used
2. generate case. token idx is not None.prev.copy_(cur) should be used as well.
3. other case. should use torch.cat (prefix tuning finetune and eval)

jiminha:
> A couple of questions:
> 
> * Why not checking if we are currently training instead of checking if the index is None?
> * If Llama also raises an error, how does that PR solve it? Since the implementation for Llama would still be the same

Original commit was to solve all the issue(checking self.training), but we found further issue that's not working for all case(ex. eval during prompt finetuning), discussed further with Libin/Yi last night and decided that we will put the same code as v1.14 back to minimize the risk, and have Yi further look and fix all the cases.

jiminha:
> @jiminha If you want a specific implementation for Llama, you could just define a new class that inherits from `KVCache` and overrides the right method:
> 
> ```python
> class LlamaKVCache(KVCache);
>     @staticmethod
>     def update(...)
>         ...
> ```
> 
> That way there is less duplicated code.

will update.

regisss (approved):

PR #1501, opened 2024-11-20
Adding support for Context Parallelism using Deepseed's DistributedAt…

bhargaveede:
Adding support for Context Parallelism using Deepseed's DistributedAttention

This PR adds support to enable Context Parallelism for Llama models using deepspeed.
This feature can be enabled by using context_parallel_size flag.

This feature enables us to train/eval longer context lengths by parallelizing inputs across Context Parallel group.
For Attention, It uses Deepspeed's DistributedAttention which gathers the sequences for all heads and distributes the heads across the Context Parallel Group so that Attention per head has entire context and are distributed within the Context Parallel group.
Once the attention is done, Outputs are scattered based on sequence length across the group.

Verified Llama3.1 8B and Llama 3.1 70B finetuning for 32K seq length on 8 ranks using this feature.

- Could fit 32K sequence length Llama 3.1-8B LoRa finetuning on 4 ranks with selective recompute instead of gradient_checkpointing.
- Could fit 32k sequence length for Llama 3.1-70B LoRa finetuning on 4 ranks and 8 ranks with gradient_checkpointing 

Llama 3.1-8B command
[...]

PR #1598, opened 2024-12-12
Limit position embeddings in inference

bhargaveede:
Moving the max_position_embeddings truncation from model code as it hampers with training.
Moved it to text-generation/utils.py as that's better place

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -441,6 +441,8 @@ def setup_distributed_model(args, model_dtype, model_kwargs, logger):
     if load_to_meta:
         # Construct model with fake meta tensors, later will be replaced on devices during ds-inference ckpt load
         with deepspeed.OnDevice(dtype=model_dtype, device="meta"):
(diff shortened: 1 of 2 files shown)


bhargaveede:
@regisss There's a bug which got exposed now and causing OOM with https://github.com/huggingface/optimum-habana/commit/0fbc4576796a69eaca10166b50f556e43a741eae
.
This change is to fix that in a better way such that it won't interfere with training.
Please review and merge this to main and also 1.15-release

regisss (approved):
LGTM.

@bhargaveede Can you also address the unsolved comments in https://github.com/huggingface/optimum-habana/pull/1501 please? At least moving `parallel_state.py` to `optimum/habana/distributed`. Using more recent versions of Llama for CI can be done after release, but moving `parallel_state.py` should be done before release. Otherwise that will be a breaking change in the release after and we should avoid that. It's just about moving one file and updating a couple of imports.

PR #1601, opened 2024-12-12
SDPA flag update - static code analysis

astachowiczhabana:
This PR reviews and updates all models in the Optimum-Habana repository that use Scaled Dot-Product Attention (SDPA) to ensure compatibility with the PyTorch 2.5 upgrade. It also revises all affected examples and README files. These models were identified through static code analysis rather than QA detection.


skaulintel (approved):
lgtm

regisss (approved):

PR #1617, opened 2024-12-17
Add the inline_inbuilt_nn_modules option 

chaojun-zhang:
Disabling the inline_inbuilt_nn_modules switch will restore ALBERT-XXL's performance to PyTorch 2.4 levels

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -892,7 +892,6 @@ class GaudiTrainingArguments(TrainingArguments):
         if self.torch_compile and self.cache_size_limit is not None:
             torch._dynamo.config.cache_size_limit = self.cache_size_limit
 
-
         logger.info("PyTorch: setting up devices")
         if not is_accelerate_available():
             raise ImportError(
--- a/optimum/habana/transformers/training_args.py
+++ b/optimum/habana/transformers/training_args.py
@@ -103,6 +103,8 @@ class GaudiTrainingArguments(TrainingArguments):
             Set value of 'dynamic' parameter for torch.compile.
         use_regional_compilation (`bool`, *optional*, defaults to `False`):
             Whether to use regional compile with deepspeed
+        inline_inbuilt_nn_modules (`bool`, *optional*, defaults to `None`):
+            Set value of 'inline_inbuilt_nn_modules' parameter for torch._dynamo.config. Currently, disabling this parameter improves the performance of the ALBERT model.
         disable_tensor_cache_hpu_graphs (`bool`, *optional*, defaults to `False`):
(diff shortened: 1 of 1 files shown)


ssarkar2 (approved):
Adding notes for my understanding

without this change:
torch._dynamo.config.inline_inbuilt_nn_modules = True
https://github.com/pytorch/pytorch/blob/main/torch/_dynamo/config.py#L412C33-L412C42

with this change we give an option to set it to True or False. We set it to false in case of albert

optimum/habana/transformers/training_args.py
```diff
@@ -170,6 +172,11 @@ class GaudiTrainingArguments(TrainingArguments):
         metadata={"help": ("Set value of 'dynamic' parameter for torch.compile.")},
     )
 
+    inline_inbuilt_nn_modules: Optional[bool] = field(
+        default=None,
```
regisss, line 176: If I understand correctly, this should be True for all models except ALBERT right?
So we might as well set the default to True no?

optimum/habana/transformers/training_args.py
```diff
@@ -101,6 +101,8 @@ class GaudiTrainingArguments(TrainingArguments):
             Whether to use compiled autograd for training. Currently only for summarization models.
         compile_dynamic (`bool|None`, *optional*, defaults to `None`):
             Set value of 'dynamic' parameter for torch.compile.
+        inline_inbuilt_nn_modules (`bool`, *optional*, defaults to `None`):
+            Set value of 'inline_inbuilt_nn_modules' parameter for torch._dynamo.config.
```
mlapinskix, line 105: If default value is dependent on model, maybe add this information in argument description.

ssarkar2:
@chaojun-zhang could you please resolve conflict. We'll try to merge it in for 1.20

optimum/habana/transformers/training_args.py
chaojun-zhang: We want to keep the default value of this switch consistent with different versions of Pt. For example, pt 2.4 defaults to False, while pt 2.5 defaults to True. Therefore, using None here means not applying any configuration to maintain alignment with the respective version's defaults.

chaojun-zhang:
> @chaojun-zhang could you please resolve conflict. We'll try to merge it in for 1.20

Updated

optimum/habana/transformers/training_args.py
chaojun-zhang: Currently, this parameter is effective for ALBERT, but this does not imply that other models won't benefit from it. Disabling it might also improve performance for other models, though this has not been tested yet

regisss:
@chaojun-zhang One more merge conflict to solve since I just merged #1619

regisss (approved):
LGTM!

PR #1631, opened 2024-12-18
add sdp_on_bf16 to controlnet

skaulintel:
add sdp_on_bf16 in train_controlnet script. allows corresponding diffusers CI test to pass.

### DIFF
--- a/examples/stable-diffusion/text_to_image_generation.py
+++ b/examples/stable-diffusion/text_to_image_generation.py
@@ -441,7 +441,7 @@ def main():
     kwargs_call["quant_mode"] = args.quant_mode
 
     # Instantiate a Stable Diffusion pipeline class
-    #import habana_frameworks.torch.core as htcore  # noqa: F401
+    import habana_frameworks.torch.core as htcore  # noqa: F401
 
     if sdxl:
         # SDXL pipelines
--- a/examples/stable-diffusion/text_to_image_generation.py
+++ b/examples/stable-diffusion/text_to_image_generation.py
@@ -441,7 +441,7 @@ def main():
     kwargs_call["quant_mode"] = args.quant_mode
 
     # Instantiate a Stable Diffusion pipeline class
-    import habana_frameworks.torch.core as htcore  # noqa: F401
+    #import habana_frameworks.torch.core as htcore  # noqa: F401
 
     if sdxl:
         # SDXL pipelines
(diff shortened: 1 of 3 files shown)


regisss (commented):
@skaulintel For this argument to have any kind of effect, you have to pass it at pipeline initialization here: https://github.com/huggingface/optimum-habana/blob/a51475f477715715f7debd064db6f312bc3bbdf0/examples/stable-diffusion/training/train_controlnet.py#L122

skaulintel:
> @skaulintel For this argument to have any kind of effect, you have to pass it at pipeline initialization here:
> 
> https://github.com/huggingface/optimum-habana/blob/a51475f477715715f7debd064db6f312bc3bbdf0/examples/stable-diffusion/training/train_controlnet.py#L122

unless I am mistaken, that is what I am doing?

skaulintel:
I saw the discussion in [1636](https://github.com/huggingface/optimum-habana/pull/1636
).

examples/stable-diffusion/text_to_image_generation.py
```diff
@@ -441,7 +441,7 @@ def main():
     kwargs_call["quant_mode"] = args.quant_mode
 
     # Instantiate a Stable Diffusion pipeline class
-    import habana_frameworks.torch.core as htcore  # noqa: F401
+    #import habana_frameworks.torch.core as htcore  # noqa: F401
```
regisss, line 444: If we don't need this, we should rather remove it

dsocek: @regisss @skaulintel  We need this for quantization.  If we remove we are getting weird error.   I will investigate if there is some other workaround that can work instead. Notice that I had to add ` # noqa: F401` because ruff will else remove it whe we do `make style`..

regisss: It's fine to have `# noqa:`. I don't think it's an issue to keep the import here, or have commented it because it doesn't work @skaulintel ?

dsocek: We kind of have "catch-22 problem" here I think:
1. If we remove, we will get error with quantization but will resolve issue in this PR
2. If we keep, quantization works, but this PR has issue
..

skaulintel: I will remove this change for now then. @dsocek Could you open a separate PR to address this when you find a fix?

dsocek: @skaulintel  sounds good.  I will work on this PR

libinta (approved):

PR #1639, opened 2024-12-19
Fix for models without rope_scaling config

ugolowic:
Fix AtrtibuteError when accessing config.rope_scaling for models without this attribute.


libinta:
in v1.15.0-release,

regisss:
Already in main as it was cherry-picked from https://github.com/huggingface/optimum-habana/pull/1625

PR #1633, opened 2024-12-19
fix diffuser save_lora_weights crash

sywangyi:
fix crash in train_dreambooth_lora_sdxl.py in save_lora_weights
python ../../gaudi_spawn.py --world_size 8 --use_mpi train_dreambooth_lora_sdxl.py   --pretrained_model_name_or_path="stabilityai/stable-diffusion-xl-base-1.0"    --instance_data_dir="dog"   --pretrained_vae_model_name_or_path="madebyollin/sdxl-vae-fp16-fix"   --output_dir="lora-trained-xl"   --mixed_precision="bf16"   --instance_prompt="a photo of sks dog"   --resolution=1024   --train_batch_size=1   --gradient_accumulation_steps=4   --learning_rate=1e-4   --lr_scheduler="constant"   --lr_warmup_steps=0   --max_train_steps=100   --validation_prompt="A photo of sks dog in a bucket"   --validation_epochs=25   --seed=0   --use_hpu_graphs_for_inference   --use_hpu_graphs_for_training   --gaudi_config_name Habana/stable-diffusion

PR #1626, opened 2024-12-18
Update save lora weights for diffusers with text_encoder_2 layers

skavulya:
# What does this PR do?
Fixes saving lora weights for diffuser pipelines which do not support text_encoder_2 layers.

For example,
```
python examples/trl/ddpo.py \
  --num_epochs=200 \
  --train_gradient_accumulation_steps=1 \
  --sample_num_steps=50 \
  --sample_batch_size=6 \
  --train_batch_size=3 \
  --sample_num_batches_per_epoch=4 \
  --per_prompt_stat_tracking=True \
  --per_prompt_stat_tracking_buffer_size=32 \
  --train_learning_rate=1e-05 \
  --tracker_project_name="stable_diffusion_training" \
  --log_with="tensorboard" \
  --use_habana \
  --use_hpu_graphs \
  --bf16 \
  --hf_hub_model_id="ddpo-finetuned-stable-diffusion" \
  --push_to_hub False
```

 File "/root/tf/optimum-habana-new/examples/trl/ddpo.py", line 245, in <module>
    trainer.train()
  File "/usr/local/lib/python3.10/dist-packages/trl/trainer/ddpo_trainer.py", line 604, in train
    global_step = self.step(epoch, global_step)
  File "/root/tf/optimum-habana-new/optimum/habana/trl/trainer/ddpo_trainer.py", line 315, in step
    self.accelerator.save_state()
  File "/usr/local/lib/python3.10/dist-packages/accelerate/accelerator.py", line 2987, in save_state
    hook(self._models, weights, output_dir)
  File "/usr/local/lib/python3.10/dist-packages/trl/trainer/ddpo_trainer.py", line 439, in _save_model_hook
    self.sd_pipeline.save_checkpoint(models, weights, output_dir)
  File "/usr/local/lib/python3.10/dist-packages/trl/models/modeling_sd_base.py", line 883, in save_checkpoint
    self.sd_pipeline.save_lora_weights(save_directory=output_dir, unet_lora_layers=state_dict)
  File "/root/tf/optimum-habana-new/optimum/habana/diffusers/pipelines/pipeline_utils.py", line 397, in save_lora_weights
    return super().save_lora_weights(
TypeError: StableDiffusionLoraLoaderMixin.save_lora_weights() takes from 2 to 8 positional arguments but 9 were given

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
[...]

PR #1643, opened 2024-12-19
Fix `save_lora_weights` in `pipeline_utils.py`

regisss:
# What does this PR do?

https://github.com/huggingface/optimum-habana/pull/1626 introduced a condition check on whether `text_encoder_2_lora_layers` is None or not and then call `save_lora_weights` in different ways: https://github.com/huggingface/optimum-habana/pull/1626/files#diff-bfc760c4e8acf1425990d609ecd6f1cadb2e027a0d20f027f652375f012e484dR399

However, `text_encoder_2_lora_layers` can be None and still be passed as an argument to `save_lora_weights`, as shown with SDXL in https://github.com/huggingface/optimum-habana/pull/1633#issuecomment-2553393043.

This PR changes the check so that `text_encoder_2_lora_layers` is passed if and only if it is part of the signature of `save_lora_weights`.

## Before submitting

### DIFF
--- a/optimum/habana/diffusers/pipelines/pipeline_utils.py
+++ b/optimum/habana/diffusers/pipelines/pipeline_utils.py
@@ -28,10 +28,10 @@ from diffusers.utils.hub_utils import load_or_create_model_card, populate_model_
 from diffusers.utils.torch_utils import is_compiled_module
 from huggingface_hub import create_repo
 
-from optimum.habana.utils import to_device_dtype
 from optimum.utils import logging
 
 from ...transformers.gaudi_configuration import GaudiConfig
+from ...utils import to_device_dtype
 
 
 logger = logging.get_logger(__name__)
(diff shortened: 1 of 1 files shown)


sywangyi:
no, it will fix the crash but will cause text_encoder_2_lora_layers missing in checkpoint saving. the root cause of the issue is for XL. it should go to https://github.com/huggingface/diffusers/blob/b756ec6e80b3d94c3ae7dc356bdbbdb426a05dca/src/diffusers/loaders/lora_pipeline.py#L1009 instead of https://github.com/huggingface/diffusers/blob/b756ec6e80b3d94c3ae7dc356bdbbdb426a05dca/src/diffusers/loaders/lora_pipeline.py#L466. this is what  https://github.com/huggingface/optimum-habana/pull/1633 is doing. go to StableDiffusionXLPipeline.save_lora_weights directly. super() in this func is StableDiffusionPipeline

sywangyi (approved):

PR #1649, opened 2024-12-20
Fix distributed issue for ST Trainer

ZhengHongming888:
# What does this PR do?

This PR will solve the issue from the running of ST trainer over multiple HPU cards as shown in the readme file as 
https://github.com/huggingface/optimum-habana/tree/main/examples/sentence-transformers-training/nli

the command as
HABANA_VISIBLE_MODULES="2,3" python ../../gaudi_spawn.py --use_deepspeed --world_size 2 training_nli.py bert-base-uncased

Error happened as below

image

Now it is solved and no error with multiple HPU cards for ST trainer and verified. 

image

Thanks. 

Fixes # (issue)

## Before submitting


ZhengHongming888:
> @ZhengHongming888 the PR solves the issue, but you are disabling data track for model card generation. This can be a temp fix for the release but not a solution to the issue!

yes agree. This could be temp fix and will try to fix before next release. thanks.

regisss (approved):

PR #1657, opened 2024-12-20
Quick fix for quantization/custom op list loading

dsocek:
# What does this PR do?

Quick WA fix for issue mentioned in PR #1631


dsocek:
@libinta here is minimal/temp fix we discussed offline

PR #1676, opened 2025-01-03
Enable warmup also for full prompt length case in text generation

yeonsily:
# What does this PR do?
The prompt_length condition check for warm up is a legacy code for some previous demonstration.
We tried to show that even we do not warmup with exact (dynamic) length we still can get some performance.
But this demo case is no longer needed and all this kind of things are handled in vllm now.

Fixes # (issue)

This is a fix for performance degradation of dynamic shape run when datasets given.

For example, 
PT_HPU_LAZY_MODE=0  python3  run_generation.py  --num_beams 1 --model_name_or_path meta-llama/Llama-2-7b-chat-hf  --attn_softmax_bf16  --use_kv_cache --max_new_tokens 128 --bf16 --batch_size 8     --trim_logits --max_input_tokens -1 --n_iterations 500 --warmup 1 --dataset_name tatsu-lab/alpaca   --run_partial_dataset   --torch_compile 

## Before submitting

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -719,22 +719,21 @@ def main():
             return prompt, outputs
 
         # warmup
-        if prompt_length > 0:
-            from optimum.habana.utils import HabanaProfile
+        from optimum.habana.utils import HabanaProfile
 
-            # compilation stage disable profiling
-            HabanaProfile.disable()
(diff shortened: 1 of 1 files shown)


jiminha (approved):
LGTM!

regisss (approved):

PR #1680, opened 2025-01-06
Add checks for parallel_state initialization

yafshar:
# What does this PR do?

- Modified parallel_state initialization to include a check for uninitialized state.
- Added validation to ensure sequence parallel world size matches context parallel size.
- Included a check to ensure FP8 amax reduction group is not already initialized.

## Background 

The https://github.com/huggingface/optimum-habana/pull/1501 PR added support for Context Parallelism and took advantage of model and data parallel groups used in Megatron-LM. In the original (Megatron-LM) implementation the global objects treated with an init function, while the https://github.com/huggingface/optimum-habana/pull/1501 PR included the parallel_state in a class https://github.com/huggingface/optimum-habana/blob/f48dda8204f046fc46d1c09da7e2916cb7444aca/optimum/habana/accelerate/state.py#L31 at https://github.com/huggingface/optimum-habana/blob/f48dda8204f046fc46d1c09da7e2916cb7444aca/optimum/habana/accelerate/state.py#L91

This usage caused issues like what happens in https://github.com/huggingface/optimum-habana/pull/1649 where the object created using `GaudiTrainingArguments` is created more than once.  

Fixes # (issue)

https://github.com/huggingface/optimum-habana/pull/1649

## Before submitting


yafshar:
@bhargaveede  can you review this PR?

optimum/habana/accelerate/state.py
```diff
@@ -88,7 +88,18 @@ def __init__(self, cpu: bool = False, **kwargs):
...
+                    parallel_state.initialize_model_parallel(
+                        sequence_parallel_size=context_parallel_size, use_fp8=False
+                    )
+                else:
+                    if parallel_state.get_sequence_parallel_world_size() != context_parallel_size:
+                        raise ValueError(
+                            "The initialized sequence parallel world size does not match the context parallel size."
+                        )
+                    # Ensure that the parallel_state is initialized similarly with use_fp8=False
+                    if parallel_state._AMAX_REDUCTION_GROUP is not None:
```
bhargaveede, line 101: if parallel_state._AMAX_REDUCTION_GROUP is not None:

This is just to check fp8 is set as False, am I right? Rest all seems fine.

yafshar: Yes, since this constructor is using `use_fp8=False` this is extra check to ensure it is consistent.

yafshar: Do you think this check is unnecessary? I should add a get function in parallel_state if we use this

yafshar: @bhargaveede I changed this check to info and added a function in parallel_state. If you have time please check the changes.

yafshar:
@libinta would you please check this PR? (add a run_label) It is necessary for PEFT + sentence transformers

regisss:
@yafshar LGTM! Can you also merge the latest main branch into your branch? I'll trigger the CI after that :slightly_smiling_face:

yafshar:
@regisss please trigger the CI .

regisss (approved):

PR #1689, opened 2025-01-09
Added Mamba model using kernel to improve the performance

zzhang37:
# What does this PR do?
This PR is for mamba using kernel.

Fixes # (issue)

## Before submitting


zzhang37:
@regisss The unit test failure is due to the docker mapped the folder to default huggingface cache, like 
-v /checkpoint/huggingface/hub:/root/.cache/huggingface/hub. Without that, the unit test should be able to pass.

zzhang37:
@astachowiczhabana The mamba TPC kernel is merged to standard TPC kernel library, so the mamba code can be simplified because of that, so this old implementation is no longer needed. We can close it.

PR #1690, opened 2025-01-10
Add gated models to diffusers CI tests

dsocek:
# What does this PR do?

Adding support for running gated models in diffusers CI

### DIFF
--- a/.github/workflows/slow_tests_gaudi2.yml
+++ b/.github/workflows/slow_tests_gaudi2.yml
@@ -60,7 +60,7 @@ jobs:
             --net=host \
             --ipc=host \
             vault.habana.ai/gaudi-docker/1.19.0/ubuntu22.04/habanalabs/pytorch-installer-2.5.1:latest \
-            /bin/bash tests/ci/slow_tests_diffusers.sh
+            /bin/bash tests/ci/slow_tests_diffusers.sh ${{ secrets.TEXT_GENERATION_CI_HUB_TOKEN }}
   deepspeed:
     name: Test DeepSpeed models
     if: ${{ !cancelled() && (success() || failure()) }}
--- a/tests/ci/slow_tests_diffusers.sh
+++ b/tests/ci/slow_tests_diffusers.sh
@@ -2,6 +2,7 @@
 
 python -m pip install --upgrade pip
 export RUN_SLOW=true
+huggingface-cli login --token $1
 make test_installs
 CUSTOM_BF16_OPS=1 python -m pytest tests/test_diffusers.py -v -s -k "test_no_throughput_regression_autocast"
 make slow_tests_diffusers
--- a/tests/test_diffusers.py
+++ b/tests/test_diffusers.py
@@ -33,6 +33,7 @@ from unittest import TestCase, skipUnless
 import diffusers
 import habana_frameworks.torch.hpu as hthpu
 import numpy as np
+import PIL
 import pytest
 import requests
 import safetensors
@@ -69,7 +70,8 @@ from diffusers.utils.testing_utils import (
     require_torch,
 )
 from diffusers.utils.torch_utils import randn_tensor
-from huggingface_hub import snapshot_download
+from huggingface_hub import HfApi, hf_hub_download, snapshot_download
+from huggingface_hub.utils import HfHubHTTPError
 from parameterized import parameterized
 from PIL import Image
 from transformers import (
@@ -140,7 +142,9 @@ if IS_GAUDI2:
     THROUGHPUT_UNCONDITIONAL_IMAGE_BASELINE_BF16 = 0.145
     SDXL_THROUGHPUT = 0.301
     SVD_THROUGHPUT = 0.012
+    SD3_THROUGHPUT = 0.006
     FLUX_THROUGHPUT = 0.03
+    FLUX_DEV_I2I_THROUGHPUT = 0.12
 else:
     THROUGHPUT_BASELINE_BF16 = 0.275
     THROUGHPUT_BASELINE_AUTOCAST = 0.114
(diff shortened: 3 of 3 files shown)


tests/ci/slow_tests_diffusers.sh
```diff
@@ -2,6 +2,7 @@
 
 python -m pip install --upgrade pip
 export RUN_SLOW=true
+huggingface-cli login --token $1
```
uartie, line 5: There are other places that call this script (e.g. Makefile).  We should not expect parameters to be passed to this script.  Instead, we should just assume the HF_TOKEN has been configured at a higher level (i.e. by the user/caller/ci).

.github/workflows/slow_tests_gaudi2.yml
```diff
@@ -60,7 +60,7 @@ jobs:
             --net=host \
             --ipc=host \
             vault.habana.ai/gaudi-docker/1.19.0/ubuntu22.04/habanalabs/pytorch-installer-2.5.1:latest \
-            /bin/bash tests/ci/slow_tests_diffusers.sh
+            /bin/bash tests/ci/slow_tests_diffusers.sh ${{ secrets.TEXT_GENERATION_CI_HUB_TOKEN }}
```
uartie, line 63: I wonder if the runner already has a token configured in it's default environment.  @regisss?

How about set it on the container environment instead of passing it to the script (i.e. `-e HF_TOKEN=${{secrets.TEXT_GENERATION_CI_HUB_TOKEN}}`?

uartie (changes requested):

.github/workflows/slow_tests_gaudi2.yml
dsocek: @uartie Thanks Artie.  There are few other places which use this approach to pass token, e.g. see [here](https://github.com/huggingface/optimum-habana/blob/main/tests/ci/slow_tests_deepspeed.sh).  We should agree on the best method for all, and update all cases.  @regisss WDYT?  (also someone with admin privileges should add 2 gated models to token from secrets)

tests/ci/slow_tests_diffusers.sh
uartie: oops, sorry.  I got it backwards.  The script calls the Makefile.

uartie (commented):
k

tests/test_diffusers.py
```diff
@@ -20,6 +20,7 @@
 import inspect
 import json
 import os
+import PIL
```
Luca-Calabria, line 23: @dsocek should you add PIL (and your tested version) in the examples/stable-diffusion/requirements.txt ?

Luca-Calabria (changes requested):

tests/test_diffusers.py
```diff
@@ -1626,6 +1648,32 @@ def test_fused_qkv_projections(self):
             original_image_slice, image_slice_disabled, atol=1e-2, rtol=1e-2
         ), "Original outputs should match when fused QKV projections are disabled."
 
+    @slow
+    @check_gated_model_access("stabilityai/stable-diffusion-3-medium-diffusers")
+    @pytest.mark.skipif(not IS_GAUDI2, reason="does not fit into Gaudi1 memory")
+    def test_sd3_inference(self):
```
Luca-Calabria, line 1654: "Makefile" runs some slow tests explicitly here https://github.com/huggingface/optimum-habana/blob/main/Makefile#L101 . Do you think to do the same for this new test case?

tests/test_diffusers.py
```diff
@@ -6681,3 +6729,38 @@ def test_flux_prompt_embeds(self):
 
         max_diff = np.abs(output_with_prompt - output_with_embeds).max()
         assert max_diff < 1e-4
+
+    @slow
+    @check_gated_model_access("black-forest-labs/FLUX.1-dev")
+    @pytest.mark.skipif(not IS_GAUDI2, reason="does not fit into Gaudi1 memory")
+    def test_flux_img2img_inference(self):
```
Luca-Calabria, line 6736: "Makefile" runs some slow tests explicitly here https://github.com/huggingface/optimum-habana/blob/main/Makefile#L101 . Do you think to do the same for this new test case?

tests/test_diffusers.py
dsocek: Yes for sure.  PIL is definitely needed for image-to-image case but somehow was not added to requirements.txt.   Will add in amend.

tests/test_diffusers.py
dsocek: Yes but not in this PR.  We have SD3 improved throughput (right now is slow_ PR pending, so once thats merged, we will update this.  Also there is comprehensive CI update PR also being in review, so this section in Makefile will be properly addressed then.  Ideally it should cover running all regression tests.  We also need to update the regression thresholds with G3 values.  Thus not covered yet.

tests/test_diffusers.py
dsocek: See my comment above.

tests/test_diffusers.py
Luca-Calabria: Ok, thank you for the explanation!

tests/test_diffusers.py
dsocek: @Luca-Calabria actually I just checked, Pillow is installed automatically as a pre-requisite when you install `optimum-habana`, so I don't think we need to put it in explicit list of additional requirements via `requirements.txt`

Luca-Calabria (approved):

Luca-Calabria:
LGTM!

Luca-Calabria:
@regisss we are ready to merge it

.github/workflows/slow_tests_gaudi2.yml
regisss: This token is a secret defined at the level of the optimum-habana repository. So all workflows that are allowed to read tokens will have access to it. I'm the admin of the optimum-habana repo so I have full control over it. It's absolutely fine to use here the same token as for the text-generation tests IMO.
@dsocek What are the gated models you would like to test?

dsocek:
> LGTM! @dsocek The two gated models the token should give access to are `stabilityai/stable-diffusion-3-medium-diffusers` and `black-forest-labs/FLUX.1-dev` right?

Yes, for now these are only 2 gated models we have in diffusers tests :)

regisss:
Got it, the token should provide access to these models from now on

PR #1692, opened 2025-01-10
Lm eval upgraded to 0.4.7

12010486:
# Upgraded run_lm_eval.py

- [x] Run on top of the latest 0.4.7 of https://github.com/EleutherAI/lm-evaluation-harness.
- [x] Accuracy and throughput aligned with older version. 
- [ ] The parser might need new args to expand functionalities. Can handle in subsequent PRs

## Results

From [Examples](https://github.com/12010486/optimum-habana/tree/lm_eval_to_0.4.7/examples/text-generation#examples)

1. Evaluate Llama 7B on Gaudi on task PiQA, using the BF16 data type:

```
python run_lm_eval.py \
--model_name_or_path meta-llama/Llama-2-7b-hf \
--use_hpu_graphs \
--use_kv_cache \
--bf16 \
--batch_size=1 \
--tasks piqa \
-o eval.json
```

**Previous results:**

```
"results": {
    "piqa": {
      "acc": 0.780739934711643,
      "acc_stderr": 0.009653357463605322,
      "acc_norm": 0.7861806311207835,
      "acc_norm_stderr": 0.009565994206915602
    }
```
**This PR**

 ```
"results": {
    "piqa": {
      "alias": "piqa",
      "acc,none": 0.780739934711643,
      "acc_stderr,none": 0.00965335746360534,
      "acc_norm,none": 0.7910772578890098,
      "acc_norm_stderr,none": 0.009485227030105051
    }
```

2. Evaluate Llama 70B on 8 Gaudi2 cards on task WinoGrande, using the BF16 data type:

```
deepspeed --num_gpus 8 run_lm_eval.py \
--model_name_or_path meta-llama/Llama-2-70b-hf \
--use_hpu_graphs \
--use_kv_cache \
--bf16 \
--batch_size=1 \
--tasks winogrande \
-o eval.json
```

 **Previous results:**
```
"results": {
    "winogrande": {
      "acc": 0.8034727703235991,
      "acc_stderr": 0.011168120593569576
    }
```
**This PR**

 ```
"results": {
    "winogrande": {
      "alias": "winogrande",
      "acc,none": 0.7774269928966061,
      "acc_stderr,none": 0.011690933809712669
    }
```


examples/text-generation/run_lm_eval.py
```diff
@@ -14,32 +14,33 @@
 # limitations under the License.
 
 ###############################################################################
-# Copyright (C) 2020-2021 Habana Labs, Ltd. an Intel Company
+# Copyright (C) 2020-2024 Habana Labs, Ltd. an Intel Company
```
yafshar, line 17:
```suggestion
# Copyright (C) 2020-2025 Habana Labs, Ltd. an Intel Company
```

examples/text-generation/run_lm_eval.py
```diff
@@ -14,32 +14,33 @@
...
 import argparse
 import json
-import logging
 import multiprocessing as mp
 import os
 import time
 
-import lm_eval.evaluator
-import lm_eval.tasks
+from lm_eval import evaluator, tasks, utils
```
yafshar, line 26:
```suggestion
from lm_eval import evaluator, utils
```

I do not see tasks is used anywhere!

examples/text-generation/run_lm_eval.py
```diff
@@ -14,32 +14,33 @@
...
+from transformers.generation import GenerationConfig
 from utils import finalize_quantization, initialize_model
 
 from optimum.habana.utils import get_hpu_memory_stats
 
-
 os.environ.setdefault("TOKENIZERS_PARALLELISM", "false")
-logger = logging.getLogger(__name__)
-
+#logger = logging.getLogger(__name__)
```
yafshar, line 42:
```suggestion
```

Please remove the comment.

examples/text-generation/run_lm_eval.py
```diff
@@ -86,22 +87,35 @@ def setup_lm_eval_parser():
...
-        super().__init__()
+class HabanaModelAdapter(HFLM):
+    def __init__(
+        self,
+        tokenizer: AutoTokenizer,
+        model: AutoModelForCausalLM,
+        args: argparse.Namespace,
+        options: GenerationConfig,
+    ) -> None:
+        super().__init__(device=args.device, pretrained=args.model_name_or_path)
```
yafshar, line 110:
```suggestion
        super().__init__(pretrained=args.model_name_or_path, device=args.device)
```

examples/text-generation/run_lm_eval.py
```diff
@@ -135,82 +149,93 @@ def __init__(self, tokenizer, model, args, options):
...
 
-    lm_tasks = lm_eval.tasks.get_task_dict(args.tasks)
+    #lm_tasks = lm_eval.tasks.get_task_dict(args.tasks)
     with torch.no_grad():
         lm = HabanaModelAdapter(tokenizer, model, args, generation_config)
 
     eval_start = time.perf_counter()
     with torch.no_grad():
-        results = lm_eval.evaluator.evaluate(lm, lm_tasks, limit=args.limit_iters)
+        #results = evaluator.evaluate(lm, lm_tasks, limit=args.limit_iters)
```
yafshar, line 237:
```suggestion
```

examples/text-generation/run_lm_eval.py
```diff
@@ -135,82 +149,93 @@ def __init__(self, tokenizer, model, args, options):
...
     if args.trust_remote_code:
         # trust_remote_code fix was introduced in lm_eval 0.4.3
-        # https://github.com/EleutherAI/lm-evaluation-harness/pull/1998/files
-        # We need to cherry-pick the fix manually untill we upgrade (SW-190418)
         import datasets
 
         datasets.config.HF_DATASETS_TRUST_REMOTE_CODE = True
 
-    lm_tasks = lm_eval.tasks.get_task_dict(args.tasks)
+    #lm_tasks = lm_eval.tasks.get_task_dict(args.tasks)
```
yafshar, line 231:
```suggestion
```

examples/text-generation/run_lm_eval.py
12010486: You are right, ``make style`` already fixed it

examples/text-generation/run_lm_eval.py
```diff
@@ -86,22 +87,35 @@ def setup_lm_eval_parser():
...
-        super().__init__()
+class HabanaModelAdapter(HFLM):
+    def __init__(
+        self,
+        tokenizer: AutoTokenizer,
+        model: AutoModelForCausalLM,
+        args: argparse.Namespace,
+        options: GenerationConfig,
+    ) -> None:
+        super().__init__(pretrained=args.model_name_or_path, device=args.device)
```
alexey-belyakov, line 110: Here we initialize the base class with the pretrained model as a string, and HFLM instantiates this model. After that we just override this instance:
self._model = model
To me, it looks like overhead.

examples/text-generation/run_lm_eval.py
```diff
@@ -135,82 +149,93 @@ def __init__(self, tokenizer, model, args, options):
...
-        return self.tokenizer.encode(string)
-
-    def tok_decode(self, tokens):
-        return self.tokenizer.decode(tokens)
-
-    def _model_generate(self, context, max_length, eos_token_id):
-        raise NotImplementedError()
-
-    def find_bucket(self, length):
+    def find_bucket(self, length: int) -> list[int]:
```
alexey-belyakov, line 171:
```suggestion
    def find_bucket(self, length: int) -> int:
```

examples/text-generation/run_lm_eval.py
```diff
@@ -225,8 +250,13 @@ def main():
             mem = get_hpu_memory_stats()
             for k, v in mem.items():
                 print("{:35} = {} GB".format(k[:-5].replace("_", " ").capitalize(), v))
-        json.dump(results, open(args.output_file, "w"), indent=2)
-        print(json.dumps(results, indent=2))
+
+        json.dump(
```
alexey-belyakov, line 254:
```suggestion
json_str = json.dumps(results, indent=2, default=utils.handle_non_serializable, ensure_ascii=False)

with open(args.output_file, "w") as output_file:
    output_file.write(json_str)

if args.show_config:
    print(json_str)
```

examples/text-generation/run_lm_eval.py
yafshar: I agree, sounds like parent constructor `HFLM` executes some extra code which are not necessary for habana. Maybe instead of the full parent `class HFLM(TemplateLM)` only use `TemplateLM`

```sh
from lm_eval.models.huggingface TemplateLM
```

and instead of 
```sh
super().__init__(pretrained=args.model_name_or_path, device=args.device)
``
```sh
TemplateLM.__init__(self)
```

examples/text-generation/run_lm_eval.py
```diff
@@ -86,22 +87,35 @@ def setup_lm_eval_parser():
         default=["hellaswag", "lambada_openai", "piqa", "winogrande"],
     )
     parser.add_argument("--limit_iters", type=int, help="limit examples to run that many iterations", default=None)
+    parser.add_argument(
```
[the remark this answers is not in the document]
12010486: I believe we should look at simple_evaluator, rather than evaluator, but in any case, in the latest version of the testharness, there are really multiple args related to what to print and how, see from this line below https://github.com/EleutherAI/lm-evaluation-harness/blob/4c26a9c176ecfb40b369503ce211e356bb800fdc/lm_eval/__main__.py/#L410 

I took only one of them and tried to use it the same way it is expected from the original repo, if I did it right. The true reason why I've added it, is that is printing now really a lot of info, and I was not able to see the information related to HPU without this flag.

examples/text-generation/run_lm_eval.py
12010486: So, two options here, based on your input. 
1. Instead of a string (`args.model_name_or_path`), pass the `model ` in the __init__ like: 
`super().__init__(pretrained=model, device=args.device)`
This will automatically exclude all the `cuda` code, and it is working, both on 1 HPU and 8 HPU. Downside: warnings during the initialization, like [this](https://github.com/EleutherAI/lm-evaluation-harness/blob/v0.4.7/lm_eval/models/huggingface.py/#L99)
2. Change class from HFLM to TemplateLM. This might be lengthier with the re-definition of multiple functions, but I'm working also on this.

@yafshar, @alexey-belyakov, I appreciate your input on it. 
With both options, I can prune the code that might be now redundant

12010486: Hi @alexey-belyakov, @yafshar, I've decided to opt for option 2.b: inheriting from HFLM and redefining the whole init function except what is already set via  TemplateLM, see the code. Though I did not like this solution before, it is not malpractice to inherit from grandparents and allows to get rid of warnings and to have in general more control of what is going on

12010486: (and to avoid to redefine functions like _loglikelihood_tokens, generate_until, loglikelihood_rolling and tok_encode)

examples/text-generation/run_lm_eval.py
12010486: Why int and not list of int?

examples/text-generation/run_lm_eval.py
12010486: Redefined based on your observation, thanks!

yafshar (commented):
@12010486 thanks for the update. It sounds good to me in the current form. I am doing some more tests and finish my review soon

yafshar (approved):
LGTM!

Hi @regisss, this PR is ready for your final review. Could you please take a look?

regisss (approved):

PR #1713, opened 2025-01-23
[SW-214100] Enable attention selection in wav2vec-ac

ugolowic:
It was impossible for the user to select eager attention in wav2vec2 audio classification. It always defaulted to sdpa. Now default is eager and fused sdpa is selected with --use_flash_attention option.


regisss (approved):
LGTM!

PR #1717, opened 2025-01-24
update audio-classification/requirements.txt to fix numpy version con…

hsubramony:
numpy version conflicts fix for audio-classification

librosa installs numba latest which got upgraded recently and causes latest numpy to be installed. This causes a break in our runs in both released 1.19 versions and newer versions.

this change specificies the workign version of numba so that it doesnt upgrade numpy.

### DIFF
--- a/examples/audio-classification/requirements.txt
+++ b/examples/audio-classification/requirements.txt
@@ -1,3 +1,4 @@
-numpy==1.24.0
+datasets>=1.14.0
 evaluate
+numba==0.60.0
 librosa
--- a/examples/audio-classification/requirements.txt
+++ b/examples/audio-classification/requirements.txt
@@ -1,3 +1,3 @@
-datasets>=1.14.0
+numpy==1.24.0
 evaluate
 librosa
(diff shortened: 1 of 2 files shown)


jiminha (approved):

jiminha:
@regisss  could you review this and merge this? Without this, we are seeing failure even on v1.15.0 baseline.  We will need to merge both main and v.1.15.0

jiminha:
Without this change you will see error like below .

https://github.com/huggingface/optimum-habana/tree/v1.15.0/examples/audio-classification#single-hpu

  File "/usr/local/lib/python3.10/dist-packages/pandas/_libs/__init__.py", line 13, in <module>
    from pandas._libs.interval import Interval
  File "pandas/_libs/interval.pyx", line 1, in init pandas._libs.interval
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject

jiminha:
@libinta  @regisss  Please review this, without this we will see error on both main and 1.15.0.

regisss (approved):

PR #1726, opened 2025-01-27
Removing HL_DS_DISTRIBUTED_ATTENTION_SEQ_DIM as it's not needed from 1.20

bhargaveede:
1.20 Release

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/language-modeling/README.md
+++ b/examples/language-modeling/README.md
@@ -935,7 +935,6 @@ We have added support for [Deepspeed Ulysses](https://github.com/microsoft/DeepS
 > This feature is still in beta version and may not work out of the box for all transformer model architectures and configurations.
 
 ```bash
-HL_DS_DISTRIBUTED_ATTENTION_SEQ_DIM=1   \
 python3 ../gaudi_spawn.py  \
         --world_size 8  --use_deepspeed run_lora_clm.py \
         --model_name_or_path meta-llama/Llama-3.1-8B \


yafshar:
@bhargaveede thanks. The workaround has already been removed in the next DeepSpeed update, so it is completely safe to remove it now for the 1.20 release.

yafshar (approved):
LGTM!

regisss:
@bhargaveede This is still needed for SynapseAI 1.20 right?

yafshar:
> @bhargaveede This is still needed for SynapseAI 1.20 right?

The workaround in DeepSpeed for SynapseAI 1.19.0 included the `HL_DS_DISTRIBUTED_ATTENTION_SEQ_DIM` environment variable. However, this implementation has been corrected, and the variable has been removed in the subsequent DeepSpeed release for SynapseAI 1.20. So the variable has no effect in 1.20

regisss:
> > @bhargaveede This is still needed for SynapseAI 1.20 right?
> 
> The workaround in DeepSpeed for SynapseAI 1.19.0 included the `HL_DS_DISTRIBUTED_ATTENTION_SEQ_DIM` environment variable. However, this implementation has been corrected, and the variable has been removed in the subsequent DeepSpeed release for SynapseAI 1.20. So the variable has no effect in 1.20

Thanks for the explanation! I meant "still needed for 1.19" actually, which is the case if I understand correctly. I'll probably merge this PR in the release branch first since I prefer keeping the main branch compatible with the latest Synapse release.

regisss (approved):

PR #1729, opened 2025-01-27
Extend lm_eval functionality

12010486:
# What does this PR do?

Extended `run_lm_eval.py` to support latest tasks, like MMLU, MMLU-Pro, IFEval and GSM8K. 

## Before submitting


12010486:
@yafshar this is part II

examples/text-generation/run_lm_eval.py
```diff
@@ -73,27 +75,67 @@ def setup_lm_eval_parser():
...
+        default=False,
+        help="If True, write out all model outputs and documents for per-sample measurement and post-hoc analysis. Use with --output_path.",
+    )
+    parser.add_argument(
+        "--system_instruction",
+        type=str,
+        default=None,
+        help="System instruction to be used in the prompt",
+    )
+
```
yafshar, line 129:
```suggestion
```

yafshar:
Can you add examples that you tried to the PR README?

12010486:
Related to the examples to add, @yafshar, I've added llama-3.2-1B. While for this model (w/o Instruct) the accuracy we get is the same as reported in https://huggingface.co/meta-llama/Llama-3.2-1B-Instruct (i.e., 32.2), I've noticed that for other llama models it is sometimes around 2-3 point less, so I'm checking if everything is correct. As soon as done, I will add also an example for ifeval. I believe we should remove older examples, rather than simply add another couple. What is your thought?

yafshar:
@12010486 please let me know when the PR ready for the review! Thanks

12010486:
@yafshar ack, I will ping you. In the background I'm adding needed or useful parameters, but I'm blocked on ifeval task, as accuracies on that task are still not replicable by our code.

examples/text-generation/run_lm_eval.py
```diff
@@ -124,8 +180,15 @@ def __init__(
...
+        self.truncation = truncation
         self.logits_cache = logits_cache
         self.add_bos_token = add_bos_token
+        if "gemma" in getattr(self._config, "model_type", ""):
+            self.add_bos_token = True
+            logger.info(
+                f"Model type is '{self._config.model_type}', part of the Gemma family--a BOS token will be used as Gemma underperforms without it."
+            )
+
         self._max_length = options.max_length
```
yafshar, line 192: @12010486 is this correct? You set this before in previous PR, so the super class `max_length` is using this, but the current derived class has a different max_length
```sh
    @property
    def max_length(self) -> int:
        return self.buckets[-1]
```

veritas9872:
Hello! Thank you for the progress on this update!
Some information that may be helpful.
First, the scores on the Llama website are different from the scores obtained by LM Eval harness because Meta uses its internal code for measuring metrics. I have also found that scores drop relative to the ones reported by Meta.
A good way might be to compare scores obtained when using a vLLM API with the scores obtained by HuggingFace models.
Also, some variation in scores is inevitable. I have found that results from A100 and H100 devices are different as well.

veritas9872:
One minor comment concerning GSM8K.
The `gsm8k_cot_llama` should be used to reproduce the Llama results instead of `gsm8k_cot`. This is because Meta uses a slightly different method for GSM8K.

examples/text-generation/run_lm_eval.py
12010486: Good catch @yafshar. We need the property to be defined that way, or else max_lengh is going to change every time causing OOM. I can remove line 192, though

12010486:
@veritas9872, very useful comments! In fact, using `gsm8k_cot_llama` really boosted the exact match accuracy! As for the differences in score, I'm fine with tiny variations (as it is hard to control the whole randomness, esp if you change batch size and number of cards) but still it seems to be a bit too much in our case, so I'm double checking step by step

yafshar:
@12010486 what is the status? Is this PR ready for final review?

veritas9872:
I am also curious as to whether this can be merged in time before the next release.

yafshar:
> > I am also curious as to whether this can be merged in time before the next release.
> 
> Not before 1.16 is out, but shortly after

@12010486  then can you please change this PR to draft mode for now

PR #1733, opened 2025-01-30
Fix formatting of the kubeVersion range in Kubernetes helm chart

dmsuehir:
# What does this PR do?

This PR fixes the formatting of the `kubeVersion` in the Kubernetes example helm chart. Without this change, the helm chart fails to install when using Kubernetes v1.28:
```
Error: INSTALLATION FAILED: chart requires kubeVersion: 1.27-1.29 which is incompatible with Kubernetes v1.28.7
```
Formatting the range with spaces around the `-` has fixed the issue.

## Before submitting
There are no automated tests for this, but I manually tested with Kubernetes v1.28.7 using the single card glue fine tuning example.


regisss (approved):
LGTM!

PR #1736, opened 2025-01-30
Added sliding window feature to Giudi Gemma2 model

slokesha:
Enabled sliding window feature to gemma2 model.
Gemma2 uses sliding window in ever other layer of Gemma2DecoderLayer. The implementation is [here](https://github.com/huggingface/transformers/blob/8b7024daf5ee4a2d40ef55b576567ceb211212f4/src/transformers/models/gemma2/modeling_gemma2.py#L534) . The same logic in GaudiGemma2DecoderLayer.

The value for sliding_window is passed through [config file](https://github.com/huggingface/transformers/blob/8b7024daf5ee4a2d40ef55b576567ceb211212f4/src/transformers/models/gemma2/configuration_gemma2.py#L125).

Gemma2 can use flash_attention_2 as [attention class](https://github.com/huggingface/transformers/blob/v4.38.2/src/transformers/models/gemma/modeling_gemma.py#L579C6-L579C23). Hence enabling flash_attention_2  as a cmdline argument.

## Before submitting

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -275,6 +275,11 @@ def setup_parser(parser):
         help="Whether to enable Habana Flash Attention in fast softmax mode.",
     )
     parser.add_argument(
+        "--flash_attention_2",
+        action="store_true",
+        help="Whether to enable Flash Attention-2, provided that the model supports it.",
+    )
+    parser.add_argument(
         "--book_source",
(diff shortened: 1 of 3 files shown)


jiminha:
Do you see perf improvement with this sliding window enabled?

12010486:
Closing it until further notice

PR #1737, opened 2025-01-31
Fix the incorrect output of sdxl inpaint

yuanwu2017:
# What does this PR do?
Fix the incorrect output of sdxl inpaint.
Command:
`python text_to_image_generation.py     --model_name_or_path  diffusers/stable-diffusion-xl-1.0-inpainting-0.1     --base_image https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/inpaint.png     --mask_image https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/inpaint_mask.png     --prompts "concept art digital painting of an elven castle, inspired by lord of the rings, highly detailed, 8k"     --seed 42     --scheduler euler_discrete     --num_images_per_prompt 12     --batch_size 4     --image_save_dir ./xl2     --use_habana     --use_hpu_graphs     --gaudi_config Habana/stable-diffusion     --sdp_on_bf16     --bf16`

**Before:**

image
Images(only attach first four images ):

image_1

image_2

image_3

image_4

**After：**
image
Images(only attach first four images ):
image_1
image_2
image_3
image_4

**Seeing from the results, the performance has no decline. The output pictures are more fine**

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_inpaint.py
+++ b/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_inpaint.py
@@ -748,8 +748,6 @@ class GaudiStableDiffusionXLInpaintPipeline(GaudiDiffusionPipeline, StableDiffus
                 ).to(device=device, dtype=latents.dtype)
 
             self._num_timesteps = len(timesteps)
-            if hasattr(self.scheduler, "set_begin_index"):
-                self.scheduler.set_begin_index()
 
             outputs = {
                 "images": [],
--- a/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
+++ b/optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
@@ -657,8 +657,6 @@ class GaudiStableDiffusionXLPipeline(GaudiDiffusionPipeline, StableDiffusionXLPi
             t1 = t0
 
             self._num_timesteps = len(timesteps)
-            if hasattr(self.scheduler, "set_begin_index"):
-                self.scheduler.set_begin_index()
 
             hb_profiler = HabanaProfile(
                 warmup=profiling_warmup_steps,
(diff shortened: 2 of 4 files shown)


yuanwu2017:
@regisss @dsocek  Please help to review.

yuanwu2017:
pass ci test.
image

optimum/habana/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_inpaint.py
```diff
@@ -748,8 +748,6 @@ def denoising_value_valid(dnv):
                 ).to(device=device, dtype=latents.dtype)
 
             self._num_timesteps = len(timesteps)
-            if hasattr(self.scheduler, "set_begin_index"):
-                self.scheduler.set_begin_index()
```
regisss, line 752: @dsocek I don't remember why we added these two lines, do you?

dsocek: @regisss yes I remember.  We added support to define different number of batches to run inference with (good feature for properly assessing performance after warmup) and also we let users pick different schedulers.  To make sure we properly "reset" the selected scheduler after each batch is processed we use `init_step_index` built-in function (in the past we modified the scheduler ourselves but later find this to be supported by diffusers anyway).   However, the `init_step_index` function only works properly if we explicitly set the initial scheduler index value with  
`set_begin_index`.  Hence, we added `set_begin_index` which worked in all test cases as expected.

dsocek:
@yuanwu2017   did you test other schedulers? (e.g. default)?  Also, did you try different seeds, prompts to really see that scheduler visually impairs images if we call `set_begin_index` vs. if we do not

dsocek (changes requested):
Please run more tests and show proof that we can safely remove this.   For default scheduler removing this may cause issues.  Also, can we do a larger sweep of prompts/seeds to justify removal

dsocek:
@regisss Checked old jira info.  We were getting error when using `init_step_index` without setting `set_begin_index` first (for `ddim` scheduler).  This was around SW stack 1.16/1.17 and SDXL.   I just did some tests on my side with 1.19 and seems this issue is not present anymore.   @yuanwu2017  can you also test this?   If this is the case, we can remove the `set_begin_index` if it improves visual quality in inpainting

dsocek:
Update:  @yuanwu2017  I tested 1.19 with all schedulers and the issue we saw in older releases is not showing.   I also did observe better details in generated images, so I think it is safe to remove `set_begin_index` code.   @yuanwu2017  could  you also remove from other 3 SDXL pipelines in this PR?  Thanks!

yuanwu2017:
> Update: @yuanwu2017 I tested 1.19 with all schedulers and the issue we saw in older releases is not showing. I also did observe better details in generated images, so I think it is safe to remove `set_begin_index` code. @yuanwu2017 could you also remove from other 3 SDXL pipelines in this PR? Thanks!

Ok. I will remove. Thanks.

dsocek (approved):

PR #1739, opened 2025-02-03
edit mixtral quantization config file (#114)

dudilester:


libinta (approved):

libinta:
@regisss  let's hold this until I get accuracy from team

dudilester:
@libinta , @regisss this PR corrects the quant_config context, as the old version has invalid regular-expression strings that are not supported, as well the use of both allow-list and block-list which is invalid. the new version quantize the exact same layers as the old file did, just using correct syntax and logic, so practically these 2 files should result with the same model quantization from user perspective. Regarding the QA testing of the new config file, it was done with the new fp8 dynamic moe code which is not yet upstreamed but should be in this release (1.20). 
However even if the new fp8 dynamic moe changes are not upstreamed and we remain with old static moe code this config file should be upstreamed as the old is incorrect.

regisss (approved):
LGTM!

PR #1740, opened 2025-02-04
Resolve 'NoneType' object has no attribute 'gate_proj' err when applying EP in DeepSeek-V2

IT-Forrest:
Resolve the error of no gate_proj attribute during the Expert Parallelism of DeepSeek-V2

# What does this PR do?

Fixes # (issue)

Currently, EP doesn't work in the latest OH main. It fails with the following error
```
[rank3]:     self.experts[idx * self.expert_chunk + i].gate_proj.weight.squeeze() for i in experts_range
[rank3]: AttributeError: 'NoneType' object has no attribute 'gate_proj'
```
This PR will resolve the Error by skipping the NonType object

## Before submitting


regisss (approved):
LGTM!

PR #1742, opened 2025-02-04
Update README.md - correction in diffusers example

ramyij:
in the Diffusers example, the ouputs= called "generator" function which is not defined in the example. Replacing with the "pipeline" object works

# What does this PR do?

## Before submitting


regisss (approved):
LGTM!

PR #1744, opened 2025-02-05
[SW_212175] FLAN-T5 has bad performance when using regional compilation

chaojun-zhang:
# What does this PR do?
FLAN-T5 doesn't work with regional compilation using module.compile.

image

Fixes # (issue)

## Before submitting


regisss (approved):
LGTM!

PR #1745, opened 2025-02-05
Enable Llama 3.1 405B in FP8 (#124)

jaygala223:
* add changes for fix

* add keep_moduleon_host, modify quant json

* remove buffer check

* add llama 405 checks

* remove hardcoded path, reuse module on host check

* fix: undefined variable

* remove unused import

---------

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -429,8 +429,13 @@ def setup_distributed_model(args, model_dtype, model_kwargs, logger):
 
     logger.info("DeepSpeed is enabled.")
     deepspeed.init_distributed(dist_backend="hccl")
-    config = AutoConfig.from_pretrained(args.model_name_or_path, torch_dtype=model_dtype, **model_kwargs)
-    load_to_meta = model_on_meta(config)
+    config = AutoConfig.from_pretrained(args.model_name_or_path, **model_kwargs)
+
+    keep_module_on_host = False
+    if "Llama-3.1-405B" in args.model_name_or_path:
+        keep_module_on_host = True
+
+    load_to_meta = False if keep_module_on_host else model_on_meta(config)
 
     if args.assistant_model is None:
         assistant_model = None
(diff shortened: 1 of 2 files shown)


jiminha (approved):

examples/text-generation/utils.py
```diff
@@ -429,8 +429,13 @@ def setup_distributed_model(args, model_dtype, model_kwargs, logger):
 
     logger.info("DeepSpeed is enabled.")
     deepspeed.init_distributed(dist_backend="hccl")
-    config = AutoConfig.from_pretrained(args.model_name_or_path, torch_dtype=model_dtype, **model_kwargs)
```
regisss, line 432: We cannot remove `torch_dtype=model_dtype` here and then use `config.torch_dtype` afterwards. Because we won't be able to specify anymore if we want to run the model in fp32 or in bf16.
Why do you need to do that?

examples/text-generation/utils.py
```diff
@@ -439,7 +444,10 @@ def setup_distributed_model(args, model_dtype, model_kwargs, logger):
 
     if load_to_meta:
         # Construct model with fake meta tensors, later will be replaced on devices during ds-inference ckpt load
-        with deepspeed.OnDevice(dtype=model_dtype, device="meta"):
+
+        deepspeed_device = "cpu" if keep_module_on_host else "meta"
```
jiminha, line 448: It seems in above code  line#438, if keep_module_on_host is True, the it will never come to this line, why we need this check again?

examples/text-generation/utils.py
jiminha: It seems for distributed model, we only support deepspeed, and whenever deepspeed is on, it's aways bf16 or fp8.  I wonder why we have a limitation like that. 

@jaygala223  Any specific reason why you have to change to config.torch_dtype?   What issue were you seeing with fp8?  Also, all other code setup_model, setup_model_distributed_ep/tp are still using model_dtype by the way.

jaygala223:
Hi @jiminha and @regisss, thanks for reviewing the PR. I have made the necessary changes as per our discussions. Please help review again and let me know if any changes are required. Thanks!

regisss (approved):
LGTM!

PR #1747, opened 2025-02-05
Fix the issue with --load_quantized_model_with_autoawq

schoi-habana:
pip install autoawq downgrades transformers to 4.47. so added transformers>=4.48.2 too

## Before submitting

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ slow_tests_diffusers: test_installs
 
 # Run text-generation non-regression tests
 slow_tests_text_generation_example: test_installs
-	python -m pip install triton==3.1.0 autoawq
+	python -m pip install -r examples/text-generation/requirements_awq.txt
 	BUILD_CUDA_EXT=0 python -m pip install -vvv --no-build-isolation git+https://github.com/HabanaAI/AutoGPTQ.git
 	python -m pip install git+https://github.com/HabanaAI/DeepSpeed.git@1.19.0
 	python -m pytest tests/test_text_generation_example.py tests/test_encoder_decoder.py -v -s --token $(TOKEN)
--- a/examples/text-generation/README.md
+++ b/examples/text-generation/README.md
@@ -734,7 +734,7 @@ Currently, this support is limited to UINT4 inference of pre-quantized models on
 
 Please run the following command to install AutoAWQ:
 ```bash
-pip install triton==3.1.0 autoawq
+pip install -r requirements_awq.txt
 ```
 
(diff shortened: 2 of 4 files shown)


sywangyi (approved):

examples/text-generation/requirements_awq.txt
```diff
@@ -0,0 +1,3 @@
+triton==3.1.0
+autoawq
+transformers>=4.48.2
```
regisss, line 3: Why adding a constraint on the version of Transformers here? This is managed at the library level in `setup.py`

regisss (commented):
I left one comment.

Besides, I see this PR is currently planned to be merged into a branch called `integ_future_01312025`. Shouldn't it go directly to `main`?

examples/text-generation/requirements_awq.txt
schoi-habana: setup.py has Transformers verion constraint in it. but pip install autoawq downgrades Transformers to 4.47 so i had to set transformers version again after autoawq. is there a better way to prevent the override?

regisss: Indeed, weird. Let's put it back again then.

regisss, line 3: Let's also add `<4.49.0` to make sure v4.48 is used

Makefile
```diff
@@ -107,7 +107,7 @@ slow_tests_diffusers: test_installs
 
 # Run text-generation non-regression tests
 slow_tests_text_generation_example: test_installs
-	python -m pip install triton==3.1.0 autoawq
+	python -m pip install triton==3.1.0 autoawq --no-deps
```
regisss, line 110:
```suggestion
	python -m pip install -r examples/text-generation/requirements_awq.txt
```
Let's use the requirements file everywhere so that we don't have to modify the requirements in 3 different places

examples/text-generation/README.md
```diff
@@ -734,7 +734,7 @@ Currently, this support is limited to UINT4 inference of pre-quantized models on
 
 Please run the following command to install AutoAWQ:
 ```bash
-pip install triton==3.1.0 autoawq
+pip install triton==3.1.0 autoawq --no-deps
```
regisss, line 737:
```suggestion
pip install -r examples/text-generation/requirements_awq.txt
```
Same

regisss (commented):
3 small comments and then I'll merge!

regisss (approved):

PR #1748, opened 2025-02-05
Adjust baselines for lower number of epochs improved perplexity, lowe…

emascarenhas:
…r throughput.

# What does this PR do?

Adjusts baselines to match current results with reduced epochs.
Note there is some variability in the results hence the thresholds are set so that the tests will still pass.

Fixes issue.

## Before submitting
- [n ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [y ] Did you make sure to update the documentation with your changes?
- [n] Did you write any new necessary tests?

### DIFF
--- a/tests/baselines/falcon_40b.json
+++ b/tests/baselines/falcon_40b.json
@@ -7,9 +7,9 @@
                 "multi_card": {
                     "learning_rate": 4e-4,
                     "train_batch_size": 1,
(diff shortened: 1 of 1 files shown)


libinta (approved):

regisss (approved):
LGTM!

PR #1751, opened 2025-02-06
fix the output length in image_to_text test to make sure the outpu…

sywangyi:
…t len not be changed in benchmark along with example change

FAILED tests/test_image_to_text_example.py::test_image_to_text_fp8[token0-llava-hf/llava-v1.6-vicuna-13b-hf-1-30.9535718774675] - assert 19.668658371716237 >= ((2 - 1.05) * 30.9535718774675)


regisss (approved):
LGTM!

PR #1755, opened 2025-02-06
Fix Experts Indexing in MoE for Mixtral: Align experts_max with Number of Available Experts

deepak-gowda-narayana:
# What does this PR do?
**Fixes** # (issue)
Segfault on Mixtral transformer test due to out-of-bound reference of experts Index in MOE block
- The test fails with changes to Sparse MOE block from docker build 1.20.0-301 

In this particular failing test, the expert parameter values do not align with the number of weight tensors provided to the MOE. Specifically, the test inputs are as follows:
Expert_max = 7 , number of experts being used in this test is 2.

Hence segmentation fault occurs when it tries to access experts of index range [2-7] as only 2 experts are available.

Changes made in PR:
- Update expert max to ( number of experts - 1 ) , as this is inclusive index range starts from 0.

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/mixtral/modeling_mixtral.py
+++ b/optimum/habana/transformers/models/mixtral/modeling_mixtral.py
@@ -451,7 +451,7 @@ def gaudi_mixtral_block_dynamic_moe_forward(self, hidden_states: torch.Tensor) -
         permuted_weights=True,
         activation="silu",
         experts_min=0,
-        experts_max=1,
+        experts_max= len(self.experts) - 1,
     )
     if is_deepspeed_available() and (not self.training):
         from deepspeed import comm as dist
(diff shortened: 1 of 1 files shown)


deepak-gowda-narayana:
@kwisniewski98 @jiminha Please review PR

optimum/habana/transformers/models/mixtral/modeling_mixtral.py
```diff
@@ -451,7 +451,7 @@ def gaudi_mixtral_block_dynamic_moe_forward(self, hidden_states: torch.Tensor) -
         permuted_weights=True,
         activation="silu",
         experts_min=0,
-        experts_max=7,
+        experts_max=1,
```
jiminha, line 454: Checked further and I think we should actually set the experts_max to num_experts -1.
In this case, we should set it to "len(self.experts)-1"

https://docs.habana.ai/en/latest/PyTorch/Model_Optimization_PyTorch/Custom_Ops_PyTorch.html?highlight=mixture_of_experts

It seems they hardcoded to 7 since  "num_local_experts" is 8 in the current Mixtral models.  However in UT, num_local_experts is 2, that's why we are seeing an error.

deepak-gowda-narayana:
@jiminha @kwisniewski98 Please review updated PR

jiminha (approved):
LGTM!

kwisniewski98 (approved):
LGTM

regisss (approved):
LGTM!

PR #1756, opened 2025-02-07
fix llama FP8 perf issue, kvcache.update should be used since FP8 pat…

sywangyi:
…ches KVCache

 pytest tests/test_text_generation_example.py::test_text_generation_fp8[token0-meta-llama/Llama-2-7b-hf-1-1230-False-128-128-13152.7]

test fail because of throughput drop.


regisss:
@sywangyi I know you edited the message, but your token is still exposed as we can see previous versions of the message. Can you invalidate and rotate your token in your HF account please?

sywangyi:
> @sywangyi I know you edited the message, but your token is still exposed as we can see previous versions of the message. Can you invalidate and rotate your token in your HF account please?

thanks for reminding, I invalidate and refresh the token

regisss (approved):
LGTM!

PR #1758, opened 2025-02-07
Skip few mixtral tests due to experts value issue

jiminha:
We are disabling a few of Mixtral UT due to segfaults. 
This is due to weight tensor size is not matching with experts value in moe. 

This will be enabled once the experts values corrected properly with below PR. 
https://github.com/huggingface/optimum-habana/pull/1755/files


jiminha:
Check further comment on this PR. https://github.com/huggingface/optimum-habana/pull/1755/files

PR #1766, opened 2025-02-11
Update transformer_engine._convert_model to skip LoRA layers

vivekgoe:
This PR updates transformer_enginer._convert_model() function such that lora-A, lora-B linear layers are skipped by TE. This has no impact on device time because these layers translate into small GEMMs whereas it helps in reducing host time by getting rid of TE host overhead.


vivekgoe:
@regisss @libinta please help review and merge this PR.

PR #1771, opened 2025-02-11
Fix llama attr

atakaha:
Add norm_factor attr


libinta (approved):

regisss (approved):
LGTM!

atakaha:
@yafshar , I agree with you. This code is copied from latest OH to align with it. I'll update to use self.scaling.

PR #1774, opened 2025-02-12
update expected output for qwen2

nprotasov:
Update expected output for qwen2 model: current output looks wrong and was generated from main branch, there was error in computing logits and loss.
New output looks more correct


regisss:
@nprotasov Why targeting the branch `integ_future_01312025` and not `transformers_future`?

PR #1781, opened 2025-02-13
Fixed Gemma2 error when saving pretrain

kplau1128:
Fixed Gemma2 model error when saving pretrain.
Due to model.config has non-serializable type.

### DIFF
--- a/optimum/habana/transformers/models/gemma2/modeling_gemma2.py
+++ b/optimum/habana/transformers/models/gemma2/modeling_gemma2.py
@@ -35,7 +35,6 @@ from transformers.models.gemma2.modeling_gemma2 import (
 )
 from transformers.utils import logging
 
-from ....distributed.strategy import DistributedStrategy, NoOpStrategy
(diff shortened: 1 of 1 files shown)


regisss:
Not sure why this was added for Gemma2, I thought we only enabled it for Llama :thinking: 
Any idea @Luca-Calabria @kalyanjk ?

kplau1128:
This `paralllel_strategy` was just initialized but it never been use in Gemma2. And it is not JSON serializable.

Luca-Calabria:
@regisss unfortunately Gemma2 enabling was started by another guy and I closed it after months in standby. Can be possible that this parallel_strategy was added wrongly during the first enabling pass but not fully validated.

kplau1128:
verified text-gen examples with `google/gemma-2-9b-it` model on 1x/multi-x HPU both passed, no regression  as expected

regisss (approved):
LGTM!

PR #1784, opened 2025-02-14
Fix loss calculation (Workaround), final fix TBD

emascarenhas:
# What does this PR do?

Fixes loss calculation when gradient accumulation is on.
The fix is a work around and will be fixed (by regiss) in a future update based on changes in transformers.

## Before submitting
- [n ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [n ] Did you make sure to update the documentation with your changes?
- [n ] Did you write any new necessary tests?


regisss (approved):
LGTM!

PR #1799, opened 2025-02-26
Fix the restart issue for Sentence Transformer STS example in validation

ZhengHongming888:
# What does this PR do?

This PR fixed the restart issue in CI/CD validation caused by sentence transformer STS example. 

The reason might be hard disk full when the STS example is training and saving the checkpoints in hard disk which could up to 200-300G. So this PR set all example training scripts with no saving option for the checkpoints during training. 

All example scripts under sentence transformer has been verified by testing correctly.

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/sentence-transformers-training/nli/training_nli.py
+++ b/examples/sentence-transformers-training/nli/training_nli.py
@@ -148,7 +148,7 @@ def main():
 
     # 8. Save the trained & evaluated model locally
     if args.saving_model_checkpoints:
-        inal_output_dir = f"{output_dir}/final"
+        final_output_dir = f"{output_dir}/final"
         model.save(final_output_dir)
 
     if args.saving_model_checkpoints and args.peft:
--- a/examples/sentence-transformers-training/nli/training_nli.py
+++ b/examples/sentence-transformers-training/nli/training_nli.py
@@ -30,6 +30,7 @@ def main():
     # You can specify any Hugging Face pre-trained model here, for example, bert-base-uncased, roberta-base, xlm-roberta-base
     parser = argparse.ArgumentParser()
     parser.add_argument("model_name", help="model name or path", default="bert-base-uncased", nargs="?")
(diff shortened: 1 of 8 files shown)


examples/sentence-transformers-training/nli/training_nli.py
```diff
@@ -146,13 +146,13 @@ def main():
     test_evaluator(model)
 
     # 8. Save the trained & evaluated model locally
-    final_output_dir = f"{output_dir}/final"
-    model.save(final_output_dir)
+    #final_output_dir = f"{output_dir}/final"
```
libinta, line 149: instead of comment it out, should we have the option that if the save strategy is not no, we keep this?
same for all below?

ZhengHongming888: Thanks @libinta  i will add one more argument to switch on/off saving like as below. How is that? 

parser.add_argument("--saving_model_checkpoints", help="saving checkpoints", action="store_true", default=False)
...
save_strategy="steps" if saving_model_checkpoints else "no"
...
# 8. Save the trained & evaluated model locally
if saving_model_checkpoints:
        inal_output_dir = f"{output_dir}/final"
        model.save(final_output_dir)

if saving_model_checkpoints and args.peft:
        model.eval()
        model = model.merge_and_unload()
        model.save_pretrained(f"{output_dir}/merged")

ZhengHongming888: @libinta verified the STS example with/without saving option running correctly and no problem. Thanks.

ZhengHongming888: @libinta also verified with NLI example with/without saving option correctly. Thanks.

regisss (approved):
LGTM!

There is an issue with the CI runner, hence the failed test. But it passes when I trigger the test manually, merging.

ZhengHongming888:
Thanks @regisss

PR #1801, opened 2025-02-26
Move model to device before wrapping with FSDP

mieshkiwrk:
Seems to be a better approach due to known issue with moving view tensors to device within FSDP: https://github.com/pytorch/pytorch/issues/147321

Also PyTorch recommendation in FSDP tutorial does it like that:
https://pytorch.org/tutorials/intermediate/FSDP_tutorial.html

```
model = Net().to(rank)
model = FSDP(model)
```

### DIFF
--- a/optimum/habana/accelerate/accelerator.py
+++ b/optimum/habana/accelerate/accelerator.py
@@ -476,6 +476,7 @@ class GaudiAccelerator(Accelerator):
                         "limit_all_gathers": fsdp_plugin.limit_all_gathers,
                         "device_id": torch.device("hpu", torch.hpu.current_device()),
                     }
+                    model = model.to(kwargs["device_id"])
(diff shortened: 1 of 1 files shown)


astachowiczhabana (changes requested):
Can you please add short comment explaining why we need to place model on device before FSDP? (i.e. short version from explanation in last comment in https://jira.habana-labs.com/browse/SW-218472 )

mieshkiwrk:
> Can you please add short comment explaining why we need to place model on device before FSDP? (i.e. short version from explanation in last comment in https://jira.habana-labs.com/browse/SW-218472 )

Comment added

12010486:
Could you also add latest commits? I'd like to run a couple of ci tests before the merge

regisss (approved):

PR #1811, opened 2025-03-03
makefile improvements

jasi306:
# What does this PR do?

The Makefile stops test execution after the first failed pytest command.
We ensure that all tests are executed.

## Before submitting

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,10 +82,12 @@ slow_tests_custom_file_input: test_installs
 
 # Run single-card non-regression tests
 slow_tests_1x: test_installs
-	python -m pytest tests/test_examples.py -v -s -k "single_card"
-	python -m pip install peft==0.10.0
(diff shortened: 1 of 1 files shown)


emascarenhas:
@jasi306 , Do we want to run all the tests, so we know which tests failed or do we prefer to exit? If we run all the tests we may find other causes of failures early.

regisss:
> @jasi306 , Do we want to run all the tests, so we know which tests failed or do we prefer to exit? If we run all the tests we may find other causes of failures early.

If I understand correctly, this change will precisely trigger all the tests no?

karol-brejna-i:
@regisss Before the change, if the first test failed, the others were not executed. 
With this PR, we are 'giving a chance' for all the tests to be run.

emascarenhas (approved):

regisss (approved):
LGTM

PR #1824, opened 2025-03-05
update gaudi3 baseline numbers for ci and py tests

hsubramony:
update gaudi3 baseline numbers for ci and py tests


tests/configs/examples/CodeLlama_13b_Instruct_hf.json
regisss, line 1: Why do we need these files in `tests/configs`? Aren't the baselines supposed to be now in `tests/baselines/fixture/tests/test_examples.json`?

uartie: The original test_examples baseline files were actually a combination of test case configuration parameters and the reference baselines.  Now, the config parts are split from the baseline parts into separate locations/files.

regisss: Aaah okay got it, thanks!

regisss (approved):

PR #1807, opened 2025-02-27
Final prep for G3 device context support 

uartie:
# What does this PR do?

This is the final refactor to prep for G3 as a first-class device context.  It will enable the capability to capture separate G3-specific baselines.
1. Converts the rest of the tests to use the baseline fixture.
2. Updates the logic to make Gaudi1 the exception, not Gaudi2.
3. Delay/refactor some asserts and other setup that occurred when module was loaded.
4. Clones the "gaudi2" baselines to "gaudi3" baselines as a starting point.
5. Discontinues GAUDI2_CI environment variable.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1826, opened 2025-03-06
Conftest: get_device_name() without device occupation

dsmertin:
# What does this PR do?

This change fixes a problem with deepspeed test_examples.
`torch_hpu.get_device_name()` has been moved to a separate process, because it occupies a device and doesn't release it.
For tests what work in a separate process if there's a need for all devices not all of them will be available due the occupation from current pytest process.

## Before submitting

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -98,6 +98,19 @@ def pytest_addoption(parser):
 def token(request):
     return Secret(request.config.option.token)
 
+def get_device_name_in_separate_process():
+    # Needs to be done in the separate process, because
+    # `torch_hpu.get_device_name()` occupies a device and doesn't release it.
+    # For tests what work in a separate process if there's a need for all devices
+    # not all of them will be available due the occupation from current pytest process.
+    import subprocess
+    script = \
+    "import habana_frameworks.torch.hpu as torch_hpu\n"\
+    "print(torch_hpu.get_device_name())"
+
+    result = subprocess.run(f"python -c '{script}'", shell=True, capture_output=True, text=True)
+
+    return result.stdout
 
 def pytest_sessionstart(session):
     session.stash["baseline"] = Baseline(session)
@@ -110,12 +123,10 @@ def pytest_sessionstart(session):
         device = "gaudi2" if os.environ["GAUDI2_CI"] == "1" else "gaudi1"
     # Try to automatically detect it
     else:
-        import habana_frameworks.torch.hpu as torch_hpu
-
-        name = torch_hpu.get_device_name().strip()
+        name = get_device_name_in_separate_process()
         if not name:
             raise RuntimeError("Expected a Gaudi device but did not detect one.")
-        device = name.split()[-1].lower()
+        device = name.strip().split()[-1].lower()
 
     # torch_hpu.get_device_name() returns GAUDI for G1
     if "gaudi" == device:
--- a/conftest.py
+++ b/conftest.py
@@ -98,19 +98,6 @@ def pytest_addoption(parser):
 def token(request):
     return Secret(request.config.option.token)
 
-def get_device_name_in_separate_process():
-    # Needs to be done in the separate process, because
-    # `torch_hpu.get_device_name()` occupies a device and doesn't release it.
-    # For tests what work in a separate process if there's a need for all devices
-    # not all of them will be available due the occupation from current pytest process.
-    import subprocess
-    script = \
-    "import habana_frameworks.torch.hpu as torch_hpu\n"\
-    "print(torch_hpu.get_device_name())"
-
-    result = subprocess.run(f"python -c '{script}'", shell=True, capture_output=True, text=True)
-
-    return result.stdout
 
 def pytest_sessionstart(session):
     session.stash["baseline"] = Baseline(session)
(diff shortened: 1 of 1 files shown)


12010486:
@uartie, would you like to have a look? I'm reviewing as well

uartie:
@dsmertin, @12010486, @regisss what if we use `get_device_name()` in the [optimum/habana/utils.py](https://github.com/huggingface/optimum-habana/blob/7450d4b462be0d26de4ae88fc80ac9ac96c5d1bb/optimum/habana/utils.py#L388-L405) module, instead?  Would that prevent occupying a device?

conftest.py
```diff
@@ -98,6 +98,19 @@ def pytest_addoption(parser):
...
+    # For tests what work in a separate process if there's a need for all devices
+    # not all of them will be available due the occupation from current pytest process.
+    import subprocess
+    script = \
+    "import habana_frameworks.torch.hpu as torch_hpu\n"\
+    "print(torch_hpu.get_device_name())"
+
+    result = subprocess.run(f"python -c '{script}'", shell=True, capture_output=True, text=True)
+
+    return result.stdout
```
uartie, line 113: `return results.stdout.strip()` since we check `if not name` by caller (i.e. empty string).

uartie (changes requested):

conftest.py
uartie: That is, on non-gaudi I get:
```python
result = CompletedProcess(args="python -c 'import habana_frameworks.torch.hpu as torch_hpu\nprint(torch_hpu.get_device_name())'", returncode=0, stdout='\n', stderr='/usr/lib/python3.10/inspect.py:288: FutureWarning: `torch.distributed.reduce_op` is deprecated, please use `torch.distributed.ReduceOp` instead\n  return isinstance(object, types.FunctionType)\n/usr/local/lib/python3.10/dist-packages/habana_frameworks/torch/hpu/__init__.py:135: UserWarning: Device not available\n  warnings.warn("Device not available")\n')
```

regisss:
> @dsmertin, @12010486, @regisss what if we use `get_device_name()` in the [optimum/habana/utils.py](https://github.com/huggingface/optimum-habana/blob/7450d4b462be0d26de4ae88fc80ac9ac96c5d1bb/optimum/habana/utils.py#L388-L405) module, instead? Would that prevent occupying a device?

I think that should work yes :+1:

uartie:
> > @dsmertin, @12010486, @regisss what if we use `get_device_name()` in the [optimum/habana/utils.py](https://github.com/huggingface/optimum-habana/blob/7450d4b462be0d26de4ae88fc80ac9ac96c5d1bb/optimum/habana/utils.py#L388-L405) module, instead? Would that prevent occupying a device?
> 
> I think that should work yes 👍

@dsmertin please try this solution.

conftest.py
```diff
@@ -110,12 +123,10 @@ def pytest_sessionstart(session):
...
     # Try to automatically detect it
     else:
-        import habana_frameworks.torch.hpu as torch_hpu
-
-        name = torch_hpu.get_device_name().strip()
+        name = get_device_name_in_separate_process()
         if not name:
             raise RuntimeError("Expected a Gaudi device but did not detect one.")
-        device = name.split()[-1].lower()
+        device = name.strip().split()[-1].lower()
```
uartie, line 129: Change this block to:
```python
from optimum.habana.utils import get_device_name
device = get_device_name()
````

uartie: get_device_name already throws exception if device is unrecognized, so we don't need to do it here anymore.

uartie:
Meanwhile, would be good to fix the source `torch_hpu.get_device_name()` to actually do only what its name implies.  Otherwise, rename it to `torch_hpu.get_device_name_and_occupy()`

12010486:
Ok, I do have a concern in using get_device_name(), as when used on gaudi3, it will output gaudi3, while currently all the tests are either enabled for gaudi2 (and 3 by proxy) or also covering gaudi, a subset of those. @dsmertin, could you check how a small test is running on g3?

12010486:
> Ok, I do have a concern in using get_device_name(), as when used on gaudi3, it will output gaudi3, while currently all the tests are either enabled for gaudi2 (and 3 by proxy) or also covering gaudi, a subset of those. @dsmertin, could you check how a small test is running on g3?

Hold on, it seems now we are taking care of it correctly

regisss (commented):
LGTM!
I think you'll have to run `make style` too (there should be a blank line after the import).

regisss (approved):

uartie:
> > Ok, I do have a concern in using get_device_name(), as when used on gaudi3, it will output gaudi3, while currently all the tests are either enabled for gaudi2 (and 3 by proxy) or also covering gaudi, a subset of those. @dsmertin, could you check how a small test is running on g3?
> 
> Hold on, it seems now we are taking care of it correctly

On "main" branch, we are still waiting for #1807 to be merged.  #1807 is necessary to effectively use `--device gaudi3` or device auto-detection on G3.  Without #1807, you will still want to use "--device gaudi2" or "GAUDI2_CI=1" on G3 until then.

@hsubramony has already pulled #1807 into transformers_4_49 branch via #1824

regisss:
I regularly merge the main branch into transformers_4_49, so it should be ready in a couple of hours

PR #1841, opened 2025-03-11
update target value

Luca-Calabria:
# What does this PR do?

PR created to discuss regression on test 
python -m pytest tests/test_trainer.py -v -s -k test_gradient_accumulation_loss_alignment_with_model_loss

**AssertionError: 1.2642000000000007 not greater than 1.5 : Difference 1.2642000000000007 is not greater than 1.5**

## Before submitting

### DIFF
--- a/tests/test_trainer.py
+++ b/tests/test_trainer.py
@@ -834,7 +834,7 @@ class GaudiTrainerIntegrationPrerunTest(TestCasePlus, GaudiTrainerIntegrationCom
             self.assertLess(max(diff_truth), 0.01, f"Difference {max(diff_truth)} is not within 0.01")
 
             # max diff broken should be very off
-            self.assertGreater(max(diff_broken), 1.5, f"Difference {max(diff_broken)} is not greater than 2")
+            self.assertGreater(max(diff_broken), 1.5, f"Difference {max(diff_broken)} is not greater than 1.5")
 
             loss_base = sum(base_loss_callback.losses)
             loss_broken = sum(broken_loss_callback.losses)
--- a/tests/test_trainer.py
+++ b/tests/test_trainer.py
@@ -834,7 +834,8 @@ class GaudiTrainerIntegrationPrerunTest(TestCasePlus, GaudiTrainerIntegrationCom
             self.assertLess(max(diff_truth), 0.01, f"Difference {max(diff_truth)} is not within 0.01")
 
             # max diff broken should be very off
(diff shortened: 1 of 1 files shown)


Luca-Calabria:
@regisss running this test we see a regression like this 

            # Calculate the difference between the base loss and the grad_accum loss
            diff_truth = [
                abs(base - grad) for base, grad in zip(base_loss_callback.losses, grad_accum_loss_callback.losses)
            ]
            diff_broken = [
                abs(base - grad) for base, grad in zip(base_loss_callback.losses, broken_loss_callback.losses)
            ]
    
            # all diff truth should be quite close
            self.assertLess(max(diff_truth), 0.01, f"Difference {max(diff_truth)} is not within 0.01")
    
            # max diff broken should be very off
>           self.assertGreater(max(diff_broken), 1.5, f"Difference {max(diff_broken)} is not greater than 1.5")
E           AssertionError: 1.2642000000000007 not greater than 1.5 : Difference 1.2642000000000007 is not greater than 1.5

tests/test_trainer.py:837: AssertionError
=================================================================== short test summary info ====================================================================
FAILED tests/test_trainer.py::GaudiTrainerIntegrationPrerunTest::test_gradient_accumulation_loss_alignment_with_model_loss - **AssertionError: 1.2642000000000007 not greater than 1.5 : Difference 1.2642000000000007 is not greater than 1.5**

I'm trying to understand the reason of target value 1.5 and I tried to run the same test on huggingface/transformers tag v4.49 to check if there it passes. What I found is that on original transformers repo it fails but with a different error, so I couldn't check if the target 1.5 is correct or not.

Could you help to try to understand this please?

IlyasMoutawwakil:
it doesn't matter what the value is, the test checks if gradient accumulation works as intended (matches vanilla training loop losses) and fails as intended (does not match vanilla training loop losses). I'm changing it to 1.3 in the upstreaming PR https://github.com/huggingface/transformers/pull/36424/files#r1984837215

IlyasMoutawwakil:
> the diff value is 1.26
interesting, in transformers upstream I got 1.3x, but yeah you can go ahead with 1.2 here as well.

12010486:
LGTM, @regisss could you have a look? 
@libinta FYI

Please, also be aware of https://github.com/huggingface/optimum-habana/pull/1841#issuecomment-2714695932

regisss (approved):

PR #1842, opened 2025-03-11
Discontinue GAUDI2_CI usage

uartie:
# What does this PR do?

Print a NOTICE message if GAUDI2_CI is defined... and delete it before tests collect/run.

Also, print physical device and device context to imply they have different meaning/usage.  Warn if they don't have the same value (i.e. when user overrides it).

## Before submitting


uartie:
cc: @hsubramony

uartie:
cherry-picked from #1807

uartie:
Yep. Thanks. Closing

PR #1815, opened 2025-03-04
Workaround for Transformers 4.49.0 component loading dtype issue

dsocek:
# What does this PR do?

Fixes issue with Transformers 4.49.0 not loading assigned `dtype` in some diffusers components

Transformers release 4.49.0 has an issue that it fails to properly load assigned component dtype for some components.

This bug was created here:
https://github.com/huggingface/transformers/blob/v4.49.0/src/transformers/modeling_utils.py#L1461
```python
torch_dtype = kwargs.pop("torch_dtype", config.torch_dtype)
```

Which changed from older transformers:
https://github.com/huggingface/transformers/blob/v4.46.3/src/transformers/modeling_utils.py#L1508
```python
torch_dtype = kwargs.pop("torch_dtype", torch.get_default_dtype())
```

This bug has been fixed since in 4.50.0dev here:
https://github.com/huggingface/transformers/pull/36335

@regisss If we will still update Optimum-Habana transformers to release 4.49.0, then we need also fix from this PR which enforces assigned `dtype` in a `from_pretrained` wrapper for diffusers pipelines

# Fix Test

Easiest to test is to check text_encoder `dtype` which sometimes fails to get assigned `dtype`:
```python
#!/usr/bin/env python
import torch
from optimum.habana.diffusers import (
   GaudiStableDiffusionPipeline,
   GaudiStableDiffusionXLPipeline,
   GaudiFluxPipeline,
   GaudiStableDiffusion3Pipeline,
)
from optimum.habana.diffusers.pipelines.stable_diffusion_xl.pipeline_stable_diffusion_xl_mlperf import (
   StableDiffusionXLPipeline_HPU,
)

model_name_or_path = "stabilityai/stable-diffusion-xl-base-1.0"
sd_pipe = GaudiStableDiffusionXLPipeline.from_pretrained(
    model_name_or_path,
    torch_dtype=torch.bfloat16,
)
print(sd_pipe.text_encoder.dtype)
print(sd_pipe.text_encoder_2.dtype)

model_name_or_path = "CompVis/stable-diffusion-v1-4"
sd_pipe = GaudiStableDiffusionPipeline.from_pretrained(
    model_name_or_path,
    torch_dtype=torch.bfloat16,
)
print(sd_pipe.text_encoder.dtype)

model_name_or_path = "black-forest-labs/FLUX.1-dev"
sd_pipe = GaudiFluxPipeline.from_pretrained(
    model_name_or_path,
    torch_dtype=torch.bfloat16,
)
[...]

PR #1844, opened 2025-03-12
[transformers_4_49] Fix for AutoModelForCausalLM.from_pretrained()

dsmertin:
# What does this PR do?
Workaround for issue: https://github.com/huggingface/transformers/issues/36258
Reproducing with chatglm3 model:
`pytest tests/test_examples.py -v -s -k test_run_clm_chatglm3-6b_deepspeed`

If `torch_dtype is None`, then we check if this value can be loaded from a model's config. So it replicates behavior of `auto`.
Maybe it's better to modify the test itself to directly pass `torch_dtype=bfloat16` to the models' args.

### DIFF
--- a/examples/language-modeling/run_clm.py
+++ b/examples/language-modeling/run_clm.py
@@ -459,7 +459,7 @@ def main():
 
     # Note that chatglm2/3 has float16 dtype from config.json, and on Gaudi we need to use bfloat16.
     if config.model_type == "chatglm":
-        config.dtype = "torch.bfloat16"
+        config.torch_dtype = torch.bfloat16
 
     tokenizer_kwargs = {
         "cache_dir": model_args.cache_dir,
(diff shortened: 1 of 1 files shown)


examples/language-modeling/run_clm.py
```diff
@@ -484,6 +484,11 @@ def main():
             if model_args.torch_dtype in ["auto", None]
             else getattr(torch, model_args.torch_dtype)
         )
+        # workaraund for https://github.com/huggingface/transformers/issues/36258
+        # TODO: remove after fix is avalible in a release version of `transformers``
+        if torch_dtype is None:
+            torch_dtype = getattr(config, 'torch_dtype', None)
```
Luca-Calabria, line 490: None as default value can still be found for model different than "chatglm". Did you check if it is ok None for other models?

regisss:
I guess because I merged `transformers_4_49` into `transformers_future` and deleted `transformers_4_49`. I just restored it now, let me see if I can change the target.

regisss:
Okay I'll merge this one into `transformers_4_49` and then cherry-pick it into `transformers_future`. Waiting for @Luca-Calabria's comment to be solved and then it should be good to merge.

examples/language-modeling/run_clm.py
Luca-Calabria: I think this is the most generic solution 
**Current change which adds a new functionality: we check model's config in case it was not set earlier.**

12010486: Why not something similar to this? https://github.com/huggingface/optimum-habana/pull/1815 i.e., to modify directly the wrapper. Maybe you were already looking for something similar thought, to revert when we move to 4.50

regisss (approved):
LGTM

PR #1846, opened 2025-03-12
gemma-2b-it perplexity regression fix

skaulintel:
fixes the following pytest, which was indicating a regression in perplexity.

`python -m pytest tests/test_examples.py -v -s -k "test_run_clm_gemma-2b-it_single_card and not Eager" 

The above test started failing after [this commit ](https://github.com/huggingface/optimum-habana/commit/5711346b8ab6ec07130a464c08adfe0d73cbc33c#diff-ea95ad7b4deeb797c4c72182ea73bedb0efc6e774f362d58e92f9272d05e63ba).

I narrowed it down to the differences in the computation of the loss in trainer.py, which is what this PR is. I also noticed this more closely reflects how the loss is calculated in the transformers upstream source code.

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -900,16 +900,6 @@ class GaudiTrainer(Trainer):
         else:
             self.log_evaluate_save_time = None
 
-        # Calculate the number of items in each batch for all epochs
-        num_items_in_batches = self.get_num_items_in_batches(
-            args,
-            epochs_trained,
-            num_train_epochs,
-            train_dataloader,
-            len_dataloader,
-            num_examples,
-        )
-
         hb_profiler = HabanaProfile(
             warmup=self.args.profiling_warmup_steps,
             active=self.args.profiling_steps,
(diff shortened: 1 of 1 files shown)


regisss:
This was added in https://github.com/huggingface/optimum-habana/pull/1794 to solve pretty big throughput regressions. A similar change will probably be merged into Transformers soon: https://github.com/huggingface/transformers/pull/36596
cc @yafshar 

How big is the perplexity regression you saw?

yafshar:
@regisss I have received other reports regarding regression as well, but reverting the changes and setting get_num_items_in_batches to None is not a good solution. I am looking into it again.

regisss:
Yes, I fully agree with you

skaulintel:
@regisss it doesn't seem to be off by very much

`ERROR    root:test_examples.py:785 perplexity: le(26.682445582426915, 1.01 * 26.39)`

should we adjust the baselines for the release then?

regisss:
> @regisss it doesn't seem to be off by very much
> 
> `ERROR root:test_examples.py:785 perplexity: le(26.682445582426915, 1.01 * 26.39)`
> 
> should we adjust the baselines for the release then?

That's not a shocking difference, so I would be fine with updating the baseline. However, @yafshar mentioned above that other people met a similar issue. Let's see first what comes out of https://github.com/huggingface/optimum-habana/pull/1853.

PR #1848, opened 2025-03-13
Use token_idx_cpu int instead of token_idx tensor in slicing (#157)

jaygala223:
Token idx is a tensor and when used within python slicing, it internally calls torch's .item() function. This is unnecessary and slows down the workload. So we replace it with token idx cpu which is already an integer and perfect for this.


yafshar (approved):
LGTM!

@regisss would you please check this PR.

regisss (approved):
LGTM

PR #1850, opened 2025-03-13
Fix unexpected 'num_items_in_batch' argument in GPT-NeoX forward

mounikamandava:
Fixes # (issue) : https://github.com/huggingface/transformers/issues/35838
https://github.com/huggingface/transformers/issues/36074

Reproducing with gpt_neox model :
 pytest tests/test_examples.py::DeepspeedCausalLanguageModelingExampleTester::test_run_clm_gpt-neox-20b_deepspeed

### DIFF
--- a/optimum/habana/transformers/models/gpt_neox/modeling_gpt_neox.py
+++ b/optimum/habana/transformers/models/gpt_neox/modeling_gpt_neox.py
@@ -2,7 +2,6 @@ from typing import Optional, Tuple, Union
 
 import torch
 from transformers.cache_utils import Cache
-from transformers.modeling_flash_attention_utils import FlashAttentionKwargs
 from transformers.modeling_outputs import BaseModelOutputWithPast, CausalLMOutputWithPast
 from transformers.models.gpt_neox.configuration_gpt_neox import GPTNeoXConfig
 from transformers.models.gpt_neox.modeling_gpt_neox import (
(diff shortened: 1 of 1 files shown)


yafshar:
@mounikamandavaI don't believe this hack is an effective solution. Could we upgrade the model to use the new one that incorporates kwargs? https://github.com/huggingface/transformers/blob/v4.49-release/src/transformers/models/gpt_neox/modeling_gpt_neox.py

regisss:
I agree with @yafshar, we should add to `gaudi_gpt_neox_model_forward` the kwargs that are specified here: https://github.com/huggingface/transformers/blob/v4.49-release/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L525

I didn't add them because I naively thought that something called `flash_attn_kwargs` would only involve flash-attention-related stuff, but it seems I was wrong.

regisss:
Reopen

optimum/habana/transformers/models/gpt_neox/modeling_gpt_neox.py
```diff
@@ -269,6 +270,7 @@ def gaudi_gpt_neox_model_forward(
     return_dict: Optional[bool] = None,
     cache_position: Optional[torch.LongTensor] = None,
     token_idx: Optional[torch.Tensor] = None,
+    **flash_attn_kwargs: Unpack[FlashAttentionKwargs],
```
regisss, line 273:
```suggestion
    **kwargs,
```
Let's just add kwargs here and remove the import, we don't need `FlashAttentionKwargs`.

regisss (approved):
LGTM

PR #1851, opened 2025-03-13
Allow explicit control over flash_attention_fast_softmax setting

astachowiczhabana:
- Updated logic to prevent implicit setting of flash_attention_fast_softmax
- Users can now explicitly set flash_attention_fast_softmax to False to prevent it from being automatically set to True when use_flash_attention is enabled.
- Added warning messages to inform users of implicit changes to argument values

### DIFF
--- a/optimum/habana/utils.py
+++ b/optimum/habana/utils.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import argparse
 import random
 import subprocess
 import time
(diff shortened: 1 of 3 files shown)


optimum/habana/utils.py
```diff
@@ -415,3 +416,37 @@ def get_device_count():
         return htexp.hpu.device_count()
     else:
         raise ValueError("No hpu is found avail on this system")
+
+class SetTrueOrFalseOrNone(argparse.Action):
```
regisss, line 420: Since this is only used in the text-generation example, I think this should rather go in https://github.com/huggingface/optimum-habana/blob/main/examples/text-generation/utils.py

regisss (approved):
LGTM!

Next time let's try to merge such PRs in main first, I'll cherry-pick them in the release branch anyway.

PR #1853, opened 2025-03-13
optimize PR 1794 to improve runtime regressions in training

hsubramony:
optimize PR 1794 to improve runtime regressions in training

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -900,16 +900,6 @@ class GaudiTrainer(Trainer):
         else:
             self.log_evaluate_save_time = None
 
-        # Calculate the number of items in each batch for all epochs
-        num_items_in_batches = self.get_num_items_in_batches(
-            args,
-            epochs_trained,
-            num_train_epochs,
-            train_dataloader,
-            len_dataloader,
-            num_examples,
-        )
-
         hb_profiler = HabanaProfile(
             warmup=self.args.profiling_warmup_steps,
             active=self.args.profiling_steps,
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/trainer.py
```diff
@@ -2592,74 +2581,26 @@ def _zero_model_grad(self, model):
...
-            total_updates -= 1
-
-        num_items_in_batches = []
-        for epoch in range(epochs_trained, num_train_epochs):
-            epoch_dataloader = train_dataloader
-            if hasattr(epoch_dataloader, "set_epoch"):
-                epoch_dataloader.set_epoch(epoch)
-
-            epoch_iterator = iter(epoch_dataloader)
+    def get_batch_samples(self, epoch_iterator, num_batches):
```
yafshar, line 2584: @hsubramony you do not need to rewrite this function it is available in transformers

hsubramony: done

regisss:
Hmm how does this help to solve the runtime regressions? If I'm not mistaken, your change consists in using `get_batch_samples` from the Transformers trainer, which we know lead to some regressions no? Or am I missing something?

yafshar:
> Hmm how does this help to solve the runtime regressions? If I'm not mistaken, your change consists in using `get_batch_samples` from the Transformers trainer, which we know lead to some regressions no? Or am I missing something?

You are right. But we are evaluating which regression is worse for the current release to evaluate this issue later and address the regression

regisss:
Based on @skaulintel's comment [here](https://github.com/huggingface/optimum-habana/pull/1846#issuecomment-2722046162), there is a 1.1% regression in perplexity for one test. It's really close to our tolerance threshold of 1%.
However, I remember the throughput was easily 20% slower for a few tests without this change. Maybe we need more samples of accuracy regression to get a better estimation of it.

hsubramony:
@regisss  we are running single and mult card tests with this change. Will update eod

hsubramony:
i found atleast for 3 model training tests with improved runtime..vit,swin,wav2vec

PR #1857, opened 2025-03-18
Fix MPT regression

atakaha:
Remove softmax wrapping class to fix MPT inference regression.

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/mpt/modeling_mpt.py
+++ b/optimum/habana/transformers/models/mpt/modeling_mpt.py
@@ -43,19 +43,9 @@ except ImportError:
 logger = logging.get_logger(__name__)
 
 
-class Softmax(nn.Module):
-    def __init__(self):
-        super().__init__()
-
-    def forward(self, x, dim=None, invAttnHead=None):
-        return torch.nn.functional.softmax(x, dim)
-
-
 class GaudiMptAttention(MptAttention):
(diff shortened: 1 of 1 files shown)


regisss:
Not sure why we had this `Softmax` class in the first place actually :thinking: 
Can you provide a command to compare before/after the fix please?

atakaha:
> Not sure why we had this `Softmax` class in the first place actually 🤔 Can you provide a command to compare before/after the fix please?

The command is
`python3 run_generation.py  \
  --model_name_or_path mosaicml/mpt-7b --bf16 \
  --batch_size 1 --use_hpu_graphs \
  --n_iterations 1 --use_kv_cache --do_sample  --n_iterations 5 \
  --prompt "Some of you are too young to relate to this, but North American Gen-Xers like myself remember childhood summers where we would leave home after breakfast to play and not come home for eight or nine hours. As long as we were home in time for dinner, our parents didn't raise an eyebrow. It was considered normal. If you're a ten-year old, try that today and see what happens. Amber Alerts will rain down from the heavens in an organized fury! This isn't a criticism of the Amber Alert system. The idea of your child going missing" \
  --max_new_tokens 1932`

regisss (approved):
LGTM, thanks for fixing this!

PR #1864, opened 2025-03-19
fix typo in modeling llama

hsubramony:
fix merge typo
lower batch size of distributed tp example
matches CI test params with readme for distributed tp example


regisss (approved):
LGTM

PR #1869, opened 2025-03-20
Count `num_items_in_batch` only when necessary

IlyasMoutawwakil:
# What does this PR do?

Fixes regressions

## Before submitting


optimum/habana/transformers/trainer.py
```diff
@@ -2593,91 +2581,44 @@ def _zero_model_grad(self, model):
...
-
-            device = first_batch["labels"].device
-
-            # Reset the iterator
-            epoch_iterator = iter(epoch_dataloader)
-
-            num_items_in_batches.append([])
-            for update_step in range(total_updates):
-                num_batches = args.gradient_accumulation_steps if update_step != (total_updates - 1) else remainder
+                batch = next(epoch_iterator)
```
regisss, line 2600: Let's have first 
```py
try:
    batch = next(epoch_iterator)
except StopIteration:
    break
```
and then the block
```py
if count_num_items_in_batch:
    try:
        ...
    except ...:
        ...
```
That way the code is a bit easier to compare to Transformers (even though that's a change you should probably do in Transformers IMO) and we avoid having two nested try-except block.

WDYT?

IlyasMoutawwakil: Yes that works too, I actually changed many things compared to transformers on the journey to optimize this function xD
In this one there's also one loop vs two decoupled loops in transformers (one is list comprehension).
at some point I even pre-allocated `batch_samples = [None] * num_batches` 😭

optimum/habana/transformers/trainer.py
```diff
@@ -2593,91 +2581,41 @@ def _zero_model_grad(self, model):
...
-                epoch_dataloader = train_dataloader
-
-            if hasattr(epoch_dataloader, "set_epoch"):
-                epoch_dataloader.set_epoch(epoch)
-
-            epoch_iterator = iter(epoch_dataloader)
+        for _ in range(num_batches):
             try:
-                first_batch = next(epoch_iterator)
+                batch = next(epoch_iterator)
```
yafshar, line 2600: @IlyasMoutawwakil I think having `count_num_items_in_batch`  is very useful in some cases, and can prevent extra counting when unnecessary but separating the loops is not helping. 
```py
        for _ in range(num_batches):
            try:
                batch_samples += [next(epoch_iterator)]
            except StopIteration:
                break

        count_num_items_in_batch = count_num_items_in_batch and len(batch_samples) > 0 and "labels" in batch_samples[0]
        if count_num_items_in_batch:
```

The bottleneck is when you are accessing the batch.

IlyasMoutawwakil: I'm not sure I understand what you mean by "separating the loops" 😅
So we keep pre-computing num batches doe all steps in advance ? but add `count_num_items_in_batch` condition to that ?

yafshar: What I mean is that you trying to count the num_items in one loop `for _ in range(num_batches):` different than up stream. transfomerser first get the batch_samples then in the second loop count the items. In my profiling counting in one loop like yours does not address the issue. The bottleneck is when you are accessing the data in batch. So you can do as up stream and only add your new flag which helps in some cases

IlyasMoutawwakil: I reverted the whole thing to the original transformers code because imo, a step by step approach results in better runtime overall, allowing better overlap of compute and i/o. while a simulating the training loop and computing everything beforehand has more chance of a bottleneck, because you're iterating on the dataloader too fast.

IlyasMoutawwakil: > The bottleneck is when you are accessing the data in batch. So you can do as up stream and only add your new flag which helps in some cases

Ah yes, yeah I addressed that in https://github.com/huggingface/optimum-habana/pull/1869#discussion_r2005290704, it's left over from my attempt to reduce time taken in the function, I can remove that.

optimum/habana/transformers/trainer.py
```diff
@@ -2593,91 +2581,42 @@ def _zero_model_grad(self, model):
...
-                # Set to None if no items in batch
-                if num_items_in_batch == 0:
-                    num_items_in_batch = None
+        count_num_items_in_batch = (
+            len(batch_samples) > 0
+            and "labels" in batch_samples[0]
+            and (
+                # num_items_in_batch is passed to model forward
+                # https://github.com/huggingface/transformers/blob/v4.49.0/src/transformers/trainer.py#L3757
+                self.model_accepts_loss_kwargs
```
hsubramony, line 2600: cant we move this check before calling get_batch_samples() ?

yafshar: we can but this is not a bottleneck at all.

regisss: Actually we still need to call `get_batch_samples` to get the batches

regisss:
@yafshar My last commit also removes the call to `.item` which is not needed and fixes some regressions (GPT2 single card for example)

yafshar:
> @yafshar My last commit also removes the call to `.item` which is not needed and fixes some regressions (GPT2 single card for example)

I just tested it. It recovers the performance for that case

optimum/habana/transformers/trainer.py
```diff
@@ -2593,91 +2581,42 @@ def _zero_model_grad(self, model):
...
+            try:
+                num_items_in_batch = sum([(batch["labels"].ne(-100)).sum() for batch in batch_samples])
+            except (TypeError, AttributeError):
+                pass
 
-            # For iterable datasets, don't do more than max_steps steps
-            if len_dataloader is None and global_step >= args.max_steps:
-                break
+        if self.args.average_tokens_across_devices and num_items_in_batch is not None:
+            num_items_in_batch = self.accelerator.gather(num_items_in_batch).sum()
```
yafshar, line 2617:
```suggestion
        if num_items_in_batch is not None:
            if self.args.average_tokens_across_devices:
                num_items_in_batch = self.accelerator.gather(num_items_in_batch).sum()
            if torch.is_tensor(num_items_in_batch):
                num_items_in_batch = num_items_in_batch.to(device)
```

optimum/habana/transformers/trainer.py
```diff
@@ -2593,91 +2581,42 @@ def _zero_model_grad(self, model):
...
 
-        return num_items_in_batches
+        if torch.is_tensor(num_items_in_batch):
+            num_items_in_batch = num_items_in_batch.to(device)
```
yafshar, line 2620:
```suggestion
```

optimum/habana/transformers/trainer.py
yafshar: in upstream there is item before that is why they have to check it

regisss: Done

regisss: > in upstream there is item before that is why they have to check it

Yes, you're right. I just removed it.

yafshar:
@IlyasMoutawwakil This is still in draft mode, can you mark it ready.

yafshar (approved):
LGTM!

@regisss there are still some regression, but small. I think this is the best solution at the moment.

regisss:
> LGTM!
> 
> @regisss there are still some regression, but small. I think this is the best solution at the moment.

Can you share more information about the other regressions?

yafshar:
> > LGTM!
> > @regisss there are still some regression, but small. I think this is the best solution at the moment.
> 
> Can you share more information about the other regressions?

Throughput regressions are OK from my side. Since our ref numbers are for cases without num_items. But I see some regression in perplexity. For example,

```sh
>>> python -m pytest tests/test_examples.py -s -v -k MultiCardCausalLanguageModelingVeraExampleTester 
ERROR    root:test_examples.py:785 perplexity: le(10.276098174155887, 1.01 * 9.064502567217577)
```
there is an increase from 9.15 to 10.27

This could be bug or simple mistake in numbers elsewhere and not related to this change

yafshar:
Upon further review, these changes might be expected because our loss computation has been slightly modified. This alteration affects the train_loss, which in turn influences the model’s training process. As a result, the model performs evaluation based on a different training trajectory, leading to variations in eval_loss and perplexity, since perplexity is derived as exp(eval_loss) and would naturally reflect these changes. Therefore, we should update our reference metrics to account for this shift.

slokesha:
Adding a warmup epoch_iterator is helping with the eval regression seen.

Adding the following code at L920 before the use if epoch_iterator is increasing the eval in few models:
```
      epoch_iterator = iter(epoch_dataloader)
      next(epoch_iterator)

```
For **test_run_summarization_flan-t5-xxl_deepspeed**
Using warmup iterator   eval_rougeLsum   =    30.1148 instead of 27.9095

For **test_run_clm_gemma-2b-it_deepspeed**
Using warmup iterator  perplexity   =   846.3532 instead of 1126.229

yafshar:
> Adding a warmup epoch_iterator is helping with the eval regression seen.
> 
> Adding the following code at L920 before the use if epoch_iterator is increasing the eval in few models:
> 
> ```
>       epoch_iterator = iter(epoch_dataloader)
>       next(epoch_iterator)
> ```
> 
> For **test_run_summarization_flan-t5-xxl_deepspeed** Using warmup iterator eval_rougeLsum = 30.1148 instead of 27.9095
> 
> For **test_run_clm_gemma-2b-it_deepspeed** Using warmup iterator perplexity = 846.3532 instead of 1126.229

@slokesha Thanks for the feedback. What you are suggesting could be a valuable improvement after thorough testing. I recommend opening a new PR with this enhancement. I don't believe it is related to the current PR and should not be included here.

PR #1879, opened 2025-03-25
Disabling timers synchronization (#154)

bhargaveede:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/language-modeling/llama2_ds_zero3_config.json
+++ b/examples/language-modeling/llama2_ds_zero3_config.json
@@ -13,4 +13,10 @@
         "contiguous_gradients": false,
         "stage3_gather_16bit_weights_on_model_save": true
     }
+    "timers": {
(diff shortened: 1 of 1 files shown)


vidyasiv:
@bhargaveede thanks for your PR. Could you provide a description of the change and why it is needed?

vidyasiv (changes requested):
@bhargaveede thanks for your PR. Could you provide a description of the change and why it is needed?

bhargaveede:
This change is done to improve some perf. Without this change, Timer synchronization waits on host and that creates little idle time. This change avoids that synchronization resulting in better device utilization.

vidyasiv (approved):
@regisss please take a look and let us know if any further testing is needed

regisss (approved):
LGTM

PR #1876, opened 2025-03-24
Update GaudiAccelerator

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1885, opened 2025-03-25
fix lost modules in regional compilation

xinyu-intel:
Change-Id: I8cd89a0f2e58e94582c603f3d936a6d013c7c125

# What does this PR do?

we also need to compile the leaf modules when regional compilation is enabled.

## Before submitting

### DIFF
--- a/optimum/habana/accelerate/accelerator.py
+++ b/optimum/habana/accelerate/accelerator.py
@@ -75,8 +75,15 @@ def compile_regions(model, compile_kwargs):
             module.__dict__.pop("_parameters", None)
             setattr(model, name, module)
     else:
-        for _, module in model.named_children():
-            compile_regions(module, compile_kwargs)
+        if model._modules:  # If model has submodules, recurse and reassign
+            for name, module in model.named_children():
+                compiled_module = compile_regions(module, **compile_kwargs)
+                if compiled_module is not None:  # Only reassign if something is returned
+                    setattr(model, name, compiled_module)
+        else:  # Leaf node
+            model = torch.compile(model, **compile_kwargs)
+            model.__dict__.pop("_parameters", None)
+            return model
 
 
 class GaudiAccelerator(Accelerator):


optimum/habana/accelerate/accelerator.py
```diff
@@ -589,8 +589,14 @@ def compile_regions(self, model):
...
-            for _, module in model.named_children():
-                self.compile_regions(module)
+            if model._modules:
+                for _, module in model.named_children():
+                    self.compile_regions(module)
+            else:
+                if self.dynamic is not None:
+                    model.compile(dynamic=self.dynamic, **self.state.dynamo_plugin.to_kwargs())
+                else:
+                    model.compile(**self.state.dynamo_plugin.to_kwargs())
```
yafshar, line 599:
```suggestion
        if model._modules:  # If model has submodules, recurse and reassign
            for name, module in model.named_children():
                compiled_module = self.compile_regions(module)
                if compiled_module is not None:  # Only reassign if something is returned
                    setattr(model, name, compiled_module)
        else:  # Leaf node
            if self.dynamic is not None:
                compiled_model = torch.compile(model, dynamic=self.dynamic, **self.state.dynamo_plugin.to_kwargs())
            else:
                compiled_model = torch.compile(model, **self.state.dynamo_plugin.to_kwargs())
            compiled_model.__dict__.pop("_parameters", None)
            return compiled_model  # Return compiled leaf for reassignment
```


I think use of `model.compile(...)` for leaf nodes is not a good choice. To align with PyTorch’s standard compilation API and ensure compatibility, I’d suggest switching it to `torch.compile(model, ...)` and handling the reassignment of the compiled module. Since `torch.compile` returns a new compiled object, we need to propagate it back up the hierarchy.

bsochack: @yafshar I'm not sure if I understand your proposal.
The regional compilation is standard technique: https://pytorch.org/tutorials/recipes/regional_compilation.html
It has several advantages that helps in this case.

The patch here ensures that all modules are compiled, so there is no eager execution.

yafshar: @bsochack I did not reject the regional compilation! I just asked the author to fix the code and improve it to align with PyTorch’s standard compilation API . `model.compile` is not a good choice.

yafshar: My proposal is just improving the patch

vivekgoe: @xinyu-intel Change proposed by @yafshar seems like the correct thing to do to me. Although we do not see any side-effects from the current code for LLama3.1-8B model that you tested with, we may have problems in some other cases in future by not propagating compiled leaf module upwards.  Maybe you can make the change and do a quick sanity check to ensure it does not break LLama3.1-8B perf. & accuracy you were getting.

xinyu-intel: @yafshar @vivekgoe verified with LLama3.1-8B perf. & accuracy. thx:)

yafshar (approved):
LGTM!

@regisss this PR is ready for your final review.

regisss (approved):
LGTM!

Shouldn't be an issue, but let's wait for https://github.com/huggingface/optimum-habana/pull/1876 to be merged before merging this one.

regisss:
@xinyu-intel Following #1876, there are merge conflicts to solve

xinyu-intel:
> @xinyu-intel Following #1876, there are merge conflicts to solve

@regisss @vivekgoe @yafshar rebased, please review again:)

PR #1887, opened 2025-03-25
Support allow_unspec_int_on_nn_module

xinyu-intel:
Change-Id: Id7a21c0c1e014e9dc8c503c645a252ad4da136fb

# What does this PR do?

This is an option from https://github.com/pytorch/pytorch/pull/142829 to not specialize int from nn.Modules under torch.compile mode

## Before submitting


libinta:
@regisss  it's to support https://github.com/pytorch/pytorch/pull/142829

PR #1889, opened 2025-03-26
[SW-218526] Updated Readme files for explicite lazy mode part2 (#177)

jasi306:
# What does this PR do?

Fixes # (issue)

## Before submitting


regisss:
@jasi306 Is it possible to merge this PR and #1890 ? Unless there is a good reason to keep them separated of course

emascarenhas:
@jasi306 , Since --use_hpu_graphs works only with Lazy mode, shouldn't the PT_HPU_LAZY_MODE=1 be used anywhere we see use_hpu_graphs? Or are there cases where it makes sense to not use Lazy mode while setting use_hpu_graphs for inference/finetuning/training? See some cases in speech-recognition.

emascarenhas:
> @jasi306 , Since --use_hpu_graphs works only with Lazy mode, shouldn't the PT_HPU_LAZY_MODE=1 be used anywhere we see use_hpu_graphs? Or are there cases where it makes sense to not use Lazy mode while setting use_hpu_graphs for inference/finetuning/training? See some cases in speech-recognition.

Hi @jasi306 , Any update?

jasi306:
Just created new PR aggregating part 1 and 2:
https://github.com/huggingface/optimum-habana/pull/1921

PR #1890, opened 2025-03-26
[SW-218526] Updated Readme files for explicite lazy mode (#174)

jasi306:
# What does this PR do?

Fixes # (issue)

## Before submitting


regisss:
@jasi306 Do we know when lazy mode will not be the default anymore?

PR #1894, opened 2025-03-26
[Llama-Vision] Trim logits (#162)

ANSHUMAN87:


optimum/habana/transformers/models/mllama/modeling_mllama.py
```diff
@@ -908,6 +908,12 @@ def forward(
         )
 
         hidden_states = outputs[0]
+        _, seq_len, _ = hidden_states.shape
+        if seq_len > 1 and not self.training:
```
yafshar, line 912: @ANSHUMAN87 Does your change also require adding `trim_logits` to the forward interface? and do below modification? 
`if seq_len > 1 and trim_logits and not self.training:`

regisss: I agree with @yafshar, I'm curious about this too :eyes:

PR #1895, opened 2025-03-26
Auto pr ada07fc

ANSHUMAN87:


examples/image-to-text/run_pipeline.py
```diff
@@ -150,6 +150,19 @@ def main():
         help="The token to use as HTTP bearer authorization for remote files. If not specified, will use the token "
         "generated when running `huggingface-cli login` (stored in `~/.huggingface`).",
     )
+    parser.add_argument(
+        "--bucket_size",
```
yafshar, line 154: @ANSHUMAN87 where does the default value gets corrected? If I set --bucket_internal only I get the failure 
```sh
  File "/usr/local/lib/python3.10/dist-packages/optimum/habana/transformers/generation/utils.py", line 1080, in generate
    assert generation_config.bucket_size >= 0, "please set bucket_size to use bucket_internal"
AssertionError: please set bucket_size to use bucket_internal
```

libinta: bucket_internal and bucket_size should both be set

yafshar: > bucket_internal and bucket_size should both be set

That is correct, but the help convey a different message. It says there is a default value!

regisss: I agree with @yafshar, the help message says that we use `shape = prompt_length + max_new_tokens` if `bucket_size` is negative. But that is useless if the code asserts whether `bucket_size` is positive. What's the intended behaviour here?

ANSHUMAN87: Yes I also agree with the message being wrong. But this is borrowed from text-generation. I think better it should be taken up collectively. Let me know your thoughts. Thanks!

regisss: As a quick workaround, I suggest to update the help message of this arg right below to state that `bucket_size` cannot be negative.

regisss (approved):
LGTM

PR #1896, opened 2025-03-27
chatglm: Fix a bug when attention mask is None

mengker33:
When attention mask is None, we need to ensure the newly created mask size is a meaningful value (i.e. not zero).

# What does this PR do?
Running evaluation cmd:
`python run_lm_eval.py --model_name_or_path THUDM/chatglm3-6b --device hpu  --use_hpu_graphs --use_kv_cache -- 
 batch_size 1  --tasks winogrande --output_file lm_eval_chatglm3-6b.json --bf16`
we will see the failure invoking `_gaudi_prepare_4d_causal_attention_mask`:
` return causal_4d_mask.masked_fill(expanded_attn_mask, torch.finfo(dtype).min)
RuntimeError: The size of tensor a (384) must match the size of tensor b (0) at non-singleton dimension 3
`

This error logs shows the attention_mask shape's 2nd dim is 0, and it is because in ChatGLMModel forward path, if attention_mask is None, a mask is created with shape [batch_size, past_seen_tokens]; past_seen_tokens can be 0 when no caching available. 

As for the reason behind attention_mask is None, I think it relates to the way run_lm_eval.py invokes the models' generation method. So here in run_lm_eval.py, the generation does not go into the normal process of OH generation (i.e. first model's prepare_inputs_for_generation which will build the compact inputs, including input_ids, attention_mask and all other params, then into model's generation), instead it directly calls self.model(xxx) without input preparation step. Then we will see the failure in the log above.

Therefore, I modified the code as in this PR to ensure the newly created mask shape always has a valid value.

## Before submitting


mengker33:
> @mengker33 Could you please add the issue that caused the problem in the PR README, along with the results before and after your fix? This will make the PR more informative

Thanks for your comment! I have updated more details in this PR's description, please check above.

regisss (approved):
LGTM

PR #1898, opened 2025-03-28
Implementation of GRPO in Gaudi TRL

alekseyfa:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/habana/transformers/trainer.py
```diff
@@ -1015,8 +1015,11 @@ def hpu_deepspeed_checkpointing(function, *checkpoint_args, use_reentrant: Optio
 
                     # attn_softmax_bf16 and use_flash_attention is enabled only for llama, qwen2, starcoder2, gemma, baichuan and chatglm
                     # lazy_mode for llama, qwen2, starcoder2 and mistral
+
+                    # FIXME: This is a temporary solution to avoid breaking the training loop
```
yafshar, line 1019: @alekseyfa this temporary solution is wrong? It will break everything else

alekseyfa: I've reverted my changes. This was a temporary workaround, just to quickly test the functionality. Fixed by updating the data collator

PR #1899, opened 2025-03-28
detr: fix possible incorrect tensor type

mgonchar:
# What does this PR do?

DETR model assumes that ```class_label``` is a ```LongTensor```, see docs [here ](https://huggingface.co/docs/transformers/en/model_doc/detr)

Current code may default to other type in case of empty ```entries```, i.e. in such situation in most cases it produces empty float tensor, which subsequently causes issues during indexing in loss calculation

## Before submitting


imangohari1:
@mgonchar can you share a test case that this code as is crashes and the PR fixes it? 
can you also run the example with changes and make sure no change is functionality/perf is observed? https://github.com/huggingface/optimum-habana/blob/5247cf91994caee9cf3788bcafb9f1f6b9cdd08b/examples/object-detection/run_example.py#L35 tnx.

mgonchar:
hi @imangohari1 the issue is reproduced on my training script, which uses ```GaudiTrainer``` and ```keremberke/license-plate-object-detection``` dataset. Problem arises on samples, which have empty ```class_label``` list.

I ran test that you suggested and there is no diff introduced by this change.

before change
```
------------------------------------------------------------------------------
Detected cat with confidence 1.0 at location [344.0, 25.25, 640.0, 376.0]
Detected remote with confidence 0.996 at location [328.0, 76.5, 372.0, 188.0]
Detected remote with confidence 0.996 at location [38.75, 70.5, 175.0, 118.0]
Detected cat with confidence 1.0 at location [14.38, 52.5, 316.0, 472.0]
Detected couch with confidence 0.996 at location [-1.25, 0.94, 640.0, 472.0]

Stats:
-----------------------------------------------------------
Total latency (ms): 740.450382232666 (for n_iterations=10) 
Average latency (ms): 74.0450382232666 (per iteration) 
-----------------------------------------------------------
```


after change
```
Detected cat with confidence 1.0 at location [344.0, 25.25, 640.0, 376.0]
Detected remote with confidence 0.996 at location [328.0, 76.5, 372.0, 188.0]
Detected remote with confidence 0.996 at location [38.75, 70.5, 175.0, 118.0]
Detected cat with confidence 1.0 at location [14.38, 52.5, 316.0, 472.0]
Detected couch with confidence 0.996 at location [-1.25, 0.94, 640.0, 472.0]

Stats:
------------------------------------------------------------
Total latency (ms): 739.0251159667969 (for n_iterations=10) 
Average latency (ms): 73.90251159667969 (per iteration) 
------------------------------------------------------------
```

sandeep-maddipatla:
Hi @imangohari1, @mgonchar - I added the `get_targets_without_no_objects` function as a means to enable bucketing for the input shapes involved with this model. The idea was that the calling script would pad the target object list to the maximum length across the training set, and mark the padded objects (which are invalid). This function is supposed to filter those objects out, so as to not affect the training accuracy.

I think the code I had doesn't consider the case that there are no target objects for a given image. For this case, I think @mgonchar 's is right that if `entries` turns out to be empty, it needs to be explicitly cast to `torch.int64`. The datasets I tested this code for earlier didn't have such cases.

> the issue is reproduced on my training script, which uses GaudiTrainer and keremberke/license-plate-object-detection dataset. Problem arises on samples, which have empty class_label list.

However, I couldn't find such an image in the above dataset. Below query (shown for `full_train` split) yields no results in all splits of the dataset. Can you pls confirm we are hitting this bug for only a case not having empty `class_label` list? I just want to make sure we are not missing anything.

```
SELECT * 
FROM full_train
WHERE NOT list_contains(objects.category, 0)
```

sandeep-maddipatla:
> it turned out that for some input images augmentation just drop bboxes and objects (maybe because they are considered small or for whatever other reason). So dataloader may produce images without labels even though there are no such inside dataset itself.

Thank you @mgonchar for looking into this, and confirming the reason. I am fine with the change you proposed - I think it can be merged in.

imangohari1 (approved):
LGTM! @regisss please take a look. Thank you.

regisss (approved):
LGTM

PR #1905, opened 2025-04-01
Fix the flant5-xl training issue 

libinta:
Fix the training issue with an unexpected keyword argument 'num_items_in_batch'

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/t5/modeling_t5.py
+++ b/optimum/habana/transformers/models/t5/modeling_t5.py
@@ -516,7 +516,7 @@ def gaudi_T5ForConditionalGeneration_forward(
     return_dict: Optional[bool] = None,
     cache_position: Optional[torch.LongTensor] = None,
     token_idx: Optional[torch.LongTensor] = None,
-    num_items_in_batch:Optional[int] = None,
+    **kwargs,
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/t5/modeling_t5.py
```diff
@@ -516,6 +516,7 @@ def gaudi_T5ForConditionalGeneration_forward(
     return_dict: Optional[bool] = None,
     cache_position: Optional[torch.LongTensor] = None,
     token_idx: Optional[torch.LongTensor] = None,
+    num_items_in_batch:Optional[int] = None,
```
yafshar, line 519:
```suggestion
    **kwargs,
```

@libinta For several models, we are utilizing `**kwargs` to handle variable keyword arguments. I recommend continuing this approach instead of using specific parameters like `num_items_in_batch` to maintain consistency and flexibility across our codebase.

regisss:
This issue doesn't happen with https://github.com/huggingface/optimum-habana/pull/1869. The reason why is that `num_items_in_batch` will be None and thus it is not passed to the model forward.

But I'm going to merge it as I think this issue could still occur if the user provides a custom loss function.

regisss (approved):

PR #1913, opened 2025-04-09
Update setup.py with Xet Storage

12010486:
### Why this PR

Solve the warning:
```
WARNING  huggingface_hub.file_download:file_download.py:1670 Xet Storage is enabled for this repo, but the 'hf_xet' package is not installed. Falling back to regular HTTP download. For better performance, install the package with: `pip install huggingface_hub[hf_xet]` or `pip install hf_xet`
```
I've noticed an improve in performance for `clip-roberta_multi_card`, when running:

```
python -m pytest tests/test_examples.py::MultiCardVisionLanguageExampleTester::test_run_clip_clip-roberta_multi_card --device gaudi3 --log-cli-level 20 -v -s
```
after `pip install hf_xet` results are less flaky and we get:

```
--------------------------------------------------------------------------------------------------------- live log call ----------------------------------------------------------------------------------------------------------
INFO     root:conftest.py:72 gaudi3.train_runtime:actual = 67.7085
INFO     root:conftest.py:73 gaudi3.train_runtime:ref    = 71.4956
INFO     root:conftest.py:72 gaudi3.train_samples_per_second:actual = 19684.296
INFO     root:conftest.py:73 gaudi3.train_samples_per_second:ref    = 20525.583
PASSED
```


12010486:
@hsubramony what are your thoughts about it?

regisss:
I think we'll want to install `hf_xet` by default. Maybe we can even replace the transformers dependency by `transformers["hf_xet"]`, WDYT?

12010486:
Sure, thanks for the advice. Do you think there is the need for further tests if we install it as `transformers["hf_xet"]`? I can allocate some time in that case

regisss:
I think it would make sense yes, especially since you noticed an improvement for `clip-roberta_multi_card`.
Xet should make model upload and download faster, so all tests where model checkpoints are saved or loaded from the hub during training may be (positively) impacted. If that's the case, we'll have to update our baselines accordingly to make sure we don't miss other potential regressions.

regisss:
@12010486 Closing this PR as I added the xet extra to Transformers in `setup.py` in the PR to upgrade to Transformers v4.51: https://github.com/huggingface/optimum-habana/pull/1943/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R32

PR #1914, opened 2025-04-09
Use data cache in slow_tests_8x (#214)

karol-brejna-i:
Allows for using cached datasets (http://images.cocodataset.org) in slow tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ slow_tests_1x: test_installs
 
 # Run multi-card non-regression tests
 slow_tests_8x: test_installs
-	python -m pytest tests/test_examples.py -v -s -k "multi_card"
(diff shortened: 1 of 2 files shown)


regisss:
Does this change anything when actually running the tests?

regisss:
> @regisss this saves the model download time

Yes, I do understand that but that's already possible. This change is needed specifically for having the env variable specified in the Makefile?

gplutop7:
@regisss Yes, in theory we could set DATA_CACHE as an environment variable in Jenkins and have it inherited by Make and Python. However, there's no guarantee it would consistently propagate through all layers — especially if the Makefile modifies the environment or runs commands in a subshell. Passing DATA_CACHE explicitly into the test command ensures reliability and makes the behavior consistent both in Jenkins and when running the Makefile locally.

PR #1916, opened 2025-04-09
Skipping falcon rope scaling test (#221)

karol-brejna-i:
Unskipping revealed that SW-209093 is still not fixed.
This PR skips a test that has no chance of passing (due to known and not fixed bug).


regisss (approved):
LGTM

PR #1919, opened 2025-04-10
fix-disable-tensor-cache(Auto pr eb769a1)

TANA-BHU:
set DISABLE_TENOR_CACHE to run_pipeline.py (#184)
PR: https://github.com/habana-internal/optimum-habana-fork/pull/184


mandy-li (commented):
what is the reason to disable tensor cache in the script? and why hard-code max_graphs?

Issue #227, opened 2023-05-05
Fine-tuning BERT model without Trainer


shoang22:
Hello,

I have a custom model that I've incorporated BERT into. Is it possible to train this model using a normal training loop?

Example:
```
def training_loop(dataloader, model1):
    device = torch.device('hpu')
    model1 = model1.to(device)
    model2 = AutoModel.from_pretrained('bert-base-uncased').to(device)
    custom_model = some_wrapper(model1, model2)
    for batch in dataloader:
        batch = batch.to(device)
        output = custom_model(batch)
    
    ...
```

regisss:
Hi @shoang22! You would need to follow [Habana's SDK documentation](https://docs.habana.ai/en/latest/PyTorch/Getting_Started_with_PyTorch_and_Gaudi/Getting_Started_with_PyTorch.html) to have a properly working training loop. But then you would also need to add mixed precision, distributed training and so on to make the most of your Gaudi instance. This is all taken care of in the `GaudiTrainer` provided in the Optimum Habana library. Do you have any constraint that prevents you from using it?

shoang22:
I was actually having some issues with Trainer, so I opted for a direct implementation. I've since resolved the issue, and the Trainer is working perfectly. Thanks

PR #1923, opened 2025-04-11
Limit max pos embeds to 8k to prevent OOM (#227)

jaygala223:
# What does this PR do?

This PR caps the max pos embeds limit to 8k for the llama family of models. Earlier this check was for just one llama 3 model with a specific config. Now it applies to all llama family models and can be extended to others in the future.

Capping the max pos embeds for models in which we dont support higher context len saves memory and prevents OOM in bigger models like llama 3.1 405B

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -437,6 +437,9 @@ def setup_distributed_model_ep(args, model_dtype, model_kwargs, logger):
 def setup_distributed_model(args, model_dtype, model_kwargs, logger):
     import deepspeed
 
+    # List of model types that need max position embeddings capped at 8192
+    MODELS_WITH_POS_EMBEDDING_LIMIT = ["llama"]
+
     logger.info("DeepSpeed is enabled.")
     deepspeed.init_distributed(dist_backend="hccl")
     config = AutoConfig.from_pretrained(args.model_name_or_path, torch_dtype=model_dtype, **model_kwargs)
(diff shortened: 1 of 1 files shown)


regisss:
Hmm I'm thinking if we should not do that in https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/models/llama/configuration_llama.py, WDYT?
Since it should apply to all Llama models, I think it would make sense to move this piece of code in the init of the config.

examples/text-generation/utils.py
```diff
@@ -451,9 +454,7 @@ def setup_distributed_model(args, model_dtype, model_kwargs, logger):
         # Construct model with fake meta tensors, later will be replaced on devices during ds-inference ckpt load
         with deepspeed.OnDevice(dtype=model_dtype, device="meta"):
             if (
-                hasattr(config, "rope_scaling")
-                and config.rope_scaling
-                and config.rope_scaling["rope_type"] == "llama3"
+                any(model_type in args.model_name_or_path.lower() for model_type in MODELS_WITH_POS_EMBEDDING_LIMIT)
```
regisss, line 457:
```suggestion
                any(model_type == config.model_type for model_type in MODELS_WITH_POS_EMBEDDING_LIMIT)
```
This will be more generic as some Llama models may not have "llama" in their name.

jaygala223: Okay. There are also some llama models which have a different model_type like mllama or mllama_text_model. Can we run those via text-generation?

image
image

regisss: Nope, mllama can be used with the image-to-text example: https://github.com/huggingface/optimum-habana/tree/main/examples/image-to-text

jaygala223: Okay then it makes sense to use config.model_type. I will commit your suggestion. Thanks!

regisss (approved):
LGTM

PR #1925, opened 2025-04-11
Update makefile for explicit lazy mode

jasi306:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ DEFAULT_CLONE_URL := https://github.com/huggingface/optimum-habana.git
 # If CLONE_URL is empty, revert to DEFAULT_CLONE_URL
 REAL_CLONE_URL = $(if $(CLONE_URL),$(CLONE_URL),$(DEFAULT_CLONE_URL))
 
+export PT_HPU_LAZY_MODE=1
+# will be removed when lazy is disabled
 
 .PHONY:	style test
 


Makefile
```diff
@@ -17,6 +17,8 @@ DEFAULT_CLONE_URL := https://github.com/huggingface/optimum-habana.git
 # If CLONE_URL is empty, revert to DEFAULT_CLONE_URL
 REAL_CLONE_URL = $(if $(CLONE_URL),$(CLONE_URL),$(DEFAULT_CLONE_URL))
 
+export PT_HPU_LAZY_MODE=1
+# will be removed when lazy is disabled
```
regisss, line 21: Hmm not sure we want to do that for any command defined in this file. For example, I wouldn't expect `make style` to export such an environment variable.
Is the logic here to make lazy the default when launching tests?

libinta: yes. Do you have other suggestion?

libinta: @regisss do you mean to update files such as https://github.com/huggingface/optimum-habana/blob/main/tests/ci/slow_tests_1x.sh to include the env?

regisss: Yes, or just putting `PT_HPU_LAZY_MODE=1` at the front of all pytest commands in the Makefile

regisss (approved):

PR #1928, opened 2025-04-11
fix MPT inference inaccuracy output

astachowiczhabana:
call nn.functional.softmax directory instead of class Softmax()


regisss:
I think this is a duplicate of #1857

astachowiczhabana:
yes, closing

PR #1932, opened 2025-04-14
Workaround for DS issue in Llama

ugolowic:
DeepSpeed module doesn't handle keyword arguments and throws a RuntimeError when OH calls it with kwargs. This is a workaround for this issue in Llama.

### DIFF
--- a/optimum/habana/transformers/models/llama/modeling_llama.py
+++ b/optimum/habana/transformers/models/llama/modeling_llama.py
@@ -1349,28 +1349,34 @@ class GaudiLlamaModel(LlamaModel):
                 )
                 hidden_states = layer_outputs[0]
             else:
+                # Calling the layer with positional arguments
+                # This is a workaround for an issue with DeepSpeed where
+                # it cannot handle keyword arguments and throws a RuntimError
                 use_prev_layer_residual = attn_batch_split > 1 and past_key_values is None
+                layer_prev_layer_residual = prev_layer_residual if use_prev_layer_residual else None
+                layer_hidden_states = hidden_states_split if split_prompt else hidden_states
+                past_key_value = None if past_key_values is None else past_key_values[layer_idx]
                 layer_outputs = decoder_layer(
-                    hidden_states=hidden_states_split if split_prompt else hidden_states,
-                    attention_mask=causal_mask,
-                    position_ids=position_ids,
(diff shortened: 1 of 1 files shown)


yafshar:
@ugolowic Could you please add the example that causes the issue to the PR README? This will help us clearly identify the problem and provide a solution if we encounter the same issue with other models.

ugolowic:
@yafshar 
`MASTER_ADDR=localhost MASTER_PORT=12345 PT_HPU_LAZY_MODE=0   python3 examples/gaudi_spawn.py   --world_size 8 --use_deepspeed examples/language-modeling/run_lora_clm.py --dataset_name tatsu-lab/alpaca --bf16 False --dataset_concatenation --max_seq_len 4096 --validation_split_percentage 4 --save_strategy no --warmup_ratio 0.03 --lr_scheduler_type "cosine" --logging_steps 10 --do_train --use_habana --throughput_warmup_steps 3 --use_flash_attention True --flash_attention_causal_mask True --num_train_epochs 1 --eval_delay 2 --lora_rank 4 --lora_target_modules "q_proj" "v_proj" "k_proj" "o_proj" --per_device_eval_batch_size 1 --attn_softmax_bf16 True --evaluation_strategy epoch --dataloader_num_workers 4 --pipelining_fwd_bwd --learning_rate 0.0018 --deepspeed examples/language-modeling/llama3_ds_zero3_4k.json --gradient_accumulation_steps 32 --flash_attention_recompute True --context_parallel_size 2 --per_device_train_batch_size 1 --use_zero3_leaf_promotion --cache_size_limit 256 --compile_dynamic 0 --model_name_or_path meta-llama/Llama-3.3-70B-Instruct --fp8 True --output_dir /tmp/llama2_70b_hf_g2 --minimize_memory --use_lazy_mode False --torch_compile_backend hpu_backend --torch_compile`

yafshar:
@ugolowic, using this changes makes the code order-dependent. I suggest to add the comment for those argument which have different naming.

yafshar (approved):
LGTM!

@regisss this PR is ready. Would you please do the final check!

regisss (approved):
LGTM

I guess this happens with other models too then, because they all use keyword arguments in the non-gradient-checkpointing case.

PR #1936, opened 2025-04-17
Reduce LM_Eval output size after upgrade

astachowiczhabana:
After upgrading to lm_eval latest version samples are logged by default, we don't want that to reduce logs sizes


astachowiczhabana:
This change is already on upstream, closing

PR #1946, opened 2025-04-23
Move bitsandbytes requirements from setup.py to bnb tests

ckvermaAI:
Instead of adding bitsandbytes (bnb) requirement to setup.py, move it to bnb test files (tests/test_bnb_qlora.py, tests/test_bnb_inference.py). 

Details:
1) Because for HPU, bitsandbytes loads the CPU binaries (https://github.com/bitsandbytes-foundation/bitsandbytes/blob/multi-backend-refactor/bitsandbytes/cextension.py#L73), which are not required. Failure to load these binaries throws OSError (OSError: /usr/local/lib/python3.10/dist-packages/bitsandbytes/libbitsandbytes_cpu.so: cannot open shared object file: No such file or directory) for other tests.
2) For now, we move the BNB installation to the above-specified tests. And **both of these tests passed**.
3) Later, we'll try to upstream the fix to the bitsandbytes repo.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,6 @@ TESTS_REQUIRE = [
     "torchsde",
     "timm",
     "peft",
-    "bitsandbytes @ git+https://github.com/bitsandbytes-foundation/bitsandbytes.git@multi-backend-refactor",
 ]
 
 QUALITY_REQUIRES = [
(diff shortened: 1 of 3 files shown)


vivekgoe (approved):

regisss (approved):
LGTM

@ckvermaAI Can you also merge the latest main branch into yours please? The runners we use for CI have changed and the Transformers and Diffusers tests won't run without this.

PR #1941, opened 2025-04-21
Pad the examples for QLoRa finetuning test

ckvermaAI:
1. Pad the examples up to max_seq_len of 1024
2. Increase the max_steps (from 5 to 50) and eval_steps (from 3 to 10). Set throughput-related arguments (adjust_throughput, throughput_warmup_steps)
4. Update the fine-tuning test name
5. Also, adjust the reference "eval loss" value for fine-tuning test and "output" for inference test.

Additional updates
6. Enable the eager mode for the test (disable the torch.compile mode for now).
7. Add new requirement for installing the bitsandbytes (from https://github.com/bitsandbytes-foundation/bitsandbytes/tree/multi-backend-refactor)

PR #1951, opened 2025-04-25
Bitsandbytes installation for qlora tests

ckvermaAI:
Move bitsandbytes installation (required for qlora tests: tests/test_bnb_qlora.py, tests/test_bnb_inference.py) to conftest.py

Details:
1) pytest runs the pytest_configure function before executing the test itself (https://github.com/huggingface/optimum-habana/blob/main/conftest.py#L118)
2) Now, this function imports optimum.habana, which imports transformers.utils, which checks whether specific packages (for example: torch, peft, bitsandbytes) are installed (and transformers saves this state). 
3) Bitsandbytes is installed later (during the test execution) because the transformer has not updated its state, and the QLORA tests fail for the above reason. 
4) This issue does not occur in internal (OH-fork) repo because of the absence of the pytest_configure function from conftest.py file

5) Currently, we cannot add bitsandbytes to the requirements of tests in setup.py as discussed earlier.(https://github.com/huggingface/optimum-habana/pull/1941#issuecomment-2823330870). Once we upstream the a proper fix to bitsandbytes repo, we'll revert this change in conftest.py

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -113,6 +113,22 @@ def token(request):
 
 
 def pytest_configure(config):
+    # Bitsandbytes installation for {test_bnb_qlora.py test_bnb_inference.py} tests
+    # This change will be reverted shortly
+    bnb_tests = any("bnb" in name for name in config.known_args_namespace.file_or_dir)
+    if bnb_tests:
+        import subprocess
+        import sys
+
+        subprocess.check_call(
+            [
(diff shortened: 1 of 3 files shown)


ckvermaAI:
@regisss @libinta Please review this change (details are available in description)

regisss (approved):
LGTM

PR #1954, opened 2025-04-28
add average latency metrics

RongLei-intel:
# What does this PR do?
for example/text-generation/run_generation.py:
Add TTFT/TPOT/E2E latency output for each generation, and average latency at the end of output and result.

Fixes # https://github.com/huggingface/optimum-habana/issues/1952 (issue)


regisss (approved):
Nice, LGTM!

PR #1931, opened 2025-04-14
Profiling improvements

ugolowic:
* Refactor HabanaProfile so that it can be used more than once
* Add option for the user to collect profiling in GaudiTrainer's eval
* In text-generation, user can now select whether a profiling step means generating one token or one sequence
* Added integration and unit tests for HabanaProfile

PR #1958, opened 2025-04-28
enable profiling for video-comprehension sample

alexey-belyakov:
# What does this PR do?

Enable profiling for video-comprehension sample. I needed to collect profiling results for performance degradation issue, to do it
- I've added profiling arguments as it works in text-generation/run_generation.py
- To avoid code duplication I've added common_parser.py and moved profiling options to this file.
Some command lines to test:
python examples/video-comprehension/run_example.py --help
python examples/text-generation/run_generation.py --help
python examples/video-comprehension/run_example.py --model_name_or_path LanguageBind/Video-LLaVA-7B-hf --bf16 --use_hpu_graphs --output_dir /tmp/tmpssf594uz --profiling_steps 10 --profiling_warmup_steps 3 --profiling_record_shapes

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -23,6 +23,7 @@ import json
 import logging
 import math
 import os
+import sys
 from itertools import cycle
 from pathlib import Path
 
@@ -38,6 +39,12 @@ from utils import (
 )
 
 from optimum.habana.utils import HabanaGenerationTime, get_hpu_memory_stats
+
+
+project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../"))
+if project_root not in sys.path:
+    sys.path.insert(0, project_root)
+from examples.common_parser import add_profiling_args  # noqa: E402
 
 
 logging.basicConfig(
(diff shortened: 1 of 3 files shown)


examples/text-generation/run_generation.py
```diff
@@ -40,6 +41,12 @@
 from optimum.habana.utils import HabanaGenerationTime, get_hpu_memory_stats
 
 
+project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../"))
+if project_root not in sys.path:
+    sys.path.insert(0, project_root)
```
regisss, line 46: Why do we need these lines?

examples/video-comprehension/run_example.py
```diff
@@ -32,6 +33,12 @@
 )
 
 
+project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../"))
+if project_root not in sys.path:
+    sys.path.insert(0, project_root)
```
regisss, line 38: Same

examples/text-generation/run_generation.py
alexey-belyakov: @regisss can you check my comment please

12010486: @alexey-belyakov, I would suggest to wait for https://github.com/huggingface/optimum-habana/pull/1931 to be merged as there is some rebase of profiling code. 

Besides that, probably another solution in case you do not want to rewrite code multiple times is to create a file like examples_utils.py, or the like, under optimum/habana which is already in the path. To avoid the solution you provided which is not very clear.

12010486: Honestly I'm also fine in not creating a new file for the three profiling args that are repeating : ) But I can understand the pain

regisss: I think that solution is okay. Can you still check if it still holds after #1931 as suggested by @12010486 please?

PR #1961, opened 2025-04-30
Added torch.compile WA to disable VAE decode func

kplau1128:
Added torch.compile WA to disable VAE decode func


kplau1128:
No longer need disable vae decode. Replace with #2000

PR #1982, opened 2025-05-17
Switch to Spawn in PyTorch DataLoader when num_worker>0

Wei-Lin-Intel:
# What does this PR do?

In [SW-207456](https://jira.habana-labs.com/browse/SW-207456) , it is found the multi-node training would crash at the begining when `dataloader_num_workers` is set to larger than 0.

According to the [official habana document](https://docs.habana.ai/en/latest/PyTorch/Getting_Started_with_PyTorch_and_Gaudi/Getting_Started_with_PyTorch.html?highlight=spawn#torch-multiprocessing-for-dataloaders), it is required to change the start method to spawn or forkserver using the PyTorch API. Thus this PR updates the API for Accelerate Dataloader.

Fixes # (issue)
[SW-207456](https://jira.habana-labs.com/browse/SW-207456) 

## Before submitting

### DIFF
--- a/optimum/habana/accelerate/accelerator.py
+++ b/optimum/habana/accelerate/accelerator.py
@@ -682,6 +682,7 @@ class GaudiAccelerator(Accelerator):
         # To avoid training crash issue SW-207456 when num_worker > 0 in multi-node training tasks
         if data_loader.num_workers > 0:
             import multiprocessing
+
             multiprocessing_context = multiprocessing.get_context("spawn")
             data_loader.multiprocessing_context = multiprocessing_context
 
--- a/optimum/habana/accelerate/accelerator.py
+++ b/optimum/habana/accelerate/accelerator.py
@@ -680,7 +680,7 @@ class GaudiAccelerator(Accelerator):
         ###############################################################################################################
 
         # To avoid training crash issue SW-207456 when num_worker > 0 in multi-node training tasks
-        if int(os.environ["WORLD_SIZE"]) > 8 and data_loader.num_workers > 0:
+        if int(os.environ.get("WORLD_SIZE", 1)) > 8 and data_loader.num_workers > 0:
             import multiprocessing
 
             multiprocessing_context = multiprocessing.get_context("spawn")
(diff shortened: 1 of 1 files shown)


Wei-Lin-Intel:
@libinta  @regisss  Please review this PR as this is a bug fix. Thanks.

optimum/habana/accelerate/accelerator.py
```diff
@@ -679,6 +679,13 @@ def prepare_data_loader(
             process_index = int(process_index / parallel_state.get_sequence_parallel_world_size())
         ###############################################################################################################
 
+        # To avoid training crash issue SW-207456 when num_worker > 0 in multi-node training tasks
+        if data_loader.num_workers > 0:
```
regisss, line 683:
```suggestion
        if os.environ["LOCAL_WORLD_SIZE"] != os.environ["WORLD_SIZE"] and data_loader.num_workers > 0:
```
Let's do this only for the multi-node case.

Wei-Lin-Intel: The condition has been modified as:
`if int(os.environ["WORLD_SIZE"]) > 8 and data_loader.num_workers > 0:`

Because if `LOCAL_WORLD_SIZE` is not defined in the training task, then this change would throw error (verified). Customers don't like to add more flags.

Since for the multi-node training tasks the `WORLD_SIZE` must be larger than 8, so `int(os.environ["WORLD_SIZE"]) > 8` should be good to filter the single node case.

@regisss  Please review this change, thanks.

regisss: `LOCAL_WORLD_SIZE` should be automatically set by Torch, I thought it would work :thinking: 
It seems I tried to be too fancy here.

Wei-Lin-Intel: No, I tried in container 1.21.0-554 (PyTorch 2.6.0) and it throw error saying "NoneType" for `LOCAL_WORLD_SIZE`. If we must do in this way, a. Customer should define `LOCAL_WORLD_SIZE`, b. we should add more condition to judge the existence of `LOCAL_WORLD_SIZE` flag. That would be more compliated...

regisss (approved):
LGTM

PR #1984, opened 2025-05-19
fix wrong calculation of e2e latency

RongLei-intel:
# What does this PR do?
This PR corrects the calculation of the e2e latency that was previously implemented incorrectly. The fix ensures accurate reporting of latency metrics for e2e. 

This addresses the issue discussed in #1952


regisss (approved):
LGTM

PR #1993, opened 2025-05-22
force model preparation on eval if torch.compile enabled

Luca-Calabria:
Running training example as torch.compile() mode, the evaluation step skips the model compilation reporting wrong metrics.

To fix it, we force the accelerator model preparation for both training ed evaluation steps.

### DIFF
--- a/optimum/habana/transformers/trainer.py
+++ b/optimum/habana/transformers/trainer.py
@@ -1874,7 +1874,7 @@ class GaudiTrainer(Trainer):
 
         model = self._wrap_model(self.model, training=False, dataloader=dataloader)
 
-        if len(self.accelerator._models) == 0 and model is self.model:
+        if (len(self.accelerator._models) == 0 or self.args.torch_compile) and model is self.model:
             start_time = time.time()
             model = (
                 self.accelerator.prepare(model)


Luca-Calabria:
converted to draft because the fix fails running torch.compile + --use_regional_compilation

12010486:
Example of command line that is failing is

```
PT_TE_CUSTOM_OP=1 PT_HPU_LAZY_MODE=0 python ../gaudi_spawn.py     --world_size 8 --use_deepspeed run_lora_clm.py     --model_name_or_path meta-llama/Meta-Llama-3.1-8B     --dataset_name tatsu-lab/alpaca     --bf16 False     --output_dir ./model_lora_llama_8B     --num_train_epochs 3     --per_device_train_batch_size 1     --gradient_accumulation_steps 16     --eval_strategy "no"     --save_strategy "no"     --learning_rate 3e-4     --warmup_ratio 0.03     --lr_scheduler_type "constant"     --max_grad_norm 1.0     --logging_steps 10     --do_train     --do_eval     --use_habana     --use_lazy_mode False     --throughput_warmup_steps 3     --lora_rank=8     --lora_alpha=16     --lora_dropout=0.05     --lora_target_modules "q_proj" "v_proj"     --dataset_concatenation     --max_seq_length 4096     --adam_epsilon 1e-08     --validation_split_percentage 4     --deepspeed llama3_ds_zero1_config.json     --torch_compile_backend hpu_backend     --torch_compile     --fp8     --use_flash_attention True     --flash_attention_causal_mask True      --per_device_eval_batch_size 4      --cache_size_limit 64     --use_regional_compilation     --compile_from_sec_iteration     --allow_unspec_int_on_nn_module True --token xxx
```

Luca-Calabria:
example of failing commandline:

PT_HPU_LAZY_MODE=0 PT_ENABLE_INT64_SUPPORT=1 python run_qa.py \\
--model_name_or_path bert-large-uncased-whole-word-masking \\
--gaudi_config_name Habana/bert-large-uncased-whole-word-masking \\
--dataset_name squad \\
--do_train \\
--do_eval \\
--per_device_train_batch_size 32 \\
--per_device_eval_batch_size 8 \\
--learning_rate 3e-5 \\
--num_train_epochs 2 \\
--max_seq_length 384 \\
--doc_stride 128 \\
--output_dir /tmp/squad1/ \\
--use_habana \\
--torch_compile_backend hpu_backend \\
--torch_compile \\
--use_lazy_mode false \\
--throughput_warmup_steps 3 \\
--bf16 \\
--sdp_on_bf16 \\
--use_regional_compilation

yafshar:
@Luca-Calabria Sync this PR with main, please.

Luca-Calabria:
Last fix works also for --use_regional_compilation and doesn't produce regressions between "train+eval" and "just eval" run

Luca-Calabria:
This PR want to fix the regression on torch.compile without --use_regional_compilation whenyou run these two commands:
=== Train+Eval example
PT_HPU_LAZY_MODE=0 PT_ENABLE_INT64_SUPPORT=1 python run_qa.py \
--model_name_or_path bert-large-uncased-whole-word-masking \
--gaudi_config_name Habana/bert-large-uncased-whole-word-masking \
--dataset_name squad \
--do_train \
--do_eval \
--per_device_train_batch_size 32 \
--per_device_eval_batch_size 8 \
--learning_rate 3e-5 \
--num_train_epochs 2 \
--max_seq_length 384 \
--doc_stride 128 \
--output_dir /tmp/squad1/ \
--use_habana \
--torch_compile_backend hpu_backend \
--torch_compile \
--use_lazy_mode false \
--throughput_warmup_steps 3 \
--bf16 \
--sdp_on_bf16

=== just Eval example
PT_HPU_LAZY_MODE=0 PT_ENABLE_INT64_SUPPORT=1 python run_qa.py \
--model_name_or_path bert-large-uncased-whole-word-masking \
--gaudi_config_name Habana/bert-large-uncased-whole-word-masking \
--dataset_name squad \
--do_eval \
--per_device_eval_batch_size 8 \
--max_seq_length 384 \
--doc_stride 128 \
--output_dir /tmp/squad1/ \
--use_habana \
--torch_compile_backend hpu_backend \
--torch_compile \
--use_lazy_mode false \
--throughput_warmup_steps 3 \
--bf16 \
--sdp_on_bf16

BUT my first fix broken the regional compilation feature. The new fix keep working w/o regional compilation and remove the regression.

yafshar:
@Luca-Calabria I understand the regression, but I do not think your fix is correct.

yafshar:
I think if f you want to use a TorchDynamo-compiled model during evaluation, you would need to manually compile the model using torch.compile() before passing it to the Trainer

Luca-Calabria:
This portion of code on accelerator [optimum-habana/optimum/habana/accelerate/accelerator.py at main · huggingface/optimum-habana](https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/accelerate/accelerator.py#L368-L373) 

if self.use_regional_compilation:
  compile_regions(model, compile_kwargs)
else:
  model = torch.compile(model, **compile_kwargs)

show two different behaviors:
compile_regions modify the trainer instance of self.model directly
torch.compile() modify a copy of this self.model

This is why in the evaluation_loop this line [optimum-habana/optimum/habana/transformers/trainer.py at main · huggingface/optimum-habana](https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/transformers/trainer.py#L1875) 
works when run with --use_regional_compilation (because the model is already prepared and can be skipped) but doesn’t work (produce regression) when run torch.compile() without --use_regional_compilation (because it skips the model preparation).

A solution for that could be something like this:

if self.use_regional_compilation:
   import copy
   model_compiled = copy.deepcopy(model)
   compile_regions(model_compiled, compile_kwargs)
else:
   model_compiled = torch.compile(model, **compile_kwargs)

@IlyasMoutawwakil I see you fixed a similar issue about regional_compilation here https://github.com/huggingface/optimum-habana/pull/2014. Is your fix just for deepspeed initialization? or it works also for single card training?

optimum/habana/accelerate/accelerator.py
```diff
@@ -367,10 +368,12 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
             compile_kwargs = self.state.dynamo_plugin.to_kwargs()
             ############################################################################################################
             if self.use_regional_compilation:
-                compile_regions(model, compile_kwargs)
+                model_compiled = copy.deepcopy(model)
```
IlyasMoutawwakil: you do realize this a 2x memory consumption ?

Luca-Calabria: yep I'm going to change this using "compile_regions" from _accelerate.utils.compile_regions_ like you did in your PR

optimum/habana/transformers/trainer.py
```diff
@@ -1874,7 +1874,7 @@ def evaluation_loop(
 
         model = self._wrap_model(self.model, training=False, dataloader=dataloader)
 
-        if len(self.accelerator._models) == 0 and model is self.model:
+        if (len(self.accelerator._models) == 0 or self.args.torch_compile) and model is self.model:
```
IlyasMoutawwakil: I don't understand the point here, if `len(self.accelerator._models) == 0` the model will be prepared.
why re-prepare it if torch.compile ?

optimum/habana/accelerate/accelerator.py
```diff
@@ -67,24 +68,6 @@
...
-    else:
-        if model._modules:  # If model has submodules, recurse and reassign
-            for name, module in model.named_children():
-                compiled_module = compile_regions(module, compile_kwargs)
-                if compiled_module is not None:  # Only reassign if something is returned
-                    setattr(model, name, compiled_module)
-        else:  # Leaf node
-            model = torch.compile(model, **compile_kwargs)
-            model.__dict__.pop("_parameters", None)
-            return model
```
IlyasMoutawwakil: Nope we can't use the current compile_regions from accelerate as it is an out-of-place operation that breaks deepspeed engine. See https://github.com/huggingface/accelerate/pull/3609

optimum/habana/transformers/trainer.py
Luca-Calabria: It is explained in the description of this PR: when you run "Train+eval" example, after the train step this condition is false **len(self.accelerator._models) == 0** because the accelerator has the **_models** list populated with the current model; this means that when evaluation_loop is called, the model preparation is skipped but the model actually is not compiled. 
when you run "just eval" example it works because the accelerator ha **_models** list empty.

Should I force the cleanup of accelerator._model list after train step maybe?

optimum/habana/accelerate/accelerator.py
Luca-Calabria: hmm ok, I was just following your PR https://github.com/huggingface/optimum-habana/pull/2014

optimum/habana/transformers/trainer.py
IlyasMoutawwakil: > this means that when evaluation_loop is called, the model preparation is skipped but the model actually is not compiled.

I mean yes that's to be expected, compilation should only be done once. and the returned model will be compiled for the rest of its life (except at checkpoint save time where its unwrapped to be saved)

IlyasMoutawwakil: sorry if it's self evident but I'm asking because if what you're saying is true I'm wondering why it wasn't flagged in transformers before 😅 sounds like a universal issue

Luca-Calabria: @yafshar this is different from what you described as expected flow. Maybe I misunderstood?

optimum/habana/accelerate/accelerator.py
IlyasMoutawwakil: I would advised not to 😅 it's just an attempt at enabling out-of-place compilation with deepspeed, a failed one to be specific.

Luca-Calabria: :) now it is pretty clear!!!

optimum/habana/transformers/trainer.py
Luca-Calabria: Just to be clear, the expected flow is:
1) load original model
2) compile it and run the training
3) use compiled and trained model for evaluation

IlyasMoutawwakil: yes, it should be the same model

Luca-Calabria:
I updated to get **self.model = model** when run torch.compile(). Now during evaluation it will use the model compiled in training.

optimum/habana/transformers/trainer.py
IlyasMoutawwakil: there must be a misunderstanding. for me the "expected flow" is what's done in transformers, the source of truth, now if you're saying that we can do better, then it needs to be tested across the entire OH lib 😅 because unfortunately changes like these tend to break things we didn't even think about. also it needs to show a clear use case / optimization. I'm having a hard time with this PR because I feel like it doesn't clearly state what it's fixing or optimizing (those are two different things).

IlyasMoutawwakil: please take the time to explain what is being done and how it's not correct or misaligned with expectations.

IlyasMoutawwakil: if we agree to make this change, I will prefer it be submitted to transformers as well, where a maintainer from there who has a broader view of the impacts of this on the Trainer class, can make an informed comment on how to implement this behavior best. this way we avoid diverging more from the transformers trainer.

IlyasMoutawwakil: also, note that model preparation is skipped in evaluation for a reason (you can't run .generate with a pipeline parallel model for example). That's why it doesn't make sense to propagate the effects of model preparation to the evaluation phase simply for torch.compile benefits ; it makes more sense in this case is to explicitly compile the model in the evaluation loop, if that's what we're after.

yafshar: That is what I mentioned to Luca. The correct approach is to explicitly compile the model in the evaluation loop

> I think if f you want to use a TorchDynamo-compiled model during evaluation, you would need to manually compile the model using torch.compile() before passing it to the Trainer

Luca-Calabria: I removed this solution anyway. Please take a look at the new, and less intrusive, proposed solution I commit.

IlyasMoutawwakil: it still is intrusive 😅 for example fsdp+ torch.compile will result in the evaluation loop running with an FSDP wrapped model which I believe doesn't even have a .generate method 🤔

Luca-Calabria: could you share please in which part of the code you see this?

IlyasMoutawwakil: the line `self.model = model` propagates the effects of accelerator.prepare which wraps the modl in all kinds of distributed training wrappers. using this wrapped model during evaluation will break it in some scenarios.

IlyasMoutawwakil: it is a single small line change but the impacts are gigantic

Luca-Calabria:
@IlyasMoutawwakil I double checked the transformers repo on nvidia here https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L4394 and it seems to work running torch.compile() reporting on both cases "train+eval" and "just eval" the same eval_samples_per_second value. The odd thing is that **model**, in case of train+eval, is not the compiled one 
`model = BertForQuestionAnswering(
  (bert): BertModel(
    (embeddings): BertEmbeddings(
      (word_embeddings): Embedding(30522, 1024, padding_idx=0)
      (position_embeddings): Embedding(512, 1024)
      (token_type_embeddings): Embedding(2, 1024)
      (LayerNorm): LayerNorm((1024,), eps=1e-12, elementwise_affine=True)
      (dropout): Dropout(p=0.1, inplace=False)
    )
    (encoder): BertEncoder(
      (layer): ModuleList(
        (0-23): 24 x BertLayer(
          (attention): BertAttention(
            (self): BertSdpaSelfAttention(
              (query): Linear(in_features=1024, out_features=1024, bias=True)
              (key): Linear(in_features=1024, out_features=1024, bias=True)
              (value): Linear(in_features=1024, out_features=1024, bias=True)
              (dropout): Dropout(p=0.1, inplace=False)
            )
            (output): BertSelfOutput(
              (dense): Linear(in_features=1024, out_features=1024, bias=True)
              (LayerNorm): LayerNorm((1024,), eps=1e-12, elementwise_affine=True)
              (dropout): Dropout(p=0.1, inplace=False)
            )
          )
          (intermediate): BertIntermediate(
            (dense): Linear(in_features=1024, out_features=4096, bias=True)
            (intermediate_act_fn): GELUActivation()
          )
          (output): BertOutput(
            (dense): Linear(in_features=4096, out_features=1024, bias=True)
            (LayerNorm): LayerNorm((1024,), eps=1e-12, elementwise_affine=True)
            (dropout): Dropout(p=0.1, inplace=False)
          )
        )
      )
    )
  )
  (qa_outputs): Linear(in_features=1024, out_features=2, bias=True)`

( the same for **self.model**) but **self.model_wrapped** is compiled and it seems not used by **self.prediction_step**. Running "just eval" case it prepare the model here https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L4394 and **model** is the compiled one
`model = OptimizedModule(
  (_orig_mod): BertForQuestionAnswering(
    (bert): BertModel(
      (embeddings): BertEmbeddings(
        (word_embeddings): Embedding(30522, 1024, padding_idx=0)
        (position_embeddings): Embedding(512, 1024)
        (token_type_embeddings): Embedding(2, 1024)
        (LayerNorm): LayerNorm((1024,), eps=1e-12, elementwise_affine=True)
        (dropout): Dropout(p=0.1, inplace=False)
      )
      (encoder): BertEncoder(
        (layer): ModuleList(
          (0-23): 24 x BertLayer(
            (attention): BertAttention(
              (self): BertSdpaSelfAttention(
                (query): Linear(in_features=1024, out_features=1024, bias=True)
                (key): Linear(in_features=1024, out_features=1024, bias=True)
                (value): Linear(in_features=1024, out_features=1024, bias=True)
                (dropout): Dropout(p=0.1, inplace=False)
              )
              (output): BertSelfOutput(
                (dense): Linear(in_features=1024, out_features=1024, bias=True)
                (LayerNorm): LayerNorm((1024,), eps=1e-12, elementwise_affine=True)
                (dropout): Dropout(p=0.1, inplace=False)
              )
            )
            (intermediate): BertIntermediate(
              (dense): Linear(in_features=1024, out_features=4096, bias=True)
              (intermediate_act_fn): GELUActivation()
            )
            (output): BertOutput(
              (dense): Linear(in_features=4096, out_features=1024, bias=True)
              (LayerNorm): LayerNorm((1024,), eps=1e-12, elementwise_affine=True)
              (dropout): Dropout(p=0.1, inplace=False)
            )
          )
        )
      )
    )
    (qa_outputs): Linear(in_features=1024, out_features=2, bias=True)
  )
)`

Not c;lear how it can produce the same performance.

Luca-Calabria:
Hi @yafshar @IlyasMoutawwakil I double checked how the original transformers repo works on NVidia and I found the same Gaudi behavior running with 
image

The mismatch between "train+eval" and "just eval" on eval metrics can be reproduced also running on cudagraph backend. 
image
and
image

After this result I think we can close (no merging) this PR as no further action is required on our end. What  do you think?

PR #2010, opened 2025-05-28
Use the original TE FP8 defaults

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/accelerate/accelerator.py
+++ b/optimum/habana/accelerate/accelerator.py
@@ -18,9 +18,10 @@ from __future__ import annotations
 import functools
 import math
 import os
-from dataclasses import make_dataclass
+from dataclasses import dataclass, make_dataclass
 from types import MethodType
 
+import accelerate.utils.dataclasses
 import torch
 from accelerate import Accelerator
 from accelerate.accelerator import _split_batches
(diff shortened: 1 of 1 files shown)


vivekgoe (approved):
LGTM.

vivekgoe:
@IlyasMoutawwakil In my opinion moving these defaults to model gaudi_config is a good idea. Upstream to Accelerate is not needed, these defaults are not necessarily the best ones, these are something which just worked fine for 1 particular fine-tuning configuration we were using to add Transformer Engine support.

optimum/habana/accelerate/accelerator.py
```diff
@@ -143,6 +187,10 @@ def __init__(
             deepspeed_plugins=deepspeed_plugins,
         )
 
+        if self.state.mixed_precision == "fp8":
```
vivekgoe: @IlyasMoutawwakil There was some additional code to set mixed_precision == "fp8" when we pass --fp8 as a Gaudi specific training argument. In case we have deleted that then we would need to bring that back as well for this change to work.

IlyasMoutawwakil: tell me if the change I made, which picks up the env var before initializing the accelerator fixes it (accelerate seems to ignore the env var in some cases).

PR #2011, opened 2025-05-30
Enable FusedRMSNorm for FLUX

dsocek:
# What does this PR do?

Enables FusedRMSNorm instead of vanilla pytorch one, boosting the performance of FLUX model.


libinta:
@dsocek  accuracy looks good for this change?

PR #2014, opened 2025-06-04
Fix deepspeed regional compilation by doing it before engine initialization

IlyasMoutawwakil:
# What does this PR do?

This avoids loosing any hooks or custom offloaded parameters added by the deepspeed engine wrapper.
@yafshar I think this makes more sense, allowing easy decompilation when saving the model, hope we can get it tested extensively !

## Before submitting

### DIFF
--- a/optimum/habana/accelerate/accelerator.py
+++ b/optimum/habana/accelerate/accelerator.py
@@ -565,10 +565,11 @@ class GaudiAccelerator(Accelerator):
             engine, optimizer, _, lr_scheduler = deepspeed.initialize(**kwargs)
 
             if self.state.dynamo_plugin.backend != DynamoBackend.NO and not self.use_regional_compilation:
+                compile_kwargs = self.state.dynamo_plugin.to_kwargs()
                 # deepspeed native compilation is done after deepspeed initialization
                 engine.compile(
                     backend=compile_kwargs.pop("backend"),
-                    compile_kwargs=self.state.dynamo_plugin.to_kwargs(),
+                    compile_kwargs=compile_kwargs,
                     compiled_autograd_enabled=self.compiled_autograd_enable,
                 )
 


IlyasMoutawwakil:
okay I just checked the model and it is decompiled when wrapped by the engine for some reason...

optimum/habana/accelerate/accelerator.py
```diff
@@ -575,20 +556,22 @@ def _prepare_deepspeed(self, *args):
...
-                ###############################################################################################################
+                model = compile_regions(model, **compile_kwargs)
+
+            engine, optimizer, _, lr_scheduler = deepspeed.initialize(**kwargs)
+
+            if self.state.dynamo_plugin.backend != DynamoBackend.NO and not self.use_regional_compilation:
+                # deepspeed native compilation is done after deepspeed initialization
+                engine.compile(
+                    backend=compile_kwargs.pop("backend"),
+                    compile_kwargs=self.state.dynamo_plugin.to_kwargs(),
```
yafshar:
```suggestion
                compile_kwargs = self.state.dynamo_plugin.to_kwargs()
                # deepspeed native compilation is done after deepspeed initialization
                engine.compile(
                    backend=compile_kwargs.pop("backend"),
                    compile_kwargs=compile_kwargs,
```

@IlyasMoutawwakil compile_kwargs is undefined in this block! the above block has a different condition!

yafshar: or maybe the condition should be updated!

IlyasMoutawwakil: instead of `compile_kwargs = self.state.dynamo_plugin.to_kwargs()` then `compile_kwargs=compile_kwargs `
I did `compile_kwargs=self.state.dynamo_plugin.to_kwargs()` what condition do you mean 😅

yafshar: exactly `backend=compile_kwargs.pop("backend"),`  the compile_kwargs is undefined in this if block

IlyasMoutawwakil: ah right ! thanks

yafshar: the above if block has `and self.use_regional_compilation:` when you define `compile_kwargs` and this block has `and not self.use_regional_compilation:`

PR #2019, opened 2025-06-06
[speech_recoginition] Fix datasets version 

gplutop7:


regisss:
@karol-brejna-i I think your cached version of this dataset is a bit old. The error `No module named 'datasets.tasks'` should not happen with the latest version, can you try with it?

gplutop7:
@regisss Thanks for the suggestion!

I’ve looked into it, and the error is not caused by a cached version — the datasets.tasks module (including AutomaticSpeechRecognition) has been officially deprecated and removed in recent versions of the datasets library. From what I remember, we were seeing this error (No module named 'datasets.tasks') while using version 3.6.0

Here are a few sources confirming this:
  1) Official docs (deprecated): https://huggingface.co/docs/datasets/main/en/package_reference/task_templates
  2) Commit removing it: https://huggingface.co/datasets/patrickvonplaten/librispeech_asr_self_contained/commit/7c25d0ae8a393c21533d85c1848c337d4fa1ca7b

To avoid this error, the recommendation is to either:
1) Downgrade to an older version (not ideal long-term), or
2) Refactor the dataset script to not rely on datasets.tasks, which is the direction I’ve taken.

Please let me know if you'd prefer a different approach! Always happy to adjust.

PR #2022, opened 2025-06-09
Enable fp8 based on ACCELERATE_MIXED_PRECISION

hlahkar:
This introduces the logic for enabling FP8 based on ACCELERATE_MIXED_PRECISION. The idea is taken from GaudiAcceleratorState. This is needed as self._mixed_precision is set to no; as self.distributed_type == DistributedType.DEEPSPEED.

### DIFF
--- a/optimum/habana/accelerate/accelerator.py
+++ b/optimum/habana/accelerate/accelerator.py
@@ -163,7 +163,7 @@ class GaudiAccelerator(Accelerator):
         self.distribution_strategy = distribution_strategy
         self.force_autocast = force_autocast
         self.mpu = parallel_state
-
+        self.is_fp8_enabled = os.environ.get("ACCELERATE_MIXED_PRECISION", "no") == "fp8"
         super().__init__(
             device_placement=device_placement,
             split_batches=split_batches,
@@ -187,7 +187,7 @@ class GaudiAccelerator(Accelerator):
             deepspeed_plugins=deepspeed_plugins,
         )
 
-        if self.state.mixed_precision == "fp8":
+        if self.is_fp8_enabled:
             if self.fp8_recipe_handler is None:
                 self.fp8_recipe_handler = get_fp8_recipe(GaudiTERecipeKwargs())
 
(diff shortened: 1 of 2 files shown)


hlahkar:
> thanks this makes sense m, it seems that there's a chance this is a global issue in accelerate, I tried to fix it in [huggingface/accelerate#3611](https://github.com/huggingface/accelerate/pull/3611) can you also check if my PR also fixes the issue.

@IlyasMoutawwakil I checked the PR https://github.com/huggingface/accelerate/pull/3611; this will not work mainly because of the following:

1. https://github.com/huggingface/accelerate/blob/main/src/accelerate/state.py#L945 - self._mixed_precision is set to 'no' as distributed_type is DistributedType.DEEPSPEED
2. Due to the above case, self.state.mixed_precision; will always return 'bf16', 'fp16' or 'no' in case of DistributedType.DEEPSPEED; you can refer to https://github.com/huggingface/accelerate/blob/main/src/accelerate/state.py#L1026

IlyasMoutawwakil:
hmm, so with habanaAI/deepspeed + fp8 we still want `mixed_precision` to be fp8 right ? which is not the case for other accelerators in accelerate (e.g. GPU)

optimum/habana/accelerate/accelerator.py
```diff
@@ -432,7 +432,7 @@ def _prepare_deepspeed(self, *args):
             self._prepare_one(obj, first_pass=True)
             if isinstance(obj, torch.utils.data.DataLoader)
             else convert_model(obj)
-            if isinstance(obj, torch.nn.Module) and self.state.mixed_precision == "fp8"
+            if isinstance(obj, torch.nn.Module) and self.is_fp8_enabled
```
IlyasMoutawwakil, line 435: this definitely breaks other things 😅 notice that the mixed_precision that was passed during accelerator and accelerator state initialization is not used at all ..

IlyasMoutawwakil: the only way to do fp8 with the proposed changes, would be by setting the env var ?

IlyasMoutawwakil:
there are some conflicts so I moved the changes (with a slight consideration for the argument mixed_precision) directly to my PR branch, thanks !

PR #2023, opened 2025-06-09
Fix device OOM caused by buffers not being deallocated fast enough

bkowalskiINTEL:
This PR is a fix for a reported Device OOM in Granite 8B model. The issue was caused by a big amount of ~1GB unnecessary buffers not being deallocated at the end of GaudiGenerationMixin._sample() calls, which made these buffers stay alive between GaudiGenerationMixin.generate() calls, further causing the OOM.

The solution turned out to be explicitly calling gc.collect() just before the return statements in the _sample() method, so that python's garbage collector frees the objects holding the buffers.

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import copy
+import gc
 import inspect
 import math
 import warnings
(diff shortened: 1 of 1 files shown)


bkowalskiINTEL:
@vivekgoe, could You please review this PR? Thanks in advance.

deepvars (approved):

bkowalskiINTEL:
I've been informed, that this PR should be merged into another branch, so I'll close this one an create a new one.

PR #2030, opened 2025-06-11
Refactor Qwen2 Family - FP32 SDPA and max_position_embedding

Wei-Lin-Intel:
# What does this PR do?

Fixes # (issue)
SW-216073

Qwen2 family have large magnitude of q*k so it is not proper to use BF16 or lower precision for softmax, so we should use fp32 softmax mode for inference.

Hence, since HF Transformers removed `self.max_position_embeddings = config.max_position_embeddings` in Qwen2 Attention, so it will cause issue in `update_sincos_cache` [(link)](https://github.com/huggingface/optimum-habana/blob/df7db95e47be58e39eba3ba73cf7a68f6e81c46a/optimum/habana/transformers/models/qwen2/modeling_qwen2.py#L338). We should add it to OH modeling to avoid potential crash issue.

## Before submitting

### DIFF
--- a/tests/baselines/fixture/tests/test_text_generation_example.json
+++ b/tests/baselines/fixture/tests/test_text_generation_example.json
@@ -12,7 +12,7 @@
       "throughput": 85.97928766010473
     },
     "gaudi3": {
-      "throughput": 98.57537548249874
+      "throughput": 99.27203250620154
     }
   },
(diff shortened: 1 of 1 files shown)


Wei-Lin-Intel:
@regisss We have updated all the Qwen2 related UT cases, please help to trigger another UT test, thanks!

astachowiczhabana:
Hi @regisss can you please merge? We will cherry-pick this change to 1.19-release and perform tests from there

PR #2032, opened 2025-06-11
Fix PT_HPU_LAZY_MODE assertion to match updated default value

jasi306:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/utils.py
+++ b/examples/text-generation/utils.py
@@ -365,8 +365,8 @@ def setup_model(args, model_dtype, model_kwargs, logger):
 
     if args.torch_compile:
         model = get_torch_compiled_model(model, logger, args)
-        assert "PT_HPU_LAZY_MODE" in os.environ and os.environ["PT_HPU_LAZY_MODE"] == "0", (
-            "Please set PT_HPU_LAZY_MODE=0 on command line when using `--torch_compile`"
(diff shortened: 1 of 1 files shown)


examples/text-generation/experimental/utils.py
```diff
@@ -459,8 +459,8 @@ def setup_model(args, model_dtype, model_kwargs, logger):
 
     if args.torch_compile:
         model = get_torch_compiled_model(model, logger, args)
-        assert "PT_HPU_LAZY_MODE" in os.environ and os.environ["PT_HPU_LAZY_MODE"] == "0", (
-            "Please set PT_HPU_LAZY_MODE=0 on command line when using `--torch_compile`"
+        assert "PT_HPU_LAZY_MODE" in os.environ and os.environ["PT_HPU_LAZY_MODE"] != "1", (
```
yafshar: @jasi306 Is this assertion valid for synapse > 1.21? Does the user need to explicitly set `PT_HPU_LAZY_MODE` ?

yafshar:
```suggestion
        if "PT_HPU_LAZY_MODE" in os.environ:
            assert os.environ["PT_HPU_LAZY_MODE"] != "1", (
```

Should we fix this to st like above? for > 1.21

AKloniecki:
I've created new PR with cherry-picked commit from this PR: #2189 .
This PR should be closed and discussion moved to new PR.

PR #2037, opened 2025-06-12
Add Eval Script for LLaMA 3.2 and other Multimodal Models

jaygala223:
# What does this PR do?
Adds a new script for evaluating multimodal models like Llama 3.2. It also includes measurements creation for running the model in FP8 later.

### DIFF
--- a/examples/image-to-text/requirements_vlm_eval.txt
+++ b/examples/image-to-text/requirements_vlm_eval.txt
@@ -1,0 +1,6 @@
+lm-eval==0.4.7
+datasets==2.21.0
+evaluate == 0.4.3
+rouge_score == 0.1.2
+accelerate < 0.34.0
+pandas <= 2.2.2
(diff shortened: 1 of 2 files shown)


examples/image-to-text/requirements_vlm_eval.txt
```diff
@@ -0,0 +1,6 @@
+lm-eval==0.4.7
+datasets==2.21.0
+evaluate == 0.4.3
+rouge_score == 0.1.2
+accelerate < 0.34.0
```
regisss, line 5: Is this a hard constraint? This won't work as we now need https://github.com/huggingface/optimum-habana/blob/5448802fb0590f471db3a79356120633447ed6fa/setup.py#L35

examples/image-to-text/run_vlm_eval.py
regisss, line 1: Can you add a section to https://github.com/huggingface/optimum-habana/blob/main/examples/image-to-text/README.md to explain a bit how to use this script please?

PR #2052, opened 2025-06-12
Extra Content: support FP8 model with bf16 KVCache do generation

astachowiczhabana:
This pull request introduces a conditional check in the `get_k_proj_weight_dtype` method to handle cases where `k_proj.weight.dtype` contains "float8". If this condition is met and `k_cache` is an instance of `KVCache`, the method now returns `k_proj.hp_dtype` instead of the default `k_proj.weight.dtype`.

Key change:

* [`optimum/habana/transformers/models/llama/modeling_llama.py`](diffhunk://#diff-30aeee6868dd1de34878aca0583f57bb5b0dd9a2a8511a80e9a6b2645f39ce6bR525-R526): Updated the `get_k_proj_weight_dtype` method to include a conditional branch for handling "float8" data types in `k_proj.weight.dtype` when `k_cache` is an instance of `KVCache`.


regisss:
@astachowiczhabana There are merge conflicts to solve

PR #2053, opened 2025-06-12
Extra Content: remove synDeviceGaudi call

astachowiczhabana:
This pull request includes a small change to the `optimum/habana/utils.py` file. The change updates the logic in the `get_device_name` function to remove the handling of the `synDeviceGaudi` device type.

### DIFF
--- a/optimum/habana/utils.py
+++ b/optimum/habana/utils.py
@@ -412,9 +412,7 @@ def get_device_name():
 
     device_type = htexp._get_device_type()
 
-    if device_type == htexp.synDeviceType.synDeviceGaudi:
-        return "gaudi"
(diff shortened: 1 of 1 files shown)


optimum/habana/utils.py
```diff
@@ -412,9 +412,7 @@ def get_device_name():
 
     device_type = htexp._get_device_type()
 
-    if device_type == htexp.synDeviceType.synDeviceGaudi:
```
regisss, line 415: `synDeviceGaudi` is not defined anymore? Does that mean Gaudi1 is not maintained anymore and we should drop any CI workflow running on it?

astachowiczhabana: this pr was created by mistake, we are not dropping Gaudi1, sorry for confusion

PR #2059, opened 2025-06-13
Fix device OOM caused by buffers not being deallocated fast enough

bkowalskiINTEL:
This PR is a fix for a reported Device OOM in Granite 8B model. The issue was caused by a big amount of ~1GB unnecessary buffers not being deallocated at the end of GaudiGenerationMixin._sample() calls, which made these buffers stay alive between GaudiGenerationMixin.generate() calls, further causing the OOM.

The solution turned out to be explicitly calling gc.collect() just before the return statements in the _sample() method, so that python's garbage collector frees the objects holding the buffers.

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -2898,7 +2898,7 @@ class GaudiGenerationMixin(GenerationMixin):
 
         # Trigger the garbage collector to make sure that unnecessary buffers are deallocated
         gc.collect()
-        
+
         if return_dict_in_generate:
             if self.config.is_encoder_decoder:
                 return GenerateEncoderDecoderOutput(
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import copy
+import gc
 import inspect
 import math
 import time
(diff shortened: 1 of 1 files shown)


astachowiczhabana (changes requested):
@bkowalskiINTEL why this has to be fixed in user code? This looks like something that should be handled in bridge

bkowalskiINTEL:
@astachowiczhabana It is a workaround which is supposed to fix a blocker issue so that it doesn't block the upcoming release. As far as I know, the root cause and more generic fix in the bridge/pytorch-fork is going to be investigated as another lower-priority task, which might be extended till the next release.

astachowiczhabana:
thx @bkowalskiINTEL. This looks bad for overall gaudi user-experience IMHO, but as a quick solution to not block the release I think we can accept it. @regisss are you ok with this change?

astachowiczhabana:
@bkowalskiINTEL there's a problem with style of your code, you can run 'make style' and it should fix everything.

bkowalskiINTEL:
@astachowiczhabana Done

bkowalskiINTEL:
In the meantime a root cause and fix for it have been found. An upstream PR for this can be found here: [https://github.com/pytorch/pytorch/pull/155791](url). Also, the fix has been merged into our pytorch-fork, so this workaround is not necessary anymore, so I'll close this PR.

PR #2062, opened 2025-06-17
Add Qwen classification

tianyuan211:
# What does this PR do?

Fixes # (issue)

## Before submitting


examples/text-classification/run_glue.py
```diff
@@ -442,9 +442,11 @@ def main():
...
 
     if model_args.add_pad_token:
-        if not model.config.pad_token_id and not tokenizer.pad_token:
-            tokenizer.pad_token = tokenizer.eos_token
-            model.config.pad_token_id = tokenizer.eos_token_id
+        pad_id = getattr(model.config, "pad_token_id", None)
+        if pad_id is None or pad_id < 0:
+            tokenizer.add_special_tokens({"pad_token": tokenizer.eos_token})
+            model.config.pad_token_id = tokenizer.pad_token_id
+            model.resize_token_embeddings(len(tokenizer))
```
regisss, line 449: Can you check if all the text-classification tests in `tests/test_examples.py` still pass with this change?
Not setting `tokenizer.pad_token` anymore may not work as intended :thinking:

regisss (commented):
I left one comment, please also check why `make style` fails.

examples/text-classification/run_glue.py
tianyuan211: Yes, there are 3 text-classification tests in that file. `TextClassificationExampleTester` passes; `MultiCardTextClassificationExampleTester` passes; `DeepSpeedTextClassificationExampleTester` llama guard model rejected my access request, I will try it later with another account. 
For `tokenizer.pad_token`, models like qwen/qwen2-7b return `<|endoftext|>`, which is not `none`. So the original condition may not work, and the tag `--add_pad_token` doesn't work in those conditions.

tianyuan211: Now I have access to the llama guard model, the eval_f1 is slightly lower than baseline metric, 0.87 <-> 0.88, no function error though.

regisss: Sounds good :+1:

regisss (approved):

PR #2064, opened 2025-06-17
Fix HabanaGenerationTime usage in tests

pbielak:
# What does this PR do?
For two test cases: `test_feature_extraction` and `test_video_mae` the line that accessed the measured time (`HabanaGenerationTime`) was still within the scope of the context manager, so that `last_duration` was still set to `None`. This commit fixes the indentation of those lines.


pbielak:
Needs to be also cherrypicked to `v1.18-release` branch

regisss:
> Needs to be also cherrypicked to `v1.18-release` branch

Done

PR #2075, opened 2025-06-23
text-generation requirements fix for transformers upgrade 4.51

gplutop7:
This PR updates the transformers version constraint in text-generation/requirements.txt:

This change is required to align with updates introduced in [#2063](https://github.com/huggingface/optimum-habana/pull/2063), which upgraded the minimum supported transformers version to 4.51. Without this fix, certain classes such as AssistantVocabTranslatorCache used in text generation are not available, leading to import errors.

### DIFF
--- a/examples/text-generation/requirements_awq.txt
+++ b/examples/text-generation/requirements_awq.txt
@@ -1,3 +1,3 @@
 triton==3.1.0
 autoawq
-transformers>=4.48.2,<=4.49.0
+transformers>=4.51.0,<4.52.0
--- a/examples/text-generation/requirements_awq.txt
+++ b/examples/text-generation/requirements_awq.txt
@@ -1,3 +1,2 @@
 triton==3.1.0
 autoawq
-transformers>=4.51.0,<4.52.0


pbielak (approved):

examples/text-generation/requirements_awq.txt
```diff
@@ -1,3 +1,3 @@
 triton==3.1.0
 autoawq
-transformers>=4.48.2,<=4.49.0
+transformers>=4.51.0,<4.52.0
```
imangohari1:
```suggestion
intel_extension_for_pytorch==2.6.0
```

examples/text-generation/requirements_awq.txt
```diff
@@ -1,3 +1,3 @@
 triton==3.1.0
 autoawq
```
imangohari1, line 2:
```suggestion
autoawq==0.2.7
```

gplutop7: We generally avoid hardcoding exact versions in requirements.txt unless absolutely necessary (e.g., due to breaking changes or compatibility issues). Pinning autoawq==0.2.7 would unnecessarily restrict flexibility and may cause issues when updating other dependencies in the future. Unless there's a clear justification (e.g., known bugs in newer versions), we prefer to keep the unpinned version.

regisss (approved):

PR #2083, opened 2025-06-24
Pin numpy version to 1.26.4 for fast diffuser tests

tszulist-hbn:
Temporary workaround to mitigate NumPy upgrade to 2.2 in PyTorch Bridge for Intel Gaudi (HPU) until support for the new NumPy is provided.

### DIFF
--- a/examples/stable-diffusion/requirements.txt
+++ b/examples/stable-diffusion/requirements.txt
@@ -1,3 +1,4 @@
 opencv-python == 4.10.0.82
 compel
 sentencepiece
+numpy==1.26.4


astachowiczhabana:
Hi @imangohari1 we don't pin version only on main, on release branches we can pin specific versions. In fact that's the root cause of problem: on release branch we've got opencv-python == 4.10.0.82 on main it's just opencv-python. @tszulist-hbn is proposing quick fix it will only allow to build next synapse dockers. Meanwhile we are investigating if opencv can be safely upgraded without causing any other disruptions

PR #2088, opened 2025-06-26
Enable trl GRPO trainer

schoi-habana:
This enables GRPO trainer
This reproduce the example from HF https://huggingface.co/learn/cookbook/en/fine_tuning_llm_grpo_trl on gaudi

image

TODO: resolve the trl version mismatch. Other trl trainers in optimum/habana/trl/trainer/  are enabled with trl==0.9.6, but grpo needs trl==0.17.0

## Before submitting


examples/trl/requirements_grpo.txt
```diff
@@ -0,0 +1,8 @@
+trl == 0.17.0
+peft == 0.12.0
+datasets == 3.0.0
```
regisss: IT doesn't work with newer versions of `datasets`? That would be surprising

examples/trl/README.md
```diff
@@ -4,10 +4,74 @@
...
 First, you should install the requirements:
+
+- For **GRPO example**:
+```bash
+$ pip install -U -r requirements_grpo.txt
+```
+
+- For **all other examples**:
 ```bash
 $ pip install -U -r requirements.txt
```
regisss, line 15: The difference between both requirements is only the TRL version right?

optimum/habana/trl/trainer/grpo_config.py
```diff
@@ -0,0 +1,316 @@
...
+
+from ... import GaudiTrainingArguments
+
+
+@dataclass
+class GaudiGRPOConfig(GaudiTrainingArguments):
+    r"""
+    Initialize GaudiGRPOConfig.
+        Adapted from https://github.com/huggingface/trl/blob/v0.17.0/trl/trainer/grpo_config.py
+        - inherit from GaudiTrainingArguments
```
regisss, line 27: This is the only change? Since we monkey patch Transformers' `TrainingArguments` with `GaudiTrainingArguments`, it doesn't work if you don't define `GaudiGRPOConfig` here?

optimum/habana/trl/trainer/grpo_trainer.py
```diff
@@ -0,0 +1,835 @@
...
+from trl.models import create_reference_model, prepare_deepspeed, unwrap_model_for_generation
+from trl.trainer.callbacks import SyncRefModelCallback
+from trl.trainer.utils import (
+    pad,
+    print_prompt_completions_sample,
+    selective_log_softmax,
+)
+
+from optimum.habana.transformers import trainer as habana_trainer
+from optimum.habana.transformers.trainer import _get_input_update_settings
```
regisss:
```suggestion
from ...transformers import trainer as habana_trainer
from ...transformers.trainer import _get_input_update_settings
```
Let's rather use relative imports

examples/trl/README.md
schoi-habana: right

examples/trl/requirements_grpo.txt
schoi-habana: just tested with the latest datasets version, it worked

optimum/habana/trl/trainer/grpo_config.py
schoi-habana: GaudiGRPOConfig needs to be defined because it includes additional parameters that are not present in `GaudiTrainingArguments`. So without this class, things will break

PR #2089, opened 2025-06-27
Use `PT_ENABLE_INT64_SUPPORT=1` for trl examples

pbielak:
# What does this PR do?

Currently, the trl example commands fail with the following error: `RuntimeError: Error when trying to cast Long to Int, Input values range [9223372036854775807, 9223372036854775807] exceeds Int range [-2147483648, 2147483647]` which is caused by lack of default support for Long in lazy mode. The `PT_ENABLE_INT64_SUPPORT` flag enables the Long support.


pbielak:
Needs to be cherry-picked to v1.18-release

regisss:
> Needs to be cherry-picked to v1.18-release

Done

PR #2121, opened 2025-07-08
Remove `requirements.txt` file

pbielak:
# What does this PR do?
This file is not used anywhere, so it can be safely removed. It must have been wrongly transferred from the optimum-habana-fork repository during a rebase.


regisss:
It was there because at some point @astachowiczhabana wanted to have all the requirements needed for testing the release branch. So I don't know if we should remove it or not?

ugolowic (approved):

PR #2124, opened 2025-07-09
Add flex attention flags and args to Llama on Habana

kareemshaik80:
# What does this PR do?

Fixes # (issue)

## Before submitting


regisss:
@kareemshaik80 Can you provide a command to run so that we can test it?
Also, I suggest adding a test for it in https://github.com/huggingface/optimum-habana/blob/main/tests/test_text_generation_example.py

PR #2138, opened 2025-07-11
Factor out model_initialization from pipeline constructor, and pass initialized model.

AKloniecki:
# What does this PR do?

Model initialization procedure sets environmental variables, which are needed prior to habanalabs .so loading. At the same time importing pipeline, in effect, loads the library.

To resolve the egg-chicken-problem, initialization of the model needed to be extracted from the constructor and performed in outside context.

### DIFF
--- a/examples/text-generation/text-generation-pipeline/run_pipeline.py
+++ b/examples/text-generation/text-generation-pipeline/run_pipeline.py
@@ -3,6 +3,8 @@ import logging
 import math
 import sys
 from pathlib import Path
+
+
 sys.path.append(str(Path(__file__).parent.parent))
 
 from run_generation import setup_parser
@@ -43,9 +45,11 @@ def main():
 
     logger.info("Initializing text-generation pipeline...")
     from utils import initialize_model
+
     model, _, tokenizer, generation_config = initialize_model(args, logger)
 
     from pipeline import GaudiTextGenerationPipeline
+
     pipe = GaudiTextGenerationPipeline(args, logger, model, tokenizer, generation_config)
 
     from optimum.habana.utils import HabanaGenerationTime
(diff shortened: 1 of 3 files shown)


karol-brejna-i:
This PR resolves a blocker for 1.19 release as well as a few examples.

examples/text-generation/text-generation-pipeline/run_pipeline.py
```diff
@@ -1,12 +1,12 @@
 import argparse
 import logging
 import math
+import sys
+from pathlib import Path
+sys.path.append(str(Path(__file__).parent.parent))
```
regisss: Why do we need to do that?

examples/text-generation/text-generation-pipeline/run_pipeline_langchain.py
```diff
@@ -17,10 +17,13 @@
...
 import logging
 import math
+import sys
+from pathlib import Path
 
 from langchain_core.prompts import PromptTemplate
 from langchain_huggingface.llms import HuggingFacePipeline
-from pipeline import GaudiTextGenerationPipeline
+
+sys.path.append(str(Path(__file__).parent.parent))
```
regisss, line 27: Same

examples/text-generation/text-generation-pipeline/run_pipeline.py
AKloniecki: When we're running the examples from designated directory - text-generation-pipeline in this case, PYTHONPATH doesn't include text-generation directory, which has the run_pipeline script, which we're importing just underneath this path adjustment.
Previously, importing pipeline has performed this:
`SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
sys.path.append(os.path.dirname(SCRIPT_DIR))`
but now that this import is moved, paths need to be adjusted accordingly.

examples/text-generation/text-generation-pipeline/run_pipeline_langchain.py
AKloniecki: Same reason as above.

examples/text-generation/text-generation-pipeline/run_pipeline.py
regisss: Ah okay I see, thanks for the explanation !

PR #2140, opened 2025-07-11
Factor out model_initialization from pipeline constructor, and pass initialized model.

AKloniecki:
# What does this PR do?

Model initialization procedure sets environmental variables, which are needed prior to habanalabs .so loading. At the same time importing pipeline, in effect, loads the library.

To resolve the egg-chicken-problem, initialization of the model needed to be extracted from the constructor and performed in outside context.


AKloniecki:
Yes, it is exactly the same.

Please cherry-pick the change.

PR #2147, opened 2025-07-15
Remove unused protobuf dependecy from examples

karol-brejna-i:
Protobuf (3.20.3) appeared in our security scans (for known vulnerabilities).

Protobuf doesn't seem to be used in the examples, where is listed as a dependency:
- examples/language-modeling/requirements.txt
- examples/summarization/requirements.txt
- examples/text-classification/requirements.txt
- examples/translation/requirements.txt

This PR removes the unnecessary dependency.

### DIFF
--- a/examples/language-modeling/requirements.txt
+++ b/examples/language-modeling/requirements.txt
@@ -1,6 +1,5 @@
 datasets >= 2.14.0
 sentencepiece != 0.1.92
-protobuf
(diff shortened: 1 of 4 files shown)


regisss:
> @regiss can you please take a look at this?

We need to make sure these examples still work as intended in CI. If yes, then LGTM.

PR #2154, opened 2025-07-17
Use profiler in text-generation-pipeline

pbielak:
# What does this PR do?

The examples in `text-generation-pipeline` fail due to profiler args still being passed separately instead of a single `profiler` argument. This commit updates the creation of the profiler and fixes the arguments passed to `self.model.generate(...)`.


regisss (approved):
LGTM

regisss:
> Needs to be cherry-picked to v1.19-release

Done

PR #2156, opened 2025-07-18
Pin datasets to version 3.6.0 to preserve trust_remote_code compatibility

gplutop7:
This PR explicitly pins the datasets library to version ==3.6.0 in order to ensure compatibility with lm-eval, which still relies on the trust_remote_code parameter.

Details:
- In recent versions of datasets (starting from 4.0.0), the trust_remote_code argument has been removed.
- Since lm-eval (including some load_dataset() calls) still depends on this parameter, version 3.6.0 is the latest compatible release.
- If an upper-bound version constraint was already set (e.g., datasets >= 1.18.0, <= 2.19.2), it remains unchanged. This PR does not modify it.

Impact:
- Prevents runtime errors caused by incompatible versions of the datasets library.
- No effect on environments that already limit datasets below 4.0.0.

### DIFF
--- a/examples/audio-classification/requirements.txt
+++ b/examples/audio-classification/requirements.txt
@@ -1,4 +1,4 @@
-datasets >= 3.0.2, <= 3.6.0
+datasets == 3.6.0
 evaluate == 0.4.3
 numba==0.60.0
 librosa == 0.10.2.post1
--- a/examples/image-to-text/requirements.txt
+++ b/examples/image-to-text/requirements.txt
@@ -3,4 +3,4 @@ Levenshtein
 sentencepiece != 0.1.92
 tiktoken
 blobfile
-datasets
+datasets == 3.6.0
--- a/examples/protein-folding/requirements.txt
+++ b/examples/protein-folding/requirements.txt
@@ -1,2 +1,2 @@
-datasets>=2.14.0
+datasets == 3.6.0
 scikit-learn == 1.5.2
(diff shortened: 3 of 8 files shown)


gplutop7:
> This should only be done in `requirements_lm_eval.txt`. For other examples, please let me know the datasets where remote code is needed and I'll upload new versions on the hub so that remote code is not needed anymore.

Thank you for the clarification — you're absolutely right that lm-eval is only used in one place, and the version pin for datasets==3.6.0 should be explicitly applied there (i.e., in requirements_lm_eval.txt) to avoid issues related to the removal of trust_remote_code in >=4.0.0.

The additional version constraint introduced elsewhere (only where no previous bounds were set) was mainly intended to ensure consistency across all environments, so that we always use a single, known-compatible version of datasets (3.6.0), even outside of lm-eval. This helps avoid silent inconsistencies that might arise from differing dataset-loading behavior in other components.

That said, we treat this as a temporary safeguard, and once all datasets relying on trust_remote_code are updated on the Hub and lm-eval no longer uses that parameter, we plan to migrate to datasets>=4.0.0 globally.

regisss:
After looking more into it, there is another issue with Datasets v4:  it relies on `torchcodec` for audio decoding, which is compatible Torch 2.7 only. So let's keep these changes and we'll update everything once `torchcodec` is supported on Gaudi.

Is it possible to make a list of all the datasets relying on remote code in your tests please? So that I upload all of them on the hub without remote code.

PR #2158, opened 2025-07-18
Fix requirements in text-classification

pbielak:
# What does this PR do?

The currently pinned `protobuf` version is incompatible with the tensorboard version that is shipped in the Gaudi containers. Moreover, this commit bumps the version of `datasets` due to an imcompatibility with newer Numpy versions.

### DIFF
--- a/examples/text-classification/requirements.txt
+++ b/examples/text-classification/requirements.txt
@@ -1,7 +1,8 @@
-datasets >= 2.4.0, <= 2.19.2
+datasets == 3.6.0
 sentencepiece != 0.1.92
 scipy == 1.13.1
 scikit-learn == 1.5.2
 protobuf == 5.29.4
+tensorboard == 2.19.0
 torch >= 1.3
 evaluate == 0.4.3


astachowiczhabana (approved):

examples/text-classification/requirements.txt
```diff
@@ -1,7 +1,8 @@
-datasets >= 2.4.0, <= 2.19.2
+datasets == 3.6.0
```
regisss, line 1: Same as my comment in another PR, pinning `datasets` to v3.6.0 should only be done for LM Eval

karol-brejna-i:
@pbielak  I believe you observed some problem with examples (datasets version incompatibility). Can you describe the exact problem we are solving here?

examples/text-classification/requirements.txt
pbielak: Updated the version to 4.0.0

pbielak:
@regisss @karol-brejna-i With the given requirements file and datasets==4.0.0, I initially run into an error, but I noticed that by upgrading `evaluate==0.4.5` the scripts passed again. I updated the commit

regisss:
@pbielak After looking more into it, there is another issue with Datasets v4:  it relies on `torchcodec` for audio decoding, which is compatible Torch 2.7 only. So let's keep the initial changes and we'll update everything once `torchcodec` is supported on Gaudi. Sorry for switching direction, I didn't know there was more to it than just remote code execution.

regisss (approved):

examples/text-classification/requirements.txt
pbielak: @regisss reverted to original changes

PR #2162, opened 2025-07-21
Fix passing return_dict to pre_attn() in Llama (#2102)

pbielak:
(cherry picked from commit b4fe45731a6441260f73a5889a863a205b40268e)


karol-brejna-i (approved):
LGTM. 
Tested in a container. 
The same change was applied to 1.19 and works fine.

regisss (approved):

PR #2165, opened 2025-07-22
Model phi performance fix

slokesha:
# What does this PR do?

This PR fixes the performance drop for the Phi model

The dtype for the matmul_qk args is temporarily in FP32 due to performance issues caused by kernels fuser when using BF16.
Added graph break with mark_step for lazy mode and attn_softmax_bf16 flag, which also improves performance in some cases.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/habana/transformers/models/phi/modeling_phi.py
+++ b/optimum/habana/transformers/models/phi/modeling_phi.py
@@ -37,6 +37,9 @@ from transformers.models.phi.modeling_phi import (
 from transformers.processing_utils import Unpack
 from transformers.utils import logging
 
+import habana_frameworks.torch.core as htcore
+import habana_frameworks.torch.hpu as hthpu
+
 from ...modeling_attn_mask_utils import (
     _gaudi_prepare_4d_causal_attention_mask,
 )
@@ -89,6 +92,7 @@ def gaudi_eager_attention_forward(
     attention_mask: Optional[torch.Tensor],
     scaling: float,
     dropout: float = 0.0,
+    attn_softmax_bf16: Optional[bool] = False,
     **kwargs,
 ):
     bsz, q_len = kwargs["input_shape"]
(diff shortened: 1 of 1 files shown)


karol-brejna-i:
@slokesha The PR looks good. Can you share some info on how to validate (like: which test to run).

12010486:
Chiming in. 
Tests that were run, and I believe you can use to check, are:

```
cd optimum-habana/
pip install -e .
pip install pytest
PT_HPU_LAZY_MODE=1 python -m pytest tests/test_text_generation_example.py --device gaudi3 -v -s --token=xxx --junitxml=/tmp/test_run_microsoft_phi.xml --log-cli-level 20 -k test_text_generation_bf16_1x[microsoft/phi-2-1-False-False]
```
and

```
PT_HPU_LAZY_MODE=1 python -m pytest tests/test_text_generation_example.py --device gaudi2 -v -s --token=xxx --junitxml=/tmp/test_run_microsoft_phi.xml --log-cli-level 20 -k test_text_generation_fp8[microsoft/phi-2-1-1-True-128-128] 
```

examples/text-generation/utils.py
```diff
@@ -728,6 +728,7 @@ def setup_generation_config(args, model, assistant_model, tokenizer):
     generation_config.trust_remote_code = args.trust_remote_code
     generation_config.valid_sequence_lengths = None
     generation_config.attn_batch_split = args.attn_batch_split
+    generation_config.fp8 = bool(args.quant_config)
```
regisss, line 731: Not sure we should add this to the generation config as there could be other ways of running the model in fp8. Maybe we could simply check the value of the env variable `QUANT_CONFIG` in the modeling file since this is how `args.quant_config` is set up: https://github.com/huggingface/optimum-habana/blob/2ac859ec63f0c90f52fb05445e79cc9dbc0a3eeb/examples/text-generation/run_generation.py#L457
WDYT?

karol-brejna-i:
@slokesha Are you still working on this PR?

PR #2172, opened 2025-07-24
fea(timm): Updated/Fixed the readme

imangohari1:
# What does this PR do?
This PR fixes/updates the timm readme file:
- package `timm` is installed via the clone step, so no need to duplicate it in the requirment.txt. 
- All changes are tested on both G2/G3 1x/8x. 

Fixes # (issue)

## Before submitting
- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [X] Did you make sure to update the documentation with your changes?

### DIFF
--- a/examples/pytorch-image-models/requirements.txt
+++ b/examples/pytorch-image-models/requirements.txt
@@ -1,2 +1,1 @@
-timm
 datasets
--- a/examples/pytorch-image-models/requirements.txt
+++ b/examples/pytorch-image-models/requirements.txt
@@ -1,1 +1,2 @@
+timm
 datasets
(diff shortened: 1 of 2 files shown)


examples/pytorch-image-models/requirements.txt
```diff
@@ -1,2 +1 @@
-timm
```
regisss: Wouldn't it be better to actually install it from pip and not from source to make sure we use a stable release? Or do you think we need to have the most recent changes?

imangohari1: I agree. updated: https://github.com/huggingface/optimum-habana/pull/2172/commits/60edf00f998afcbe336a703d5224cb9248310735

PR #2183, opened 2025-07-30
Move torch, transformers and optimum.habana imports to local scope.

AKloniecki:
# What does this PR do?

Move some imports from global script scope to local function scope to aviod loading habana shared object before env vars are initialized in examples/image-to-text/run_pipeline.py script.
Fixes # `RuntimeError: collective nonSFG is not supported during hpu graph capturing` error when running tests using run_pipeline.py script from examples/image-to-text.


pbielak (approved):
LGTM

regisss (approved):

PR #2190, opened 2025-08-04
Allow usage of cached books from project Guttenberg.

AKloniecki:
# What does this PR do?

This PR allows using preexisting data sources for prompt creation. It isn't caching the data source as it's downloading it, because that happens in multiple processes in parallel, with separate files for each `pid`, but it allows user to present their own cached version of the file, and skip the download altogether.

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -679,22 +679,29 @@ def main():
             def download_book(book_id):
                 import os
 
-                import requests
+                from accelerate import Accelerator
 
-                url = f"https://www.gutenberg.org/cache/epub/{book_id}/pg{book_id}.txt"
-                response = requests.get(url)
-                if response.status_code == 200:
-                    pid = os.getpid()
-                    save_path = f"/tmp/{book_id}_{pid}.txt"
-                    with open(save_path, "wb") as file:
-                        file.write(response.content)
-                    print(f"Book downloaded and saved to: {save_path}")
-                    return save_path
-                else:
-                    print("Failed to download book! Exiting...")
-                    import sys
+                acc = Accelerator()
+                with acc.main_process_first():
+                    save_path = f"/tmp/{book_id}.txt"
+                    if os.path.isfile(save_path):
+                        print(f"Using preexisting file: {save_path}")
+                        return save_path
+                    else:
+                        import requests
 
-                    sys.exit()
+                        url = f"https://www.gutenberg.org/cache/epub/{book_id}/pg{book_id}.txt"
+                        response = requests.get(url)
+                        if response.status_code == 200:
+                            with open(save_path, "wb") as file:
+                                file.write(response.content)
+                            print(f"Book downloaded and saved to: {save_path}")
+                            return save_path
+                        else:
+                            print("Failed to download book! Exiting...")
+                            import sys
+
+                            sys.exit()
 
             def assemble_prompt(prompt_size, book_path):
                 prompt = ""


examples/text-generation/run_generation.py
```diff
@@ -679,6 +679,10 @@ def rounder(x):
             def download_book(book_id):
                 import os
 
+                if os.path.isfile(f"/tmp/{book_id}.txt"):
```
yafshar, line 682:
```suggestion
                import glob
                matches = glob.glob(f"/tmp/{book_id}_*.txt")
                if matches:
                    # File(s) found
                    book_path = matches[0]  # or loop through if needed
```

Shouldn't we use pid when looking for the cached copy of the book?

yafshar: Shouldn't we use pid when looking for the cached copy of the book?

AKloniecki: As mentioned in PR description - this isn't caching, rather it allows a pathway to provide cached data.
Trying to make it cached has complications, as this script is running 8 processes in parallel (that's why pid is appended to name when writing to file). Also each run of the workflow will have different pids assigned to processes, so it wouldn't work correctly.

astachowiczhabana: I think @yafshar intention was to avoid race condition. To achieve both goals (unified path + race condition prevention) you can use main_process_first context manager. It blocks execution on all nodes except for the main process. Look at the code to see example usage. Thx!

AKloniecki: @astachowiczhabana  Currently there's no race. Either there is a file without pid in the name, then all processes may read independently without issues; or there is no such file, and all processes need to download their own and save with _pid.

yafshar: @AKloniecki thanks for the explanation, now your code makes sense to me. But, if the intention is for users to provide their own cached version of the file, it might be helpful to mention this explicitly in the README or documentation. That way, users are aware of this feature when running the code.

AKloniecki: @yafshar
Following a solution suggested by @astachowiczhabana I've updated the code, to perform the download in `main_process_first` context and perform true caching of the data source file.

examples/text-generation/run_generation.py
```diff
@@ -679,22 +679,29 @@ def rounder(x):
...
-                    save_path = f"/tmp/{book_id}_{pid}.txt"
-                    with open(save_path, "wb") as file:
-                        file.write(response.content)
-                    print(f"Book downloaded and saved to: {save_path}")
-                    return save_path
-                else:
-                    print("Failed to download book! Exiting...")
-                    import sys
-
-                    sys.exit()
```
yafshar, line 697: I think `Accelerator` is too complex for the required functionality and maybe you should replace it with `PartialState` for the book download functionality to reduce overhead, as the task only requires basic distributed process management.

below is just a mere suggestion

```py
def download_book(book_id) -> Path:
    from accelerate import PartialState
    import tempfile

    state = PartialState()
    shared_dir = Path(os.environ.get("HF_DATASETS_CACHE", tempfile.gettempdir()))
    shared_dir.mkdir(parents=True, exist_ok=True)
    save_path = shared_dir / f"{book_id}.txt"

    if state.is_main_process:
        if save_path.exists():
            print(f"Using preexisting file: {save_path}")
        else:
            url = f"https://www.gutenberg.org/cache/epub/{book_id}/pg{book_id}.txt"
            try:
                resp = requests.get(url, timeout=30)
                resp.raise_for_status()
                try:
                    save_path.write_bytes(resp.content)
                    print(f"Book {book_id} downloaded → {save_path}")
                except OSError as e:
                    print(f"Failed to save book to {save_path}: {e}")
                    raise
            except requests.RequestException as e:
                print(f"Failed to download book from {url}: {e}")
                raise
    state.wait_for_everyone()
    return save_path
```

AKloniecki: @yafshar 
Thanks for the suggestion. Unfortunately, for some reason substituting `PartialState.wait_for_everyone()` in place of `Accelerator.main_process_first()` results in some kind of deadlock, and workload never finishes. I'm not sure what the underlying cause is.
Regarding creation of `tempfile`, one of main goals was to provide a pathway to feed the data from a known location, to aid in testing on isolated networks. `HF_DATASETS_CACHE` seems to be already in use for downloading HF own datasets, wouldn't want to interfere with that, and having a random `tempfile` would defeat the purpose altogether.

astachowiczhabana (approved):

PR #2191, opened 2025-08-06
Remove unnecessary comparisons to None

ugolowic:
Fixing several instances of dead code pointed out by the Coverity tool. These are comparing a value to `None` in cases where previous paths ensure that the value cannot be `None`.

* diffusers/models/attention_processor.py
`if encoder_hidden_states is None` is dead code, because encoder_hidden_states is accessed a couple of lines above

* diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py
`kwargs` will never be `None` here, since it's a standard python `**kwargs` parameter which is guaranteed to be at least an empty dict

* transformers/models/glm4v/modeling_chatglm.py
clean up logic - GLM4VModel assumes `input_ids` cannot be `None`, so moving the check to the top of the method and removing others

* transformers/models/mpt/modeling_mpt.py
- eliminating `if past_key_value is None`, because all previous paths ensure it's not None
- delete `if token_idx is not None`, because it's insde an `else` statement which ensures it's not `None`

* trl/trainer/sft_trainer.py
checks `if args is None` or `if args is not None` in GaudiSFTTrainer.__init__ are unnecessary, because it's
ensured that args are never none at the top of the `__init__` method


regisss (approved):
LGTM

PR #2194, opened 2025-08-07
Remove instances of logically dead code

ugolowic:
Removing blocks of dead code:

* transformers/models/glm4v/modeling_chatglm.py
* transformers/models/chatglm/modeling_chatglm.py
`combined_attention_mask = (expanded_attn_mask if combined_attention_mask is None else expanded_attn_mask + combined_attention_mask)` is after `if attention_mask is not None:`

* transformers/models/falcon/modeling_falcon.py
`if token_idx is not None` is in a block executed only if `token_idx is not None`

* transformers/models/llava/modeling_llava.py
```
loss = None
[...]
return (loss,) + output if loss is not None else output
```

* transformers/models/llava_next/modeling_llava_next.py
if token_idx is None:
    [...]
else:
    legacy_processing = [...] (input_ids.shape[-1] == 1 if token_idx is None else token_idx == 1)
    [...]

* transformers/models/llava_onevision/modeling_llava_onevision.py
```
if token_idx is None:
    [...]
else:
    legacy_processing = [...] (input_ids.shape[-1] == 1 if token_idx is None else token_idx == 1)

* transformers/models/mpt/modeling_mpt.py
There's no way `past_key_value is None`, because previous paths either check if it's not `None` or assign a tensor to it.

* transformers/models/phi/modeling_phi.py
* transformers/models/starcoder2/modeling_starcoder2.py
Since `use_new_cache` is set to `False`, whole block after `if use_new_cache` is dead. This was probably left for compatibility with upstream transformers, but the transformers code changed since.

### DIFF
--- a/optimum/habana/transformers/models/chatglm/modeling_chatglm.py
+++ b/optimum/habana/transformers/models/chatglm/modeling_chatglm.py
@@ -1217,9 +1217,7 @@ class ChatGLMModel(ChatGLMPreTrainedModel):
                     device=expanded_attn_mask.device,
                 )
                 expanded_attn_mask = torch.cat([pre_seq_mask, expanded_attn_mask], dim=-1)
-            combined_attention_mask = (
-                expanded_attn_mask if combined_attention_mask is None else expanded_attn_mask + combined_attention_mask
-            )
+            combined_attention_mask = expanded_attn_mask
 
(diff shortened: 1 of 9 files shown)


regisss (approved):
LGTM

PR #2197, opened 2025-08-08
Fix for eos in eager mode

12010486:
A fast repro is as below:

```
PT_HPU_LAZY_MODE=0 python run_generation.py --model_name_or_path meta-llama/Llama-3.2-1B-Instruct --use_kv_cache --max_new_tokens 100 --do_sample --batch_size 2 --prompt "Hello world" "How are you?" --sdp_on_bf16 --no-ignore_eos
```

Which would have failed without this PR. 

We get:

```
Input/outputs:
input 1: ('Hello world',)
output 1.1: ("Hello world! This is my first time using this platform. I am excited to share my knowledge and experiences with the world. I am a student of computer science and engineering, and I am currently working on my master's degree.\n\nMy area of interest is in artificial intelligence, machine learning, and deep learning. I have been studying these topics for several years and have gained a good understanding of the concepts and techniques involved. I have also been experimenting with various deep learning frameworks such as TensorFlow and PyTorch.\n\n",)

input 2: ('How are you?',)
output 2.1: ("How are you? Today is a beautiful day!\nI'm doing great, thanks for asking! It's a lovely day to be outside, isn't it? I love days like this, when the sun is shining and the birds are singing. There's something so uplifting about being in nature, don't you think?\n\nI've been meaning to get out and enjoy the great outdoors, but I've been stuck inside a lot lately. How about you? What are you up to today? Do you have any fun plans",)


Stats:
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Input tokens
Throughput (including tokenization) = 108.17886215980566 tokens/second
Average first token latency         = 22.46260239626281 ms
Average rest token latency          = 18.413477945447234 ms
Average end to end latency          = 1848.2638252025936 ms
Memory allocated                    = 5.62 GB
Max memory allocated                = 5.76 GB
Graph compilation duration          = 5.586424403998535 seconds
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
`--no-ignore_eos` is just slightly slower than `--ignore_eos` in eager, `113.073` vs `108.179` `tokens/second`. This is expected

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -2872,19 +2872,16 @@ class GaudiGenerationMixin(GenerationMixin):
 
         if batch_size > 1 and has_eos_stopping_criteria:
             eos_token_id = generation_config.eos_token_id
+            # Init eos_positions
+            eos_positions = torch.full((batch_size,), start_token_idx, dtype=torch.long, device=input_ids.device)
             # Find the positions of the first eos_token_id in each sequence
-            _, seq_len = input_ids.shape
-            if start_token_idx >= seq_len:
-                # All slices are empty, so no EOS can be found
-                eos_positions = torch.full((batch_size,), seq_len, dtype=torch.long, device=input_ids.device)
-            else:
-                mask_ = torch.isin(input_ids[:, start_token_idx:], torch.tensor(eos_token_id, device=input_ids.device))
-                # If mask is all False for a row, argmax returns 0, which is incorrect.
-                # So, set eos_positions to seq_len if no EOS is found.
-                found = mask_.any(dim=1)
-                first_eos = mask_.int().argmax(dim=1) + start_token_idx
-                eos_positions = torch.full((batch_size,), seq_len, dtype=torch.long, device=input_ids.device)
-                eos_positions[found] = first_eos[found]
+            eos_positions_tmp = torch.isin(
+                input_ids[:, start_token_idx:], torch.tensor(eos_token_id).to(device=input_ids.device)
+            ).int()
+            if eos_positions_tmp.numel() != 0:
+                # argmax(dim=1) is throwing this error in eager mode, if the tensor is empty
+                eos_positions = eos_positions + eos_positions_tmp.argmax(dim=1)
+
             # Create a mask for positions greater than the first eos_token_id
             mask = torch.arange(generation_config.max_length, device="hpu").expand(
                 batch_size, generation_config.max_length


optimum/habana/transformers/generation/utils.py
```diff
@@ -2873,10 +2873,18 @@ def _sample(
...
+                # All slices are empty, so no EOS can be found
+                eos_positions = torch.full((batch_size,), seq_len, dtype=torch.long, device=input_ids.device)
+            else:
+                mask_ = torch.isin(input_ids[:, start_token_idx:], torch.tensor(eos_token_id, device=input_ids.device))
+                # If mask is all False for a row, argmax returns 0, which is incorrect.
+                # So, set eos_positions to seq_len if no EOS is found.
+                found = mask_.any(dim=1)
+                first_eos = mask_.int().argmax(dim=1) + start_token_idx
+                eos_positions = torch.full((batch_size,), seq_len, dtype=torch.long, device=input_ids.device)
+                eos_positions[found] = first_eos[found]
```
yafshar:
```suggestion
            eos_positions = torch.full((batch_size,), seq_len, dtype=torch.long, device=input_ids.device)
            if start_token_idx < seq_len:
                # look for the first EOS in every row (starting at start_token_idx)
                slice_ = input_ids[:, start_token_idx:]
                eos_in_slice = slice_ == eos_token_id

                # argmax returns the FIRST 1 in every row; all-0 rows → 0
                first_rel = eos_in_slice.int().argmax(dim=1)
                # rows that actually contain an EOS anywhere (including index 0)
                has_eos = eos_in_slice.any(dim=1)

                eos_positions[has_eos] = first_rel[has_eos] + start_token_idx
```


@12010486 can you test the suggestion. Using `torch.isin` here for a single `eos_token_id` is not correct as it is designed to test membership against an arbitrary set of values. Internally it builds a hash-table (or sorts+binary-searches) so that every element of input_ids can be compared against every element in the test set.

12010486: @yafshar eos can be more than a single value, at least when you use this code from run_lm_eval.py (it was a list of three for llama-3.2-1b-Instruct). Let me check with the example I shared instead

12010486: I can confirm that `eos_token_id=[128001, 128008, 128009]` regardless the running function, I guess last models have more than one. I can change the code to include the special case of single `eos_token_id`

12010486: Regarding the  _I can change the code to include the special case of single eos_token_id_ this would not be my preference, it seems this way is more generic and the logic holds because eos > single-element. What do you think?

yafshar: @12010486 thanks for the clarification, I still think `torch.isin` is too complex for this simple task

```py
eos_positions = torch.full((batch_size,), seq_len, dtype=torch.long, device=input_ids.device)

if start_token_idx < seq_len:
    slice_ = input_ids[:, start_token_idx:]   # [B, L-s]
    
    # - turn the EOS list into a 1-D tensor on the same device
    eos_tensor = torch.as_tensor(eos_token_id, dtype=input_ids.dtype, device=input_ids.device)  # shape [K]

    # - fast membership test: broadcast equality then reduce with OR
    #    (slice_ == eos_tensor.unsqueeze(1).unsqueeze(0)) -> [B, K, L-s]
    #    any(dim=1) collapses the K dimension
    eos_mask = (slice_.unsqueeze(1) == eos_tensor.view(1, -1, 1)).any(dim=1)

    # - first EOS position in every row
    first_rel = eos_mask.int().argmax(dim=1)  # 0 if none
    has_eos   = eos_mask.any(dim=1)

    eos_positions[has_eos] = first_rel[has_eos] + start_token_idx
```

this is just a suggestion, if it takes time or you do not like it please go ahead with your solution, but remove the duplicate code and use only one if condition. Thanks

regisss: I'm okay with `torch.isin` if there is no perf regression.
I agree with @yafshar's suggestion to remove the duplicate code.

12010486: Just to confirm that I'm on it at the moment. I could see that the major difference in performance is actually wrt the old/legacy code. Around 27-33x slower when running on lazy (yaser's suggestion at 27, mine 33), and 8-10x when running on eager (original PR proposal is faster this time), so I'm looking for something with better perf

12010486: Done! Whenever you have time for a final check, the proposed is the fastest solution I've found. I can share more details if needed

PR #2201, opened 2025-08-11
Remove potential weakness reported by static code analysis -- CWE-476 -- bad null-like value in chatglm

karol-brejna-i:
Protect aginst None/undefined value used in chatglm.

### DIFF
--- a/optimum/habana/transformers/models/chatglm/modeling_chatglm.py
+++ b/optimum/habana/transformers/models/chatglm/modeling_chatglm.py
@@ -206,9 +206,7 @@ def gaudi_chatglm_repeat_kv(
     new_q_shape = (batch, num_key_value_heads, n_rep, q_len, head_dim)
     query_layer = query_layer.reshape(new_q_shape)
 
-    if attention_mask is not None:
-        # Add groups dim and set to 1
-        attention_mask = attention_mask.unsqueeze(1)
+    attention_mask = attention_mask.unsqueeze(1)
 
     return query_layer, key_layer, value_layer, attention_mask
 
--- a/optimum/habana/transformers/models/glm4v/modeling_chatglm.py
+++ b/optimum/habana/transformers/models/glm4v/modeling_chatglm.py
@@ -1085,7 +1085,7 @@ class GLMTransformer(torch.nn.Module):
         if self.post_layer_norm:
             hidden_states = self.final_layernorm(hidden_states)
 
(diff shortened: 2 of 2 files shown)


optimum/habana/transformers/models/chatglm/modeling_chatglm.py
```diff
@@ -206,9 +206,7 @@ def gaudi_chatglm_repeat_kv(
     new_q_shape = (batch, num_key_value_heads, n_rep, q_len, head_dim)
     query_layer = query_layer.reshape(new_q_shape)
 
-    if attention_mask is not None:
-        # Add groups dim and set to 1
-        attention_mask = attention_mask.unsqueeze(1)
+    attention_mask = attention_mask.unsqueeze(1)
```
astachowiczhabana, line 209: This is called from CoreAttention, I see there are multiple places where attention_mask is checked against None, I don't think this is correct change

optimum/habana/transformers/models/chatglm/tokenization_chatglm.py
```diff
@@ -347,7 +347,11 @@ def _pad(
         if max_length is not None and pad_to_multiple_of is not None and (max_length % pad_to_multiple_of != 0):
             max_length = ((max_length // pad_to_multiple_of) + 1) * pad_to_multiple_of
 
-        needs_to_be_padded = padding_strategy != PaddingStrategy.DO_NOT_PAD and len(required_input) != max_length
+        needs_to_be_padded = (
```
astachowiczhabana, line 350: Now in this check I see 2 logic mixed-up:
1. Should we pad
2. Can we pad

If we are changing the file, I'd change to something more readable:

should_pad = padding_strategy != PaddingStrategy.DO_NOT_PAD
can_pad = max_length is not None and len(required_input) != max_length
needs_to_be_padded = should_pad and can_pad

optimum/habana/transformers/models/glm4v/modeling_chatglm.py
```diff
@@ -1534,6 +1534,9 @@ def prepare_inputs_for_generation(
 
         if past_key_values:
```
astachowiczhabana, line 1535: I don't understand this whole condition:  How checking past_key_values can assure if token_idx is not None?

astachowiczhabana (changes requested):

karol-brejna-i:
Closed as it is not needed at this moment.

PR #2203, opened 2025-08-11
Protecting `mask` undefined value.

karol-brejna-i:


optimum/habana/transformers/models/mixtral/modeling_mixtral.py
```diff
@@ -329,7 +329,10 @@ def forward(fsdpa, q, k, v, mask, causal, q_block_size):
         for i in range(q_tiles):
             s, e = i * q_block_size, (i + 1) * q_block_size
             row_q = q[:, :, s:e, :]
-            row_mask = mask[:, :, s:e, :]
+            if mask is not None:
+                row_mask = mask[:, :, s:e, :]
+            else:
+                row_mask = None
             attn_output[:, :, s:e, :] = fsdpa(row_q, k, v, row_mask, 0.0, causal, None)
```
astachowiczhabana, line 336: I've double checked what happend if row_mask is None and confirming that FusedSDPA is handling this case

PR #2209, opened 2025-08-18
Upgrade to Transformers v4.55

regisss:
# What does this PR do?

Aligning the codebase with Transformers v4.55.2.

Known issues:
- text-generation tests are failing for a few models

## Before submitting


optimum/habana/transformers/models/paligemma/modeling_paligemma.py
```diff
@@ -49,17 +48,13 @@ def forward(
...
-        **lm_kwargs,
+        **kwargs,
     ) -> Union[tuple, PaliGemmaCausalLMOutputWithPast]:
         """
         Inherits from PaliGemmaForConditionalGeneration::forward https://github.com/huggingface/transformers/blob/v4.45.1/src/transformers/models/paligemma/modeling_paligemma.py#L402
         The only differences are:
         - add new args token_idx
         """
-
-        if (input_ids is None) ^ (inputs_embeds is not None):
```
pbielak, line 60: Should this be removed?

regisss: It's done in PaliGemmaModel: https://github.com/huggingface/transformers/blob/v4.55.4/src/transformers/models/paligemma/modeling_paligemma.py#L324
But I expect this model to fail and to require a fix, I need to run the tests.

pbielak: So if it's in the upstream, why is it removed here?

regisss: What I mean is that the upstream now does it in the class `PaliGemmaModel` and not in `PaliGemmaForConditionalGeneration` anymore. That's why I'm removing it from `GaudiPaliGemmaForConditionalGeneration`.

regisss:
Merging as discussed with @astachowiczhabana. Will solve known ongoing issues in new PRs.

PR #2222, opened 2025-08-22
Fix Habana profiler crash by making start() idempotent

gplutop7:
This PR updates the HabanaProfile class in misc.py to make start() idempotent and prevent runtime crashes when multiple profiler.start() calls are issued within the same test run (e.g. during warm-up and active steps in text generation).

Key changes:

- Replaced strict error raising with safe handling when start() is called on an already running profiler instance.
- Added a guard that skips redundant start() calls, logging a warning instead of crashing.
- Ensures that only one profiler instance runs at a time; if another is active, the previous one is stopped safely.
- Maintains backward compatibility while improving test stability for tests/test_habana_profiler_integration.py::test_integration_text_generation.

Impact:

- Eliminates RuntimeError: Cannot start profiler, another profiler instance is already running.
- Tests involving profiling (e.g., text generation integration) now pass reliably without manual intervention.
- Improves robustness of profiling logic for Habana devices across Gaudi2/Gaudi3.

### DIFF
--- a/optimum/habana/utils/misc.py
+++ b/optimum/habana/utils/misc.py
@@ -17,6 +17,7 @@ import os
 import random
 import subprocess
 import time
+import weakref
 from typing import Any, Dict, List, Optional
 
 import numpy as np
@@ -292,7 +293,7 @@ class HabanaGenerationTime(object):
 
 
 class HabanaProfile:
-    _profilers = []
+    _profilers = weakref.WeakSet()
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


gplutop7:
> Can you provide an example of command where such a crash happens?

An example command that reproduces the crash is: pytest tests/test_habana_profiler_integration.py::test_integration_text_generation --token <HF_TOKEN>

This test triggers the Habana profiler inside run_generation.py, and without the fix it fails with:
`RuntimeError: Cannot start profiler, another profiler instance is already running`

After the changes in this PR, the test completes successfully.

optimum/habana/utils/misc.py
```diff
@@ -322,11 +323,22 @@ def __init__(
...
+            return
+
+        running_others = [p for p in list(self._profilers) if (p is not self and getattr(p, "_running", False))]
+        if running_others:
+            logger.warning("Another Habana profiler instance is running; stopping it before starting a new one")
+            for p in running_others:
+                try:
+                    p.stop()
+                except Exception:
+                    pass
```
regisss: It would be good to catch the exception message and log it into a warning

optimum/habana/utils/misc.py
```diff
@@ -339,6 +351,16 @@ def step(self):
...
 
+    def __enter__(self):
+        self.start()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        try:
+            self.stop()
+        except Exception:
+            pass
```
regisss: Same

optimum/habana/utils/misc.py
gplutop7: Fixed.

optimum/habana/utils/misc.py
gplutop7: Fixed.

PR #2223, opened 2025-08-22
Respect `--dataset_max_samples` when using `--mlcommons_dataset`

pbielak:
# What does this PR do?

The current implementation of `run_generation.py` does not take into account the value of `--dataset_max_samples` when using an MLCommons dataset (via the `--mlcommons_dataset` argument). Limiting the number of dataset samples allows for faster debugging.


regisss (approved):
LGTM!

PR #2224, opened 2025-08-22
Add system_instruction support for lm_eval in OH

srajabos:
# What does this PR do?
Add system_instruction for lm_eval in OH that helps to boost accuracy for mmlu tasks

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/text-generation/run_lm_eval.py
+++ b/examples/text-generation/run_lm_eval.py
@@ -87,6 +87,12 @@ def setup_lm_eval_parser():
         action="store_true",
         help="If True, prints extra-logs for all tasks",
     )
+    parser.add_argument(
+        "--system_instruction",
+        type=str,
+        default=None,
+        help="System instruction to be used in the prompt",
+    )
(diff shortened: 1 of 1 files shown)


srajabos:
@mandy-li @skavulya - kindly review

examples/text-generation/run_lm_eval.py
```diff
@@ -110,7 +116,8 @@ def main() -> None:
     with HabanaGenerationTime() as timer:
         with torch.no_grad():
             log_samples = args.log_samples
-            results = evaluator.simple_evaluate(lm, tasks=args.tasks, limit=args.limit_iters, log_samples=log_samples)
+            system_instruction = args.system_instruction
```
mandy-li: i don't see the need for line 119. For other parameters, we directly pass from args, why not system_instruction parameter?

mandy-li (commented):
comment for minor change

examples/text-generation/run_lm_eval.py
srajabos: Just followed log_samples (line 118). I'll directly pass the args.

PR #2226, opened 2025-08-25
Ensure decoder_only tests are not performed on encoder-only models.

AKloniecki:
# What does this PR do?

Add check to ensure that encoder-only models aren't tested with decoder-only tests.


regisss (approved):
LGTM

PR #2227, opened 2025-08-25
Add slow test for installs and update example requirements (peft)

gplutop7:
This PR introduces the following updates for optimum-habana:

Slow tests

- Added test_installs under slow_tests to validate installation requirements.

Requirements updates across examples

- Stable Diffusion (examples/stable-diffusion/requirements.txt): added peft==0.15.0.
- TRL / GRPO (examples/trl/requirements_grpo.txt): upgraded peft from 0.12.0 → 0.15.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,7 +162,6 @@ slow_tests_video_llava_example: test_installs
 	python -m pytest tests/test_video_llava.py
 
 slow_tests_fsdp: test_installs
-	python -m pip install tensorboard==2.19.0
 	python -m pytest tests/test_fsdp_examples.py -v -s --token $(TOKEN)
 
 slow_tests_trl_ddpo: test_installs
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,7 @@ slow_tests_video_llava_example: test_installs
 	python -m pytest tests/test_video_llava.py
 
 slow_tests_fsdp: test_installs
+	python -m pip install tensorboard==2.19.0
 	python -m pytest tests/test_fsdp_examples.py -v -s --token $(TOKEN)
 
 slow_tests_trl_ddpo: test_installs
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -162,6 +162,7 @@ slow_tests_video_llava_example: test_installs
 	python -m pytest tests/test_video_llava.py
 
 slow_tests_fsdp: test_installs
+	python -m pip install tensorboard==2.19.0
```
regisss: Why do we need tensorboard here?

examples/language-modeling/requirements.txt
```diff
@@ -4,3 +4,4 @@ protobuf
 evaluate
 scikit-learn == 1.5.2
 peft == 0.12.0
+tensorboard == 2.19.0
```
regisss: Same question

Makefile
gplutop7: We previously observed failures in the FSDP language-modeling tests due to Trainer requiring tensorboard when report_to=["tensorboard"] was set. At that time, we added tensorboard to the installation step in order to unblock the tests.

Currently, the described issue no longer occurs (the tests run successfully without tensorboard), so this extra installation is not needed anymore. The change has been removed to align with the current state of the examples/tests.

examples/language-modeling/requirements.txt
gplutop7: We previously observed failures in the FSDP language-modeling tests due to Trainer requiring tensorboard when report_to=["tensorboard"] was set. At that time, we added tensorboard to the installation step in order to unblock the tests.

Currently, the described issue no longer occurs (the tests run successfully without tensorboard), so this extra installation is not needed anymore. The change has been removed to align with the current state of the examples/tests.

regisss (approved):
LGTM

PR #2229, opened 2025-08-27
Add configurable DeepSpeed installation via DEEPSPEED_SPEC env var in Makefile

gplutop7:
This PR updates the Makefile so that DeepSpeed installation is no longer hardcoded.
Instead, the version (or fork) to install can now be customized through the environment variable DEEPSPEED_SPEC.

- The default value of DEEPSPEED_SPEC is defined directly in the Makefile (currently pointing to the HabanaAI repository and version).
- Users can override it by exporting their own value before running the tests.

Examples:

Default behavior (uses the value defined in the Makefile):
`make slow_tests_text_generation_example_8x`

Custom behavior (installs from internal fork with a GitHub token):
```
export DEEPSPEED_SPEC="git+https://${GH_TOKEN}@github.com/habana-internal/deepspeed-fork.git@v1.22.0"
make slow_tests_text_generation_example_8x
```

This makes it easier to switch between official and internal DeepSpeed forks or test new versions without modifying the Makefile itself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,8 +101,7 @@ slow_tests_8x: test_installs
 	exit $$((status1 + status2))
 
 # Run DeepSpeed non-regression tests
-slow_tests_deepspeed: test_installs
-	python -m pip install git+https://github.com/HabanaAI/DeepSpeed.git@1.21.0
+slow_tests_deepspeed: test_installs install_deepspeed
 	python -m pytest tests/test_examples.py -v -s -k "deepspeed"
 
 slow_tests_diffusers: test_installs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -225,3 +220,11 @@ clean:
 
 test_installs:
 	python -m pip install .[tests]
+
+DEEPSPEED_SPEC ?= git+https://github.com/HabanaAI/DeepSpeed.git@1.21.0
+
+install_deepspeed:
+	@set -eu
+	@echo "Installing DeepSpeed (customizable via DEEPSPEED_SPEC env var)"
+	# Uwaga: nie echo-ujemy wartości z tokenem
```
regisss: To translate in English maybe?

gplutop7: Fixed

PR #2234, opened 2025-08-31
transformer qwen3 Porting to optimum-habana

rkumar2patel:
This PR ports qwen3 model testing to the optimum-habana library for G2 hardware acceleration. The testing suite provides comprehensive coverage for the qwen3 model implementation on Habana Processing Units (HPUs).

Key changes:

Added complete test suite for qwen3 model with HPU-specific adaptations
Configured HPU device targeting and Habana-specific transformers integration

### DIFF
--- a/tests/transformers/tests/causal_lm_tester.py
+++ b/tests/transformers/tests/causal_lm_tester.py
@@ -1,0 +1,454 @@
+
+# Copyright 2025 HuggingFace Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tempfile
+from inspect import signature
+
+import pytest
+from parameterized import parameterized
+
+from transformers import set_seed
+from transformers.testing_utils import (
+    is_flaky,
+    require_flash_attn,
+    require_torch_gpu,
+    slow,
+)
+
+from .test_configuration_common import ConfigTester
+from .test_modeling_common import (
+    GenerationTesterMixin,
+    ModelTesterMixin,
+    ids_tensor,
+    is_torch_available,
+    require_torch,
+    torch_device,
+)
+from .test_pipeline_mixin import PipelineTesterMixin
+
+
+if is_torch_available():
+    import torch
+
+
+class CausalLMModelTester:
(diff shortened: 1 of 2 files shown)


tests/transformers/tests/models/qwen3/test_modeling_qwen3.py
```diff
@@ -0,0 +1,731 @@
...
+from ...causal_lm_tester import CausalLMModelTest, CausalLMModelTester
+
+
+class Qwen3ModelTester(CausalLMModelTester):
+    config_class = Qwen3Config
+    if is_torch_available():
+        base_model_class = Qwen3Model
+        causal_lm_class = Qwen3ForCausalLM
+        sequence_class = Qwen3ForSequenceClassification
+        token_class = Qwen3ForTokenClassification
```
gplutop7: sequence_classification_class = Qwen3ForSequenceClassification
token_classification_class    = Qwen3ForTokenClassification

tests/transformers/tests/causal_lm_tester.py
```diff
@@ -0,0 +1,454 @@
...
+        return self.config_class(**kwargs)
+
+    def create_and_check_model(
+        self, config, input_ids, token_type_ids, input_mask, sequence_labels, token_labels, choice_labels
+    ):
+        model = self.base_model_class(config=config)
+        model.to(torch_device)
+        model.eval()
+        result = model(input_ids, attention_mask=input_mask)
+        result = model(input_ids)
```
gplutop7: The second call overwrites result, so you never verify the masked path; it also doubles compute.

tests/transformers/tests/causal_lm_tester.py
```diff
@@ -0,0 +1,454 @@
...
+        self.mamba_d_conv = mamba_d_conv
+        self.mamba_expand = mamba_expand
+        self.mamba_chunk_size = mamba_chunk_size
+
+    def prepare_config_and_inputs(self):
+        input_ids = ids_tensor([self.batch_size, self.seq_length], self.vocab_size)
+
+        input_mask = None
+        if self.use_input_mask:
+            input_mask = torch.tril(torch.ones_like(input_ids).to(torch_device))
```
gplutop7: Did you intentionally make attention_mask lower-triangular? In HF, attention_mask is usually shape [batch, seq] with 1 for real tokens and 0 for padding.

tests/transformers/tests/models/qwen3/__init__.py
gplutop7, line 1: Suggestion - Add the guards: 

```
if not is_habana_available():
    pytest.skip("HPU not available", allow_module_level=True)

if os.environ.get("HF_HUB_OFFLINE") == "1":
    pytest.skip("Requires HF Hub access (HF_HUB_OFFLINE=1).", allow_module_level=True)
```

rkumar2patel: Since this guard is not present in other optimum-habana models, I have followed the same approach for consistency. If we determine it is required, we should incorporate it across all models rather than selectively, to avoid any inconsistency.

PR #2236, opened 2025-09-02
[image-classification] README example fix

ugolowic:
One of the examples in image-classification README failed with an exception. The fix requires that train and eval batch sizes are be equal.


astachowiczhabana:
LGTM

astachowiczhabana:
@regisss can we merge this?
Next Synapse Release Validation is failing from unrelated reasons, the change was confirmed to be working

PR #2251, opened 2025-09-10
Disable SDP on BF16 default for generic diffusers HPU support

dsocek:
# What does this PR do?

HPU OOB functionality for diffusers' models use GPU-HPU Migration Toolkit.  After 1.21 we default to eager mode where the default SDP kernel in FP32 precision. Previously, the default lazy mode in GPU-HPU Migration Toolkit used SDP in BF16 precision (SW-226104).  This causes performance degradation. https://github.com/huggingface/diffusers/pull/12310 PR is submitted to improve performance on HPU devices.

In OH we want to keep the default as is, which is what this PR is for.


regisss (approved):
LGTM

PR #2257, opened 2025-09-12
Fix HabanaProfile unit tests to align with idempotent start behavior

gplutop7:
This PR updates the HabanaProfile unit tests to match the current implementation of HabanaProfile.start(), which no longer raises a RuntimeError when another profiler instance is running. Instead, the implementation gracefully stops any running profiler and starts the new one.

Added a cleanup fixture that:

- Removes the ./hpu_profile directory before each test run.
- Resets HabanaProfile._profilers to a fresh WeakSet() to avoid state leakage between tests.

Updated the last test:

- Replaced test_cannot_start_profiler_when_another_is_running with test_starting_new_profiler_stops_previous.
- The new test asserts that starting a new profiler instance stops the previous one and marks the new profiler as running, reflecting the current semantics of HabanaProfile.start().


regisss (approved):
LGTM

PR #2264, opened 2025-09-17
Check if generation_config has attr 'static_shapes' before accessing value

AKloniecki:
# What does this PR do?

This is W/A for not having properly assigned static_shapes property of generation_config.
This should unblock CI runs of Model Llama 3-2-11B 2048/2048 Tokens bfloat16 tests, which have been failing for some time now.

### DIFF
--- a/optimum/habana/transformers/generation/utils.py
+++ b/optimum/habana/transformers/generation/utils.py
@@ -1034,7 +1034,7 @@ class GaudiGenerationMixin(GenerationMixin):
                 if generation_config.decoder_start_token_id is None:
                     generation_config.decoder_start_token_id = self.generation_config.decoder_start_token_id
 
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/generation/utils.py
```diff
@@ -1034,7 +1034,7 @@ def _prepare_generation_config(
                 if generation_config.decoder_start_token_id is None:
                     generation_config.decoder_start_token_id = self.generation_config.decoder_start_token_id
 
-        if generation_config.static_shapes is None:
+        if not hasattr(generation_config, "static_shapes") or generation_config.static_shapes is None:
```
yafshar:
```suggestion
        if getattr(generation_config, "static_shapes") is None:
```

regisss (approved):
LGTM

PR #2271, opened 2025-09-19
Backport GPTBigCodeAttention from v4.53.0 to avoid major refactoring.

AKloniecki:
# What does this PR do?

Between transformers versions v4.53 and v4.54, a major refactor of GPTBigCodeAttention was made. This caused our GaudiBPTBigCodeAttention implementation incompatible.
This PR backports `_get_mask_value()` function from version v4.53, as well as overwrites `attn_dropout` property to make it an operator once again, as it was being stored as a scalar in v4.54+, and used in new `forward()` implementation differently than before.
This reenables text generation using bigcode, i.e:
`cd examples/text-generation`
`pip install -r requirements.txt`
`export PT_HPU_ENABLE_REFINE_DYNAMIC_SHAPES=1`
`PT_HPU_LAZY_MODE=1  python3 ../gaudi_spawn.py  --world_size 1 run_generation.py --batch_size 1 --bf16 --model_name_or_path bigcode/starcoder --use_hpu_graphs --use_kv_cache --n_iterations 1 --max_new_tokens 100`

### DIFF
--- a/optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
+++ b/optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
@@ -60,6 +60,20 @@ class GaudiGPTBigCodeAttention(GPTBigCodeAttention):
 
         self.fused_scaled_dot_product_attention = ModuleFusedSDPA(FusedSDPA) if FusedSDPA is not None else None
         self.block_size = 4096
+        # Starting with v4.54 self.attn_dropout in GPTBigCodeAttention constructor is only assigned config.attn_pdrop value
+        # and not a Dropout operator, due to overall refactor of the class. Here we overwrite the value back with correct
+        # OP, to make use of the previous flow without changes.
+        self.attn_dropout = torch.nn.Dropout(config.attn_pdrop)
+
+    def _get_mask_value(self, device, dtype):
+        """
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
```diff
@@ -60,6 +60,20 @@ def __init__(self, config, is_cross_attention=False, layer_idx=None):
 
         self.fused_scaled_dot_product_attention = ModuleFusedSDPA(FusedSDPA) if FusedSDPA is not None else None
         self.block_size = 4096
+        # Starting with v4.54 self.attn_dropout in GPTBigCodeAttention constructor is only assigned config.attn_pdrop value
+        # and not a Dropout operator, due to overall refactor of the class. Here we overwrite the value back with correct
+        # OP, to make use of the previous flow without changes.
+        self.attn_dropout = torch.nn.Dropout(config.attn_pdrop)
```
pbielak: As dropout is only used in L147, maybe we should use the functional equivalent, i.e., `torch.nn.functional.dropout(attn_weights, p=self.config.attn_pdrop, training=self.training)` directly in L147, similarly to how `softmax` is applied in L145. WDYT @AKloniecki ?

AKloniecki: done

PR #2276, opened 2025-09-22
[opt] Fix in attention after transformers upgrade

ugolowic:
This commit fixes two issues that appeared in opt attention after transformers upgrade.
* incorrect shape of attn_weights caused mismatch in subsequent torch.bmm operation
* applying scaling to attn_weights significantly worsened accuracy

This change should also be applied to v1.20-release

Example:
```
PT_HPU_LAZY_MODE=1  python3  run_generation.py --batch_size 1 --bf16 --model_name_or_path facebook/opt-125m --use_hpu_graphs --n_iterations 1 --use_kv_cache --max_new_tokens 100
```

### DIFF
--- a/optimum/habana/transformers/models/opt/modeling_opt.py
+++ b/optimum/habana/transformers/models/opt/modeling_opt.py
@@ -107,9 +107,15 @@ def gaudi_opt_attention_forward(
     key_states = key_states.view(*proj_shape)
     value_states = value_states.view(*proj_shape)
 
-    attn_weights = torch.bmm(query_states, key_states.transpose(1, 2)) * self.scaling
+    src_len = key_states.size(1)
+    attn_weights = torch.bmm(query_states, key_states.transpose(1, 2))
+
     if attention_mask is not None:
-        attn_weights = attn_weights + attention_mask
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/opt/modeling_opt.py
```diff
@@ -107,9 +107,15 @@ def gaudi_opt_attention_forward(
     key_states = key_states.view(*proj_shape)
     value_states = value_states.view(*proj_shape)
 
-    attn_weights = torch.bmm(query_states, key_states.transpose(1, 2)) * self.scaling
+    src_len = key_states.size(1)
+    attn_weights = torch.bmm(query_states, key_states.transpose(1, 2))
```
astachowiczhabana, line 111: Can we really safely remove scaling? Original transformers code still have it: https://github.com/huggingface/transformers/blob/main/src/transformers/models/opt/modeling_opt.py#L160

astachowiczhabana: NVM. The scaling is removed here from BMM OP, but it's still applied to query_states before

astachowiczhabana:
"Next Synapse Release Validation" is failing on one test with unrelated reason. It looks clean.

regisss (approved):
LGTM

PR #2278, opened 2025-09-22
Ensure NumPy conversion and add scalar input warning in GaudiTrainer evaluation (datasets>=4.0.0)

gplutop7:
This PR strengthens GaudiTrainer’s evaluation loop to prevent type and shape mismatches when 
using datasets>=4.0.0.

Fix:

- Explicitly convert `all_preds` and `all_labels` to NumPy arrays once after accumulation.
- Added a diagnostic warning if predictions or labels are scalars (`ndim == 0`), which may indicate a dataset/collator issue.

### DIFF
--- a/optimum/habana/transformers/models/mllama/modeling_mllama.py
+++ b/optimum/habana/transformers/models/mllama/modeling_mllama.py
@@ -344,6 +344,7 @@ class GaudiMllamaTextCrossAttention(MllamaTextCrossAttention):
             )
 
         if FusedSDPA and use_flash_attention:
+            attn_weights = None
             import habana_frameworks.torch.hpu as ht
 
             if q_len == 1:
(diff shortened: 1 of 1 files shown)


regisss:
@gplutop7 Can you give an example of command that enables to reproduce this issue please?

regisss:
I think this should not happen because we do call `get_arrays` at line https://github.com/huggingface/optimum-habana/blob/1ce0173309db543dd13359ccf3183ce323f70950/optimum/habana/transformers/trainer.py#L2213
And `get_arrays` should already converts the variable to numpy and put it on CPU: https://github.com/huggingface/transformers/blob/d79b2d981f28b2730d402244ac3c2e9a8c054eee/src/transformers/trainer_pt_utils.py#L339

So, theoretically, we should not need this fix. But hard to say what happened when you got this issue.
I propose to keep this PR open for a while and we will revisit it if the issue occurs again. WDYT?

PR #2283, opened 2025-09-25
Fixes for Device Mismatch and Configuration Conflict

astachowiczhabana:
This PR addresses two issues introduced after upgrading to transformers==4.45:

**1. Device Mismatch in _expand_mask**
    **Issue Origin:** Introduced by upstream PR https://github.com/huggingface/transformers/pull/38637, which changed scalar creation from 1.0 to torch.tensor(1.0) on CPU.
    **Impact:** This led to a mix of CPU and HPU tensors in model files, causing unhandled exceptions/perf drops due to device mismatch.
    **Fix:** Rewrote _expand_mask to ensure the scalar tensor (inverted_mask) is created on the same device as the input mask (mask.device) to maintain device consistency.

**2. Configuration Conflict with LlamaConfig**
    **Issue Origin:** A combination of changes:
https://github.com/huggingface/transformers/issues/33771 introduced a docstring decorator that triggers lazy initialization of LlamaConfig.
Global import of modeling_declim.py in optimum-habana indirectly imports transformers.models.llama.modeling_llama, blocking LlamaConfig from optimum-habana.
    **Fix:** Used AutoConfig.register("llama", LlamaConfig, exist_ok=True) to overwrite the lazy-initialized LlamaConfig from transformers with the correct version from optimum-habana.

### DIFF
--- a/optimum/habana/transformers/modeling_attn_mask_utils.py
+++ b/optimum/habana/transformers/modeling_attn_mask_utils.py
@@ -132,7 +132,6 @@ class GaudiAttentionMaskConverter(AttentionMaskConverter):
         # [bsz, seq_len] -> [bsz, 1, tgt_seq_len, src_seq_len]
         return self._expand_mask(attention_mask_2d, dtype, tgt_len=input_shape[-1]).to(device)
 
-
     @staticmethod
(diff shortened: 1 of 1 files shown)


astachowiczhabana:
@regisss can you please check?

regisss (approved):

PR #2284, opened 2025-09-25
Fix next_decoder_cache formula based on use_new_cache after tranformers 4.55 upgrade

karol-brejna-i:
The statement `next_decoder_cache = () if not use_new_cache else None` was always evaluating to () because `use_new_cache` is hardcoded to False, making `not use_new_cache` always True.

Similar thing happens for optimum/habana/transformers/models/mistral/modeling_mistral.py, where there is even more dead code because of hardcoded `use_new_cache`.

This PR allows as to pass security scans required for the release.


karol-brejna-i:
Removed unused declaration, rebased + `make style`.

regisss (approved):

PR #2285, opened 2025-09-26
Revert to datasets==3.6.0 in audio classification module

gplutop7:
This PR reverts the audio classification module back to datasets==3.6.0.
The recent upgrade to datasets==4.0.0 caused compatibility issues with audio dataset loading (e.g., common_language parquet conversion). Restoring version 3.6.0 ensures stable dataset handling until a proper migration path is implemented.


regisss:
If the only blocking point here is datasets that are not in parquet format, I can convert them pretty quickly. WDYT?
And are there other known issues related to Datasets v4 so far?

PR #2293, opened 2025-10-02
Enable --dynamo_allow_unspec_int_on_nn_module by default in text-generation example

gplutop7:
This PR updates the text-generation example (run_generation.py) to set --dynamo_allow_unspec_int_on_nn_module to True by default.

In PyTorch 2.8, torch._dynamo.config.allow_unspec_int_on_nn_module needs to be explicitly enabled to restore the behavior from PT 2.7.1 and reduce unnecessary recompilations.

Previously, the flag existed in the example but defaulted to False unless explicitly passed.

With this change, the flag is automatically enabled for all text-generation tests and benchmarks, ensuring consistent and stable execution without requiring QA or users to modify their configs.

Users can still override the behavior by explicitly setting --no-dynamo_allow_unspec_int_on_nn_module if needed.

This aligns the text-generation example with the recommended configuration for PT 2.8+

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -451,6 +451,7 @@ def setup_parser(parser):
     parser.add_argument(
         "--dynamo_allow_unspec_int_on_nn_module",
         action="store_true",
+        default=True,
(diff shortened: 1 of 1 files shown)


regisss (approved):
LGTM

PR #2309, opened 2025-10-16
Fix `run_clm.py` for streaming datasets

pbielak:
# What does this PR do?

The variable `max_eval_samples` in L769 was referenced before assignment. This commit fixes the computation of `eval_samples` for the metrics report by computing it in a similar way as for the training split.

Note:
- bug introduced during Transformers 4.55 upgrade
- applies only to scenarios with `--streaming` enabled


pbielak:
Needs to be cherry-picked to `v1.20-release`

regisss (approved):
LGTM

PR #2310, opened 2025-10-16
Switch RAFT dataset source from ought/raft to regisss/raft for compatibility with datasets>=4.0.0

gplutop7:
This PR updates the RAFT dataset source from ought/raft to regisss/raft.
The change is required due to the upgrade to datasets 4.0.0, which no longer supports the original dataset format.
The new dataset version (regisss/raft) provides Parquet-based splits, ensuring full compatibility with the latest datasets library.


regisss (approved):
LGTM

PR #2301, opened 2025-10-09
[wav2vec2] Bring back gaudi fused sdpa attention

ugolowic:
One of the previous commits (https://github.com/huggingface/optimum-habana/pull/2248/commits/8cae0696885df7df9c37d3941cd0dff33bbf4eff) removed fused sdpa attention from wav2vec2, b/c it seemed like it wasn't used. This caused a performance drop in speech-recognition and audio-classification examples, as they defaulted to sdpa attention.

This PR brings back the attention implementation in alignment with the new attention mechanism introduced in transformers 4.55.

PR #2314, opened 2025-10-20
Remove flash attention flags from `run_clm.py`

pbielak:
# What does this PR do?

Due to PR [1], the arguments related to the usage of flash attention are now available in the training args. This commit removes these arguments from the `ModelArgs` class and updates the code that sets the model generation flags.

[1] https://github.com/huggingface/optimum-habana/pull/2301/files


AKloniecki (approved):
LGTM

PR #2316, opened 2025-10-21
Set missing baseline values for text_generation tests.

AKloniecki:
# What does this PR do?

Two test cases didn't have baseline values set up on Gaudi3, causing slow_tests failures.
This PR sets these baselines to results gathered from single device.


regisss (approved):
LGTM

PR #2319, opened 2025-10-23
Align MLlama code with Transformers 4.55

pbielak:
# What does this PR do?

After the Transformers 4.55 update, one of the attention classes failed to compute the attention scores due to mismatches between arguments in the `torch.matmul` op. This commits updates the whole `Mllama` code base to be fully aligned with the code in Transformers 4.55. In particular, it uses the `_attn_implementation` instead of custom classes.

### DIFF
--- a/examples/image-to-text/run_pipeline.py
+++ b/examples/image-to-text/run_pipeline.py
@@ -18,6 +18,7 @@ import glob
 import json
 import logging
 import os
+from contextlib import nullcontext
 from pathlib import Path
 
 import PIL.Image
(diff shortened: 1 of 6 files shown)


examples/image-to-text/run_pipeline.py
```diff
@@ -325,6 +325,11 @@ def main():
             ]
             args.prompt = processor.apply_chat_template(conversation, add_generation_prompt=True)
 
+    if model_type == "mllama" and args.use_flash_attention:
```
astachowiczhabana: why not just allow user to select attn_implementation + add readme section about it?

optimum/habana/transformers/integrations/gaudi_fused_sdpa_attention.py
```diff
@@ -18,13 +18,13 @@ def gaudi_fused_sdpa_attention_forward(
 ) -> tuple[torch.Tensor, None]:
     bsz, num_heads, tgt_len, head_dim = query.shape
 
+    softmax_mode = "fast" if os.getenv("FLASH_ATTENTION_FAST_SOFTMAX") == "1" else "None"
```
astachowiczhabana, line 21: since attention implementation is now separated from model we don't have to use env vars, can you explore if it's possible to use kwargs instead of env vars?

examples/image-to-text/run_pipeline.py
pbielak: This is done, because this script is also used by other models (such as llava), which are not yet aligned with the attn_implementation interface.

optimum/habana/transformers/integrations/gaudi_fused_sdpa_attention.py
pbielak: Makes sense - I will have a look at it

pbielak: We decided with Adam to leave the usage of env vars as is

examples/image-to-text/run_pipeline.py
pbielak: As currently only Mllama is using the `attn_implementation` interface, we will leave it as is

PR #2331, opened 2025-10-30
GaudiTrainer: Take accuracy measurement outside of timer

ugolowic:
In GaudiTrainer, accuracy calculation happened between start_time and end_time, so its duration was included in calculating time for performance measurements.
It was negligible for datasets < 4.0.0, but with the new package, it takes much more time due to numpy array conversion deep in the scikit-learn library, causing incorrect perf reporting.

### DIFF
--- a/optimum/habana/diffusers/pipelines/controlnet/pipeline_controlnet.py
+++ b/optimum/habana/diffusers/pipelines/controlnet/pipeline_controlnet.py
@@ -634,11 +634,13 @@ class GaudiStableDiffusionControlNetPipeline(GaudiDiffusionPipeline, StableDiffu
                     self.htcore.mark_step()
 
             hb_profiler.stop()
+            end_time = time.time()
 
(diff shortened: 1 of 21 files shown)


ugolowic:
> This change looks good to me. I see that `speed_metrics` is also called in [trainer_qa.py](https://github.com/huggingface/optimum-habana/blob/main/examples/question-answering/trainer_qa.py), [trainer_seq2seq_qa.py](https://github.com/huggingface/optimum-habana/blob/main/examples/question-answering/trainer_seq2seq_qa.py) and [some diffusion pipelines](https://github.com/huggingface/optimum-habana/tree/main/optimum/habana/diffusers/pipelines), I guess they should be updated too right?

You're right about the pipelines, I'll fix that. However, `trainer_qa.py` and `trainer_seq2seq.py` call `speed_metrics()` from `transformers.trainer_utils`, not from `optimum.habana.utils`.

PR #2336, opened 2025-11-04
add flux kontext related files

tianyuan0211:
# What does this PR do?

Fixes # (issue)

## Before submitting


regisss:
I agree with @karol-brejna-i + please add this model to README/docs tables ([here](https://github.com/huggingface/optimum-habana/blob/c7c1b3d630e1f0c523aa687c8ccf272144b14b9d/README.md?plain=1#L296) and [there](https://github.com/huggingface/optimum-habana/blob/c7c1b3d630e1f0c523aa687c8ccf272144b14b9d/docs/source/index.mdx?plain=1#L121)) and add some tests that can be directly taken and adapted from Diffusers: https://github.com/huggingface/diffusers/blob/main/tests/pipelines/flux/test_pipeline_flux_kontext.py

PR #2339, opened 2025-11-06
Set default generation config for TextGenerationPipeline and TextToAudioPipeline

AKloniecki:
# What does this PR do?

After transformers upgrade pipelines need to have `_default_generation_config` set to `GaudiGenerationConfig` to properly initialize. This PR sets this value for TextGenerationPipeline and TextToAudioPipeline classes.


regisss (approved):
LGTM!

PR #2350, opened 2025-11-26
Add validation for negative_prompt_embeds to satisfy Coverity null-check analysis

gplutop7:
This PR adds an explicit validation step ensuring that `negative_prompt_embeds` is not None 
when classifier-free guidance (CFG) is enabled. 

Coverity reported false-positive `null_check` and `FORWARD_NULL` warnings because 
`negative_prompt_embeds` is declared as optional, even though Diffusers always generates 
valid embeddings when CFG is active.


regisss (approved):
LGTM

PR #2353, opened 2025-12-03
Fix gradient checkpointing in GaudiGemmaDecoderLayer + align forward() signature with HF 4.55

gplutop7:
This PR fixes a hidden_states=None failure in GaudiGemmaDecoderLayer when gradient checkpointing is enabled in distributed training.

Key changes

- Remove non-standard forward arguments (token_idx, reuse_cache, attn_softmax_bf16, use_flash_attention, etc.) from the method signature.
- Move all Gaudi-specific parameters to **kwargs and retrieve them via kwargs.get(...) (HF-compatible behavior).
- Add robust reconstruction logic for hidden_states when positional args are stripped by checkpointing wrappers.
- Ensure compatibility with Transformers 4.55 forward dispatching, PEFT, DDP, and Accelerate.
- Fully restore training stability on Gaudi with gradient checkpointing enabled

### DIFF
--- a/optimum/habana/transformers/models/gemma/modeling_gemma.py
+++ b/optimum/habana/transformers/models/gemma/modeling_gemma.py
@@ -272,10 +272,11 @@ class GaudiGemmaAttention(GemmaAttention):
     def pre_attn_forward(
         self,
         hidden_states: torch.Tensor,
-        position_embeddings: tuple[torch.Tensor, torch.Tensor],
         attention_mask: Optional[torch.Tensor],
+        position_ids: Optional[torch.LongTensor] = None,
         past_key_value: Optional[Cache] = None,
-        use_cache: bool = False,
+        output_attentions: Optional[bool] = False,
+        use_cache: Optional[bool] = False,
         cache_position: Optional[torch.LongTensor] = None,
         token_idx: Optional[torch.Tensor] = None,
         attn_softmax_bf16: Optional[bool] = False,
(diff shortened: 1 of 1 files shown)


optimum/habana/transformers/models/gemma/modeling_gemma.py
```diff
@@ -450,68 +460,84 @@ def update_sincos_cache(self, seq_len):
...
         - add new args attn_softmax_bf16
         """
+        token_idx = kwargs.get("token_idx", None)
+        attn_softmax_bf16 = kwargs.get("attn_softmax_bf16", False)
+        reuse_cache = kwargs.get("reuse_cache", False)
+        use_flash_attention = kwargs.get("use_flash_attention", False)
+        flash_attention_recompute = kwargs.get("flash_attention_recompute", False)
+        flash_attention_causal_mask = kwargs.get("flash_attention_causal_mask", False)
+        cache_idx = kwargs.get("cache_idx", None)
+        position_embeddings = kwargs.get("position_embeddings", None)
```
regisss, line 524: Hmm weird we don't need to do that for other models no?

gplutop7: Gradient checkpointing in Transformers 4.55 may call layers without positional args, passing only **kwargs (due to PyTorch CheckpointFunction + HF wrappers). We observed this in Gemma, hence the need to read these params from kwargs.

PR #2362, opened 2026-01-16
Temporarily remove broken examples for OH 1.21.0 release

gplutop7:
It looks like there are still some unresolved issues with the example scripts, but we would like the OH 1.20.0 release to proceed as planned.
To avoid blocking the release, this PR temporarily removes the affected examples. We plan to fix and reintroduce them in an upcoming point release once the issues are resolved.


regisss (approved):
LGTM

